PR #31, opened 2022-01-31
Pylint - manual workflow

kksat:
I think we do not need automated workflow for pylint at the moment. Not much py code.


berndfinger:
lgtm

Issue #44, opened 2022-02-07
sap_hana_install: Support checksum checking for SAPCAR EXE and SAR files


berndfinger:
Checksums can be retrieved from the SAP download page and stored in the second field of variables sap_hana_install_sapcar_file and sap_hana_install_sarfiles, e.g.:

`# File name and checksum of SAPCAR EXE file:`
`sap_hana_install_sapcar_file:`
`  - { name: 'SAPCAR_XXXX-XXXXXXXX.EXE', checksum: 'XXX' }`
`sap_hana_install_sarfiles:`
`  - { name: 'IMDB_SERVERXXXXX_X-XXXXXXXX.SAR', checksum: 'XXX' }`
`  - { name: 'SAPHOSTAGENTXXXXX-XXXXXXXX.SAR', checksum: 'XXX' }`

Then the role can compare this checksum with the result of the sha256sum command run for the SAPCAR EXE file and each SAR file to be extracted by the role. We could store the result of the sha256sum command in a file named FILE.SAR.sha256sum. If this file does not exist, the role could created one.

berndfinger:
Solved in pull request #45.

berndfinger:
I closed the pull request without merging because we decided to use a different approach while keeping backward compatibility:
The role will perform checksum checking for any file for which a sha256sum file is available in the software directory.

berndfinger:
Solved in pull request #46.

berndfinger:
It appears that the standard naming convention for checksum files, if not stored in a central file, is without "sum" at the end. Example: IMDB_SERVERXXXXX_X-XXXXXXXX.SAR.sha256

PR #45, opened 2022-02-07
sap_hana_install: Support checksums for SAPCAR executable and SAR files (issue #44)

berndfinger:
This commit adds support for verifying checksums of SAPCAR executable and SAR files. Automatic detection of SAPCAR executables and automatic extraction of SAR files remains in place.


berndfinger:
We will choose a different approach for checksum checking, keeping the existing variable names.

Issue #50, opened 2022-02-10
sap_hana_install: Tries to change permissions on readonly directory


rhmk:
When the HANA archives are presented from read-only directory, and the follwoing variables are set:
```
sap_hana_install_sarfiles_list:
  - "SAPHOSTAGENT51_51-20009394.SAR "  
  - "IMDB_SERVER20_060_0-80002031.SAR"
sap_hana_install_software_directory: /path_to_readonly-NFS-mount
sap_hana_install_software_extract_directory: /tmp/sap-hana
```
The role runs into the following error:
```
TASK [community.sap_install.sap_hana_install : SAP HANA Pre Install - Change ownership of software directory '/rh-ava-packages/sap'] ***
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: OSError: [Errno 30] Read-only file system: b'/rh-ava-packages/sap/IMDB_CLIENT20_011_14-80002082.SAR'
```
There is no need to change ownership on readonly dir ....

rhmk:
Directory is also exported as noexec => hence "Get SAPCAR task is failing:
```
"/bin/sh: line 13: ./SAPCAR_1115-70006178.EXE: Permission denied"
```
moving .exe to /tmp for executing first

PR #51, opened 2022-02-10
sap_hana_install: Bugfix for Issue #50 

rhmk:
Issue #50

### DIFF
--- a/roles/sap_hana_install/tasks/pre_install.yml
+++ b/roles/sap_hana_install/tasks/pre_install.yml
@@ -34,15 +34,6 @@
         fail_msg: "FAIL: The software directory '{{ sap_hana_install_software_directory }}' does not exist!"
         success_msg: "PASS: The software directory '{{ sap_hana_install_software_directory }}' exist."
 
-    - name: SAP HANA Pre Install - Change ownership of software directory '{{ sap_hana_install_software_directory }}'
-      ansible.builtin.file:
-        path: "{{ sap_hana_install_software_directory }}"
-        state: directory
-        recurse: yes
-        mode: '0755'
-        owner: root
-        group: root
-
     - name: SAP HANA Pre Install - Change ownership of HANA directories
       ansible.builtin.file:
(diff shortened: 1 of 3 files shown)


roles/sap_hana_install/tasks/pre_install.yml
```diff
@@ -63,7 +54,16 @@
...
+    - name: SAP HANA Pre Install - Change ownership of software extract directory '{{ sap_hana_install_software_extract_directory }}'
+      ansible.builtin.file:
+        path: "{{ sap_hana_install_software_extract_directory }}"
+        state: directory
+        recurse: yes
+        mode: '0755'
+        owner: root
+        group: root
+
+# In case more than one installation is ongoing and extracting to the same shared directory, wait until the extraction has completed:
```
berndfinger, line 66: Needs to be removed. We are checking if the extract directory exists. If it exists, we will later search for a directory named `SAP_HANA_DATABASE`. If it does not exist or if it exists are there is no directory named `SAP_HANA_DATABASE` in it, we will remove and recreate the directory with mode 0755. This happens in  tasks/pre_install/hdblcm_prepare.yml

roles/sap_hana_install/tasks/pre_install/hdblcm_prepare.yml
```diff
@@ -58,7 +72,9 @@
                    done | awk '$2=="{{ ansible_architecture }}"{print $1}'
                 ); do
            ( printf "%s " ${sapcar_matching_arch}
-             ./${sapcar_matching_arch} --version |
+             cp ${sapcar_matching_arch} {{ sap_hana_install_software_extract_directory }} &&\
+             chmod 755 {{ sap_hana_install_software_extract_directory }}/${sapcar_matching_arch}
```
berndfinger, line 76: We can also run a chmod root:root here so we do not need to call the file module later.

berndfinger:
lgtm

PR #59, opened 2022-02-16
sap_hana_install: Improve the hdblcm configfile processing

berndfinger:
By default, the hdblcm configfile will now be created dynamically in each run, as follows: After the role has found the hdblcm command or extracted the SAP HANA SAR file, it will call the hdblcm command with the option `dump_configfile_template` to create a configfile template, which will then be converted into a Jinja2 configfile template according to the following rules: For each hdblcm parameter, the value will be either the value of the role variable prepended by the role name and an underscore, or a default (if present in the hdblcm configfile template).

Example: The value of hdblcm parameter `system_usage` will be set to the value of role variable `sap_hana_install_system_usage` or to `custom` in case the role variable has not been set.

The result of the templating is a new, customized hdblcm configfile, which will be used by the hdblcm command for the SAP HANA installation.

If there is a file named `configfile.cfg` in the directory specified by role variable `sap_hana_install_configfile_directory`, this file will be used and no dynamic hdblcm configfile processing will be performed. So any modifications to role variables after creation of this file will not be reflected.


berndfinger:
Install tests, including scale-out, succeeded.

Issue #83, opened 2022-03-09
sap_hana_install: Replace ansible.posix.firewall


berndfinger:
The firewall module is not part of ansible.builtin, so we should replace it.
We could either use the firewall role from the Linux/RHEL System Roles collection or use the command or shell module to run the firewall-cmd command.

berndfinger:
Fixed in PR #84.

PR #84, opened 2022-03-09
sap_hana_install: Replace ansible.posix.firewall (#83)

berndfinger:
We can replace ansible.posix.firewall by a executing the firewall-cmd command via the ansible.builtin.command module, by constructing an argument list from the same variable which is used by the Linux System Roles `firewall` role.
By adding a new tag `sap_hana_install_configure_firewall`, we can let the role only update the firewall settings for SAP HANA.
The associated role variable for configuring the firewall settings remains `sap_hana_install_update_firewall` with its default `no`.


kksat:
1. Ports 
              - "30105"
              - "30107"
              - "30140"
              
              are included in range 
- "3{{ sap_hana_install_number }}00-3{{ sap_hana_install_number }}90"

I do not see a reason why we should add them separately. 

2. SMART data integration ports 
5050
and SAP HANA Smart Data Streaming ports
              - "9090"
              - "9091"
              - "9092"
              - "9093"
Are not required for every HANA installation. Why are we including them by default? But not including ports for HTTP (XSA) 80<NN>? And at the same time include HTTPS 43<NN>? 

3. Don't you think this should be separate role? So it can be called as day 2 automation role for already installed HANA system? 

4. Are HA ports for CCSPs included? 
5. From your point of view should ssh port be included?

berndfinger:
1 - Taking the example of a SAP HANA installation number `77`, the port range of `3{{ sap_hana_install_number }}00-3{{ sap_hana_install_number }}90` would be `37700-37790`. So we cannot remove the three ports. However, there are some examples in the list which can be converted to a range, e.g. `'9090/tcp-9093/tcp',`
2 - I cannot answer this question, as I focused on replacing the call to `ansible.posix.firewall`. The topic of which ports we need to open for SAP HANA should also be discussed, maybe in a separate issue.
3 - Yes, having a separate role is an idea which already came up. For the time being, we could leave the current firewall related code as part of sap_hana_install, taking into account that this code is only used if the default setting for variable `sap_hana_install_update_firewall` is changed from `no` to `yes`. There are more options available:
- introduce a user variable, e.g. named `sap_hana_install_firewall_ports` which can be used to set your own range of ports
- (suggested by @rhmk): Use firewalld service configuration files as documented in https://firewalld.org/documentation/man-pages/firewalld.service.html and https://firewalld.org/documentation/howto/add-a-service.html
4 - Probably not.
5 - I cannot answer this question but in my tests, the ssh connection to the managed nodes remained open.

berndfinger:
Regarding item 5, the following output probably explains why we don't have to include ssh here:
`# firewall-cmd --list-services
cockpit dhcpv6-client ssh`

berndfinger:
Tests completed.

PR #90, opened 2022-03-21
sap_hana_install: quickfix 

rhmk:
if sapcar is in a different directory than the IMDB* file, it cannot be used. 
Hence I added "| basename" to the variable where the exact filename is needed. This doesn't change the behavior, but enables me to specify sap_hana_install_sapcar_filename relative to the HANA software directory. 
EXAMPLE:
```
# sap_hana_install
#------------------
sap_hana_install_software_directory: /software/HANA_installation
sap_hana_install_sapcar_filename: ../SAPCAR/SAPCAR_1311-80000935.EXE
sap_hana_install_common_master_password: "R3dh4t$123"
sap_hana_install_sid: 'RHE'
sap_hana_install_instance_number: "00"

```
It is tested and works


rhmk:
merged, due to urgent need in a customer demo

PR #102, opened 2022-04-06
add sles to meta

rainerleber:


sean-freeman:
Agree missing metadata needs to be appended. Merge approved

PR #117, opened 2022-04-24
SLES fix package upgrade

rainerleber:
The hana pre configure role fails because of package name changes


sean-freeman (approved):
Approved for merge

PR #140, opened 2022-05-23
initial pull request for HANA HA HSR

mmoster:
initial pull request for HANA HA HSR
to get the updated system roles

### DIFF
--- a/roles/sap_hana_install/defaults/main.yml
+++ b/roles/sap_hana_install/defaults/main.yml
@@ -91,7 +91,7 @@ sap_hana_install_components: 'all'
 
 # Instance details
 sap_hana_install_sid: "{{ sap_hana_sid }}"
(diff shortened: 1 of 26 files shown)


roles/sap_hana_install/defaults/main.yml
```diff
@@ -90,7 +90,7 @@ sap_hana_install_components: 'all'
 #sap_hana_install_hdblcm_extraargs: '--ignore=check_diskspace,check_min_mem'
 
 # Instance details
-sap_hana_install_sid:
+sap_hana_install_sid: "{{ sap_hana_sid }}"
 sap_hana_install_number:
```
berndfinger, line 94: Please also add "{{ sap_hana_instance_number }}" here, as in the other three roles.

Issue #145, opened 2022-05-31
sap*preconfigure: Flush handlers at the end of the last preconfigure role


berndfinger:
The last preconfigure role in a play should be able to perform a reboot if necessary, so that the reboot does not occur at the end of the play in case other roles (e.g. sap_ha_install_pacemaker) are called in the same play.

PR #157, opened 2022-06-22
sap*preconfigure: flush handlers at the end

berndfinger:
Notes:
- We flush the handlers at the end of roles sap_general_preconfigure and sap_hana_preconfigure, for maximum flexibility.
- We do not flush the handlers at the end of role sap_netweaver_preconfigure because the handlers/main.yml file is empty.

Solves issue #145.

Note: With the following parameter settings, only one reboot will be performed, at the end of role sap_hana_preconfigure:
```
    sap_general_preconfigure_reboot_ok: no
    sap_general_preconfigure_fail_if_reboot_required: no
    sap_hana_preconfigure_reboot_ok: yes
    sap_hana_preconfigure_fail_if_reboot_required: no
```

### DIFF
--- a/roles/sap_general_preconfigure/tasks/main.yml
+++ b/roles/sap_general_preconfigure/tasks/main.yml
@@ -86,3 +86,6 @@
     - '{{ ansible_distribution.split("_")[0] }}'
     - '{{ ansible_distribution }}'
     - '{{ ansible_os_family }}.yml'
+
+# allow a reboot at the end of the preconfigure role
+- ansible.builtin.meta: flush_handlers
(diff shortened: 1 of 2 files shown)


sean-freeman:
Agree with this approach, this will force the handler to immediately execute for any of the Ansible Tasks which have notified the handler using:
- `notify: __sap_general_preconfigure_reboot_handler`
- `notify: __sap_hana_preconfigure_reboot_handler`

This will allow an OS Reboot to execute after the Ansible Role has completed, and continue with additional activities thereafter.

sean-freeman (commented):
lgtm

PR #177, opened 2022-07-14
HA roles: bug fixes and support for IP based inventory

ja9fuchs:
Bugfixes in sap_ha_set_hana: 
- accidentally had a default variable disabled that is needed to work correctly on RHEL >8.2 managed nodes
=> default variable enabled again
- role variable name prefix inconsistency that would fail the role when run in a separate play
=> role variable names in tasks updated to the correct prefix

Improvements in sap_ha_install_hana_hsr:
- During HANA System Replication setup the task delegation to the primary node fails when the inventory is IP based and the Ansible control node does not have a hostname resolution
=> New tasks are checking connectivity, using the primary node IP as alternative, if the IP connects and the `hostname -s` matches the node_name parameter.

- For PKI file copy from primary to secondary node the Ansible control node was used as temporary file storage.
=> To avoid storage dependency on the Ansible control node, it now configures a temporary custom ssh key for a direct file sync between primary and secondary nodes. The temporary changes are reverted immediately afterwards.

This PR also contains some updates based on ansible-lint check results.


sean-freeman:
LGTM :+1:

sean-freeman (approved):
LGTM :+1:

berndfinger:
lgtm.

PR #180, opened 2022-07-18
[HSR] PKI filersync fix and [HA roles] linter config symlinks

ja9fuchs:
**sap_ha_install_hana_hsr**
- PKI file copy via rsync not working when size and mtime of source and destination are identical
-> added `--checksum` for more reliable file comparison
-> added `-vv` for more output and parsing as `changed_when` indicator

**HA roles**
- missing role based ansible-lint and yamllint config files leads to more errors/warnings based on linter defaults
-> symlinks to repo parent created for a consistent linter config in all roles

### DIFF
--- a/roles/sap_ha_install_hana_hsr/.ansible-lint
+++ b/roles/sap_ha_install_hana_hsr/.ansible-lint
@@ -1,0 +1,1 @@
+../../.ansible-lint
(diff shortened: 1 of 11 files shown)


sean-freeman (approved):
lgtm!

Issue #181, opened 2022-07-19
sap_ha_install_hana_hsr: Doesn't work with multiple interfaces


rhmk:
The variable configuration asks for a node name, which has to be a shortname and not a FQDN 
This does not work on a multihomed system where the names and interfaces  are configured like this
```
72.25.250.22    hana1.lab.example.com hana1
172.25.250.23    hana2.lab.example.com hana2
192.168.0.22    hana1.private.example.com
192.168.0.23    hana2.private.example.com
``` 
If I choose the following config, the system fails with multiple entries of  hana1 in /etc/hosts

```
  - node_name: hana1
    node_ip: "172.25.250.22"
    node_role: primary
    hana_site: DC01

  - node_name: hana2
    node_ip: "172.25.250.23"
    node_role: secondary
    hana_site: DC02
```

If I choose this config  the system doesn't do anything due to a wrong comparison (when statement)
```
  - node_name: hana1.lab.example.com
    node_ip: "172.25.250.22"
    node_role: primary
    hana_site: DC01

  - node_name: hana2.lab.example.com
    node_ip: "172.25.250.23"
    node_role: secondary
    hana_site: DC02
```
Output from playbook
```
[community.sap_install.sap_ha_install_hana_hsr : SAP HSR - Enable HANA System Replication on primary node] ***************************************************************************************************
task path: /home/student/.ansible/collections/ansible_collections/community/sap_install/roles/sap_ha_install_hana_hsr/tasks/configure_hsr.yml:17
skipping: [hana1.lab.example.com] => (item=hana1.lab.example.com)  => {"ansible_loop_var": "item", "changed": false, "item": {"hana_site": "DC01", "node_ip": "172.25.250.22", "node_name": "hana1.lab.example.com", "node_role": "primary"}, "skip_reason": "Conditional result was False"}
skipping: [hana1.lab.example.com] => (item=hana2.lab.example.com)  => {"ansible_loop_var": "item", "changed": false, "item": {"hana_site": "DC02", "node_ip": "172.25.250.23", "node_name": "hana2.lab.example.com", "node_role": "secondary"}, "skip_reason": "Conditional result was False"}
[...]

PR #182, opened 2022-07-19
Dev: sap_ha_install_hana_hsr: Doesn't work with multiple interfaces #181

rhmk:
I started to implement the first fixes for sap_ha_install_hana_hsr to work in the lab environment
The most important fixes are:
- accepts now fqdn instead of short name for hsr interface
- proper check if full qualified domain name is given instead of short
- proper check if short name exists as substring in other names (not fully tested)
- check if hsr interface is configured
- creates internal variable with proper part of sap_hana_cluster_nodes dict

Not fixed:
- can't define HSR interface on an interface that has the same short name like the primary. This will make a wrong entry to /etc/hosts. Identical short hostnames with different domains on different interfaces will create duplicate host entry for short name

Role tested against RH445 environment

### DIFF
--- a/roles/sap_ha_install_hana_hsr/tasks/configure_hsr.yml
+++ b/roles/sap_ha_install_hana_hsr/tasks/configure_hsr.yml
@@ -26,7 +26,7 @@
   register: enablesr
   changed_when: "'successfully enabled system as system replication source site' in enablesr.stdout"
   when:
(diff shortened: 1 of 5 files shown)


ja9fuchs:
Test deployment using the code is ongoing...

ja9fuchs (approved):
lgtm

PR #186, opened 2022-07-23
sap_hana_preconfigure - Fixes fatal looping error, reboot logic

rdvansloten:
A `dict` is being passed into a `string` field. This is not supported. Running the role as-is from main produces this error for every step that uses the `name: "{{ packages }}"` syntax:

```
TASK [sap_hana_preconfigure : 1944799 - PDF 8.1 Package List Packages (SLE-Module-Legacy15)] ***
fatal: [heiq1hhdb05]: FAILED! => {"changed": false, "msg": "argument 'name' is of type <class 'dict'> and we were unable to convert to list: <class 'dict'> cannot be converted to a list"}
```

```
TASK [sap_hana_preconfigure : 2578899 - SAP HANA database] *********************
fatal: [heiq1hhdb05]: FAILED! => {"changed": false, "msg": "argument 'name' is of type <class 'dict'> and we were unable to convert to list: <class 'dict'> cannot be converted to a list"}
```

I have converted all package installs to a `with_items` list that is fed to the `name:` field using the `{{ item }}` variable. This allows the role to execute correctly.

Additionally, the reboot logic handler is incorrect. The handler asserts that no reboot is required when the "Reboot the managed node" step is played, but then asserts a reboot is required during the "Let the role fail if a reboot is required" step. This has been repaired by adding the same checks to the reboot step as the playbook failure handler. It will now reboot the server correctly if a reboot is required. See for the current logic error:

```
RUNNING HANDLER [sap_hana_preconfigure : Reboot the managed node] **************
skipping: [myvm] => {"changed": false, "skip_reason": "Conditional result was False"}

RUNNING HANDLER [sap_hana_preconfigure : Let the role fail if a reboot is required] ***
fatal: [myvm]: FAILED! => {"changed": false, "msg": "Reboot is required!"}

RUNNING HANDLER [sap_hana_preconfigure : Show a warning message if a reboot is required] ***
```

The loop variable `{{ item }}` is already in use and generates warnings:

```
TASK [sap_hana_preconfigure : 2588899 - I/O scheduler] *************************
[WARNING]: TASK: sap_hana_preconfigure : 2588899 - I/O scheduler: The loop
variable 'item' is already in use. You should set the `loop_var` value in the
`loop_control` option for the task to something else to avoid variable
collisions and unexpected behavior.
```

I have fixed this with a `loop_control` block.

### DIFF
--- a/roles/sap_hana_preconfigure/handlers/main.yml
+++ b/roles/sap_hana_preconfigure/handlers/main.yml
@@ -60,7 +60,7 @@
 - name: "Debug grub-mkconfig UEFI"
   ansible.builtin.debug:
     var: __sap_hana_preconfigure_register_grub2_mkconfig_uefi_mode.stdout_lines,
-         __sap_hana_preconfigure_register_grub2_mkconfig_uefi_mode.stderr_lines
+      __sap_hana_preconfigure_register_grub2_mkconfig_uefi_mode.stderr_lines
   listen: __sap_hana_preconfigure_regenerate_grub2_conf_handler
   when:
     - __sap_hana_preconfigure_register_stat_sys_firmware_efi.stat.exists
@@ -77,7 +77,7 @@
     test_command: /bin/true
   listen: __sap_hana_preconfigure_reboot_handler
   when:
-    - sap_hana_preconfigure_reboot_ok|d(false)
+    - not sap_hana_preconfigure_reboot_ok|d(false)
 
 - name: Let the role fail if a reboot is required
   ansible.builtin.fail:
--- a/roles/sap_hana_preconfigure/tasks/sapnote/1944799/installation.yml
+++ b/roles/sap_hana_preconfigure/tasks/sapnote/1944799/installation.yml
@@ -5,39 +5,37 @@
 # show zypper patterns
 - name: "1944799 - PDF 8.1 Package List Pattern Also 3.5 Software selection"
   community.general.zypper:
-    name: "{{ pattern }}"
+    name: "{{ item }}"
     type: pattern
-  vars:
-    pattern:
-      - gnome_basic
-      - base
-      - enhanced_base
-      - apparmor
-      - 32bit
-      - yast2_basis
-      - sw_management
-      - fonts
-      - x11
-      - sap-hana
-      - sap_server
+  with_items:
+    - gnome_basic
+    - base
+    - enhanced_base
+    - apparmor
+    - 32bit
+    - yast2_basis
+    - sw_management
+    - fonts
+    - x11
+    - sap-hana
+    - sap_server
 
 # Requires SLE-Module-Legacy15 Module
 - name: "1944799 - PDF 8.1 Package List Packages (SLE-Module-Legacy15)"
   community.general.zypper:
-    name: "{{ packages }}"
+    name: "{{ items }}"
     type: package
-  vars:
-    packages:
-      - libssh2-1
-      - libopenssl1_1
-      - libstdc++6
-      - libatomic1
-      - libgcc_s1
-      - libltdl7
-      - insserv
-      - numactl
-      - system-user-uuidd
-      - unzip
+  with_items:
+    - libssh2-1
+    - libopenssl1_1
+    - libstdc++6
+    - libatomic1
+    - libgcc_s1
+    - libltdl7
+    - insserv
+    - numactl
+    - system-user-uuidd
+    - unzip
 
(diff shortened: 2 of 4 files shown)


ja9fuchs:
Regarding commit https://github.com/sap-linuxlab/community.sap_install/pull/186/commits/fbb69c53baf756e040b8a7d9904c59d547ca5837:

According to Ansible documentation of community.general.zypper, the `name` parameter type is `list / elements=string` (list of strings). The tasks should be fine as they are and no change required.
https://docs.ansible.com/ansible/latest/collections/community/general/zypper_module.html#parameter-name

rdvansloten:
> Regarding commit [fbb69c5](https://github.com/sap-linuxlab/community.sap_install/commit/fbb69c53baf756e040b8a7d9904c59d547ca5837):
> 
> According to Ansible documentation of community.general.zypper, the `name` parameter type is `list / elements=string` (list of strings). The tasks should be fine as they are and no change required. https://docs.ansible.com/ansible/latest/collections/community/general/zypper_module.html#parameter-name

Yes, but the script is passing a `dict`, according to the error. A `dict` is not a list. Read the error thrown by Ansible, or try running the role on your own machine. The task is not fine.

```
unable to convert to list: <class 'dict'> cannot be converted to a list"
```

ja9fuchs:
> > Regarding commit [fbb69c5](https://github.com/sap-linuxlab/community.sap_install/commit/fbb69c53baf756e040b8a7d9904c59d547ca5837):
> > According to Ansible documentation of community.general.zypper, the `name` parameter type is `list / elements=string` (list of strings). The tasks should be fine as they are and no change required. [https://docs.ansible.com/ansible/latest/collections/community/general/zypper_module.`tml#parameter-name](https://docs.ansible.com/ansible/latest/collections/community/general/zypper_module.%60tml#parameter-name)
> 
> Yes, but the play is passing a `dict`, according to the error. A `dict` is not a `list`. Read the error thrown by Ansible, or try running the role on your own machine. The task is not fine.
> 
> ```
> unable to convert to list: <class 'dict'> cannot be converted to a list"
> ```

I see, this is unexpected. We will need time to review your suggestions and test the issue.

roles/sap_general_preconfigure/handlers/main.yml
```diff
@@ -8,14 +8,6 @@
   when:
     - sap_general_preconfigure_reboot_ok|d(false)
 
-- name: Let the role fail if a reboot is required
```
berndfinger, line 11: This is all on purpose. We want the role to fail by default if a reboot is required but the role is not configured to automatically reboot the managed node. A reboot requirement should not remain unnoticed by default.
The failure can be switched off by either:
- setting the variable `sap_general_preconfigure_fail_if_reboot_required` to `no` or
- setting the variable `sap_general_preconfigure_reboot_ok` to `yes`.

roles/sap_hana_preconfigure/handlers/main.yml
```diff
@@ -77,14 +77,6 @@
     test_command: /bin/true
   listen: __sap_hana_preconfigure_reboot_handler
   when:
-    - sap_hana_preconfigure_reboot_ok|d(false)
```
berndfinger, line 80: See my comment https://github.com/sap-linuxlab/community.sap_install/pull/186/files#r945667552 .

roles/sap_general_preconfigure/handlers/main.yml
berndfinger: @rainerleber Can you please have a look at this and the other issues in this PR?
@rdvansloten Can you please provide your role variables with regards to the reboot handling?

rainerleber:
Hi @rdvansloten thank you for contribution which ansible version and community.general version do you use ? I can't see any issues when i am testing it on my machine. The problem with the looping is that it will take a lot of more time then calling it once.

I love the idea for the loop_control improvement.

sean-freeman:
This PR is problematic, it is describing multiple issues and some is misinterpretation of logic - therefore the PR cannot be merged as some code is not accurate.

I am also concerned with `loop` with the zypper Ansible Module, as this would create unnecessary time calling zypper multiple times. There is a specific [note](https://docs.ansible.com/ansible/latest/collections/community/general/zypper_module.html#id4) in the documentation describing this:
`When used with a loop: each package will be processed individually, it is much more efficient to pass the list directly to the name option.
`

I will close this PR and create a new PR for the valid code changes.

PR #187, opened 2022-07-25
Generation of HDBLCM config file

ranjithrajaram:
Two node install of SAP HANA using RHEL for SAP Collections fails because the config for node 2 uses the hostname of node 1

Error message
```
fatal: [dhcp200-237.redhat.com]: FAILED! =>
{
  "changed": true,
  "cmd": [
    "./hdblcm",
    "--ignore=check_diskspace,check_min_mem",
    "--hdbinst_server_xs_engine=off",
    "--hdbinst_server_import_content=off",
    "--configfile=/tmp/ansible.ia7cu4lchanaconfig/configfile.cfg",
    "-b"
  ],
  "delta": "0:00:04.955358",
  "end": "2022-07-21 12:25:10.811444",
  "msg": "non-zero return code",
  "rc": 1,
  "start": "2022-07-21 12:25:05.856086",
  "stderr_lines": [
    "Running in batch mode",
    "  IP address '10.65.200.241' (host 'dhcp200-241.redhat.com') is not assigned to a local network interface"
  ],
  "stdout_lines": [
    "",
    "",
    "SAP HANA Lifecycle Management - SAP HANA Database 2.00.063.00.1655123455",
    "************************************************************************",
    "",
    "",
    "Scanning software locations...",
    "Detected components:",
    "    SAP HANA Database (2.00.063.00.1655123455) in /software/hana/extracted/SAP_HANA_DATABASE/server",
    "Log file written to '/var/tmp/hdb_DB1_hdblcm_install_2022-07-21_12.25.06/hdblcm.log' on host 'dhcp200-237'."
  ]
}
```

### DIFF
--- a/roles/sap_hana_install/tasks/pre_install/hdblcm_configfile.yml
+++ b/roles/sap_hana_install/tasks/pre_install/hdblcm_configfile.yml
@@ -31,7 +31,7 @@
   block:
 
     - name: SAP HANA Pre Install - Create the hdblcm configfile template '{{ sap_hana_install_configfile_directory }}/{{ sap_hana_install_configfile_template_prefix }}.cfg'
-      ansible.builtin.command: "{{ __sap_hana_install_fact_hdblcm_path }}/hdblcm --dump_configfile_template={{ sap_hana_install_configfile_directory }}/{{ sap_hana_install_configfile_template_prefix }}.cfg"
+      ansible.builtin.command: "{{ __sap_hana_install_fact_hdblcm_path }}/hdblcm --dump_configfile_template={{ sap_hana_install_configfile_directory }}/{{ sap_hana_install_configfile_template_prefix }}-{{ ansible_hostname }}.cfg"
       register: __sap_hana_install_register_hdblcm_output
       changed_when: no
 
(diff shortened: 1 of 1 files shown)


sean-freeman:
PR has been reviewed and rejected, each of these Ansible Tasks are run on the remote node / are run independently. The dynamically generated config file is generated from the Jinja file inside of the Ansible Role, and pushed to `sap_hana_install_configfile_directory`. As shown in the Ansible Role defaults file, this directory is on the remote/target/managed node/s:
```
# Directory where to store the hdblcm configfile template and the Jinja2 template:
sap_hana_install_configfile_directory: "{{ sap_hana_install_software_extract_directory }}/configfiles"
```

ranjithrajaram:
The only reason for hitting this problem was when the default values were used and using NFS

Looking at the default values

sap_hana_install_software_directory points to /software/hana [default value]

The path /software/hana is typically a NFS folder in most or some use cases 

We generate the sap_hana_install_software_extract_directory from sap_hana_install_software_directory, so it becomes /software/hana/extracted

For configfiles, it would become /software/hana/extracted/configfiles (sap_hana_install_configfile_directory: "{{ sap_hana_install_software_extract_directory }}/configfiles")

Since /software/hana can be a NFS folder, /software/hana/extracted/configfiles can be the same on both the nodes. Without the PR, installation will fail.  

If not sap_hana_install_software_extract_directory has to be mandatory and a check needs to be done before proceeding

Issue #188, opened 2022-07-25
SAP HANA Install fails XXXX  is not assigned to a local network interface


ranjithrajaram:
While attempting to use sap_hana_install role, installation fails with the following message XXXX is not assigned to a local network interface

While generating the configfile.cfg for hdblcm, it uses the wrong hostname. Either one of the node installation succeeds (second node for which the configfile.cfg was generated)

To resolve this issue, had to apply the above pull request(just filed)
https://github.com/sap-linuxlab/community.sap_install/pull/187

Complete message
---
```
fatal: [dhcp200-237]: FAILED! =>
{
  "changed": true,
  "cmd": [
    "./hdblcm",
    "--ignore=check_diskspace,check_min_mem",
    "--hdbinst_server_xs_engine=off",
    "--hdbinst_server_import_content=off",
    "--configfile=/tmp/ansible.kcutby6thanaconfig/configfile.cfg",
    "-b"
  ],
  "delta": "0:00:05.065610",
  "end": "2022-07-21 11:21:01.940221",
  "msg": "non-zero return code",
  "rc": 1,
  "start": "2022-07-21 11:20:56.874611",
  "stderr_lines": [
    "Running in batch mode",
    "  IP address '10.65.200.241' (host 'dhcp200-241') is not assigned to a local network interface"
  ],
  "stdout_lines": [
    "",
    "",
    "SAP HANA Lifecycle Management - SAP HANA Database 2.00.063.00.1655123455",
    "************************************************************************",
    "",
    "",
    "Scanning software locations...",
    "Detected components:",
    "    SAP HANA Database (2.00.063.00.1655123455) in /software/hana/extracted/SAP_HANA_DATABASE/server",
    "Log file written to '/var/tmp/hdb_DB1_hdblcm_install_2022-07-21_11.20.57/hdblcm.log' on host 'dhcp200-237'."
  ]
}
```
---

sean-freeman:
PR #187  has been reviewed and rejected, each of these Ansible Tasks are run on the remote node / are run independently. The dynamically generated config file is generated from the Jinja file inside of the Ansible Role, and pushed to `sap_hana_install_configfile_directory`. As shown in the Ansible Role defaults file, this directory is on the remote/target/managed node/s:
```
# Directory where to store the hdblcm configfile template and the Jinja2 template:
sap_hana_install_configfile_directory: "{{ sap_hana_install_software_extract_directory }}/configfiles"
```

[...]

PR #189, opened 2022-07-28
sap_hana_install: installation media handling actions when using NFS share

ranjithrajaram:
Requesting review from Markus Koch. Discussed with him

This PR is required to remove the extracted left overs of the SAP Install media on the node. When using with NFS Install media, the installation fails due to issue #188 so we had to define sap_hana_install_software_extract_directory. Noticed that the extracted files were not getting cleared after the SAP HANA install

### DIFF
--- a/roles/sap_hana_install/README.md
+++ b/roles/sap_hana_install/README.md
@@ -81,9 +81,6 @@ If this role is executed on more than one host in parallel and the software extr
 the role will only extract the files on the first host on which the extraction has started. The role will proceed on the other hosts
 after the extraction of SAR files has completed.
 
-If NFS is used for sharing the SAP HANA installation media between the nodes, then it is required to define `sap_hana_install_configfile_directory`. The default for `sap_hana_install_configfile_directory` is  "{{ sap_hana_install_software_extract_directory }}/configfiles". This variable should point to a non nfs path. After installation, if a cleanup of configfile is required, then set `sap_hana_install_cleanup_configfile_directory` as true. If a cleanup of software extract directory is required then set sap_hana_install_cleanup_extract_directory as true. The default value for both these cleanup actins are false
-
-
 - Sample directory `sap_hana_install_software_extract_directory` containing extracted SAP HANA software installation files
     ```console
     [root@hanahost extracted]# ll -lrt
--- a/roles/sap_hana_install/README.md
+++ b/roles/sap_hana_install/README.md
@@ -81,11 +81,11 @@ If this role is executed on more than one host in parallel and the software extr
 the role will only extract the files on the first host on which the extraction has started. The role will proceed on the other hosts
 after the extraction of SAR files has completed.
 
-If NFS is used for sharing the SAP HANA installation media between the nodes, then it is required to define 
-`sap_hana_install_configfile_directory`. The default for `sap_hana_install_configfile_directory` is 
+If NFS is used for sharing the SAP HANA installation media between the nodes, then it is required to define
+`sap_hana_install_configfile_directory`. The default for `sap_hana_install_configfile_directory` is
 "{{ sap_hana_install_software_extract_directory }}/configfiles". This variable should point to a non nfs path. After installation,
 if a cleanup of configfile is required, then set `sap_hana_install_cleanup_configfile_directory` as true. If a cleanup of
-software extract directory is required then set `sap_hana_install_cleanup_extract_directory` as true. The default value for both 
+software extract directory is required then set `sap_hana_install_cleanup_extract_directory` as true. The default value for both
 these cleanup actions are false.
 
 
--- a/roles/sap_hana_install/README.md
+++ b/roles/sap_hana_install/README.md
@@ -81,6 +81,14 @@ If this role is executed on more than one host in parallel and the software extr
 the role will only extract the files on the first host on which the extraction has started. The role will proceed on the other hosts
 after the extraction of SAR files has completed.
 
+If NFS is used for sharing the SAP HANA installation media between the nodes, then it is required to define 
+`sap_hana_install_configfile_directory`. The default for `sap_hana_install_configfile_directory` is 
+"{{ sap_hana_install_software_extract_directory }}/configfiles". This variable should point to a non nfs path. After installation,
+if a cleanup of configfile is required, then set `sap_hana_install_cleanup_configfile_directory` as true. If a cleanup of
+software extract directory is required then set `sap_hana_install_cleanup_extract_directory` as true. The default value for both 
+these cleanup actions are false.
+
+
 - Sample directory `sap_hana_install_software_extract_directory` containing extracted SAP HANA software installation files
     ```console
     [root@hanahost extracted]# ll -lrt
--- a/roles/sap_hana_install/README.md
+++ b/roles/sap_hana_install/README.md
@@ -81,6 +81,9 @@ If this role is executed on more than one host in parallel and the software extr
 the role will only extract the files on the first host on which the extraction has started. The role will proceed on the other hosts
 after the extraction of SAR files has completed.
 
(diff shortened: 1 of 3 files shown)


kksat:
You do not have to check if directory exists before deleting it with file ansible module. File module is idempotent.

rhmk:
Hi Ranjith, 
can you make the change kksat has requested, please. 

In addition I would suggest entering the following into `defaults/main.yml`
```
# set to false if you want to remove temporary files after install
sap_hana_install_cleanup: true
```
Reduce your changes to:
```
- name: SAP HANA Post Install - Deleting software extract directory '{{ sap_hana_install_software_extract_directory }}'
  ansible.builtin.file:
       path: "{{ sap_hana_install_software_extract_directory }}"
       state: absent
when: sap_hana_install_cleanup
```
@kksat, what do you think?

BR
 Markus

rhmk (changes requested):
as discussed

rhmk (changes requested):
Hi,
have a look at your postinstall.yml.
seems there are some strange lines in the file, that do not belong in the file (looks like a patch not like the file)
BR
 Markus

rhmk (commented):
When you cleaned up the file it looks good

kksat:
> In addition I would suggest entering the following into `defaults/main.yml`
> ...
> Reduce your changes to:
> 
> ```
> - name: SAP HANA Post Install - Ensure extract directory '{{ sap_hana_install_software_extract_directory }} is deleted'
>   ansible.builtin.file:
>        path: "{{ sap_hana_install_software_extract_directory }}"
>        state: absent
>   when: sap_hana_install_cleanup
> ```
> 
> @kksat, what do you think?

1. I do not see a need to cleanup directory with extracted files.

2. from my point of view all installation configuration files for SAP HANA should be placed in root user home directory as in SAP example 

<https://help.sap.com/docs/SAP_HANA_PLATFORM/2c1988d620e04368aa4103bf26f17727/1dbba6ac03054d7eb07c819aae47d095.html?version=2.0.03&locale=en-US#loioc907788c8987464097945c5452148ab6>

See last one 

"
./hdblcm --sid=DB1 --configfile=~/configfile1.cfg
"

I would add hostname and HANA System SID to configfile name to avoid conflicts

"
./hdblcm --sid=DB1 --configfile=~/configfile_hostname_SAP HANA SID.cfg
"

This configuration information can be cleaned up, but I do not see any reason to do that. That might be useful information later to understand how SAP HANA was installed with what parameters. 

PS maybe we have to add date and time to configuration file, because installation can be run several times for one SAP HANA system on one host.

~/configfile_hostname_SAP HANA SID_date and time UTC.cfg

3. in regards to race condition when extracting files, I see following code in the 

https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_hana_install/tasks/pre_install/hdblcm_prepare.yml

```ansible
- name: SAP HANA hdblcm prepare - Create status file '{{ sap_hana_install_software_extract_directory }}/__EXTRACTION_ONGOING__'
  ansible.builtin.file:
    path: "{{ sap_hana_install_software_extract_directory }}/__EXTRACTION_ONGOING__"
    state: touch
    mode: '0755'
```

This can be used to avoid race condition. Unfortunately I do not see any code in the role that waits for extraction to finish if there is a conflict (parallel extractions on any action on sap_hana_install_software_extract_directory). Maybe we should enhance that as well.

sean-freeman:
Can I please request a summary of the workflow for this request to handle SAP HANA installation media stored on NFS mount points?

Here is my understanding from the above, is this accurate @ranjithrajaram  ?


> **SAP HANA installation media on non-NFS mount point (e.g. Virtual Disks, iSCSI etc)**

1. Identify files on non-NFS mount path
2. Extract files to non-NFS mount path
3. Generate config file to non-NFS mount path (or specified directory using var `sap_hana_install_configfile_directory`)


> **SAP HANA installation media on NFS mount point**

1. Identify files on NFS share mount path
2. Extract files to the NFS share mount path
3. Generate config file to `non-NFS` mount path (or specified directory using var `sap_hana_install_configfile_directory`)
4. OPTIONAL: Remove config file
5. OPTIONAL: Remove extracted files from NFS share mount path

rhmk:
Hi Sean,

I would summarize the workflow differentky, as Linux doesn't care about 
NFS or non-NFS, it just cares about r/w or non r/w path.

*SAP HANA installation media on read-write filesystem (e.g. NFS, local 
disk (virtual,iscsi,fc ...))*

 1. Identify files on r/w mount path
 2. Extract files to r/w mount path
 3. Generate config file to r/w  mount path (or specified directory
    using var |sap_hana_install_configfile_directory|)

*SAP HANA installation media on read-only filesystem (e.g. NFS, CIFS, ...)*

 1. Identify files on r/o share mount path
 2. Extract files to a r/w mount path (e.g. local disk, or dedicated r/w
    NFS dir (no concurrent write))
 3. Generate config file to |r/w| mount path (or specified directory
    using var |sap_hana_install_configfile_directory|)
 4. OPTIONAL: Remove config file
 5. OPTIONAL: Remove extracted files from r/w mount path

BR

  Markus

**

    **

****

rhmk (approved):

roles/sap_hana_install/tasks/post_install.yml
```diff
@@ -160,6 +160,19 @@
     __sap_hana_install_fact_hana_version: "{{ __sap_hana_install_register_install_result.stdout.split(';')[0] }}"
   when: not ansible_check_mode
 
+
```
berndfinger, line 163: Please remove empty line so the spacing between tasks is similar to other tasks.

roles/sap_hana_install/README.md
```diff
@@ -81,6 +81,9 @@ If this role is executed on more than one host in parallel and the software extr
 the role will only extract the files on the first host on which the extraction has started. The role will proceed on the other hosts
 after the extraction of SAR files has completed.
 
+If NFS is used for sharing the SAP HANA installation media between the nodes, then it is required to define `sap_hana_install_configfile_directory`. The default for `sap_hana_install_configfile_directory` is  "{{ sap_hana_install_software_extract_directory }}/configfiles". This variable should point to a non nfs path. After installation, if a cleanup of configfile is required, then set `sap_hana_install_cleanup_configfile_directory` as true. If a cleanup of software extract directory is required then set sap_hana_install_cleanup_extract_directory as true. The default value for both these cleanup actins are false
```
berndfinger, line 84: Please reduce the length of lines to 160 or less, similar to other lines in the same file.

roles/sap_hana_install/defaults/main.yml
```diff
@@ -11,6 +11,9 @@ sap_hana_install_software_directory: '/software/hana'
 # created again before the extraction starts.
 sap_hana_install_software_extract_directory: "{{ sap_hana_install_software_directory }}/extracted"
 
+# If there is a requirement to cleanup "sap_hana_install_software_extract_directory" after SAP HANA Installation, then set the value to true. By default, this directory will not be removed
```
berndfinger, line 14: line length too long

roles/sap_hana_install/defaults/main.yml
```diff
@@ -49,6 +52,9 @@ sap_hana_install_verify_signature: no
 # Directory where to store the hdblcm configfile template and the Jinja2 template:
 sap_hana_install_configfile_directory: "{{ sap_hana_install_software_extract_directory }}/configfiles"
 
+# If a custom path for sap_hana_install_configfile_directory was defined and if there is a requirement to cleanup this directory, then set "sap_hana_install_cleanup_configfile_directory" as true. Incase if a custom path was not defined and "sap_hana_install_cleanup_extract_directory" was set as true, then the configfiles will be removed.
```
berndfinger, line 55: line length too long

roles/sap_hana_install/defaults/main.yml
```diff
@@ -11,6 +11,10 @@ sap_hana_install_software_directory: '/software/hana'
 # created again before the extraction starts.
 sap_hana_install_software_extract_directory: "{{ sap_hana_install_software_directory }}/extracted"
 
+# If there is a requirement to cleanup "sap_hana_install_software_extract_directory" after SAP HANA Installation, then
```
berndfinger, line 14: Please remove the trailing spaces here and in the other lines in this file (where you split the lines).

rhmk: Ranjith, to make it faster and easier install pre-commit on your system: 
https://sap-linuxlab.github.io/contribute/__install_pre_commit/

PR #190, opened 2022-07-29
sap_swpm: increase flexibility of default mode for SAP AnyDB

sean-freeman:
Initial work to append inifile.params generation for SAP AnyDB

This PR includes breaking changes for the `sap_swpm` Ansible Role when using default mode (or default templates mode), and should be noted in the changelog once merged.

The change impacts the generation of the inifile.params. Currently tested with SAP AnyDB options SAP ASE, SAP MaxDB and IBM DB2.

Further tests are required for the password parameters, as each different SAP AnyDB creates different users and it is not 100% clear in the inifile.params descriptions which user is being referenced.


sean-freeman:
Testing confirmed executions for:
- SAP ASE
- SAP MaxDB
- IBM Db2
- Oracle DB

berndfinger:
lgtm

PR #197, opened 2022-08-30
[DRAFT] [sap_detect_cloud_provider] add role to detect cloud provider

kksat:
add role to detect cloud provider from within the instance running
 currently tested on Azure, AWS and Virtualbox VM.

### DIFF
--- a/roles/sap_detect_cloud_provider/README.md
+++ b/roles/sap_detect_cloud_provider/README.md
@@ -1,0 +1,49 @@
+# sap_detect_cloud_provider
+
+Ansible role to detect where (what cloud provider) host is running in.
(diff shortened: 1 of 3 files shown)


Klaas-:
I think you can use "when" to avoid web calls in many scenarios, so for example the azure test:
https://github.com/sap-linuxlab/community.sap_install/pull/197/files#diff-ca0dca430ef0f7d3f5882c6d9370cc57965b399219806c30f031f88221d56605R5-R13
only makes sense when ansible_virtualization_type is "VirtualPC" -- I don't use AWS but I expect it should say something like XEN there.

sean-freeman:
As per #233 , closing this PR

PR #201, opened 2022-09-12
sap_install_media_detect: overhaul

sean-freeman:
Initial overhaul, append draft install media detection for SAP AnyDB


sean-freeman:
Will need to establish how much overlap with [sap_swpm/tasks/swpm/prepare_software.yml](https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_swpm/tasks/swpm/prepare_software.yml) detection of software (basically the same code)

sean-freeman:
Testing complete, requires review now. Still overlap with `sap_swpm`, requires development code strategy discussions

rhmk:
Can only check with SAPCAR, SWPM; HANA, S4Hana 2020 Core & 2021 Foundation.

Nonetheless here are some first findings:
- It requires to download rar from the internet (no way to bypass to a trusted source or check if already installed)
- It wants to change sapcar binary in place to root:root => not necessary, execution rights are enough - no user change needed

rhmk:
I now run the test-playbook with root permissions and it still fails with:
```
TASK [community.sap_install.sap_install_media_detect : SAP Install Media Detect - SAP HANA - Get all IMDB SAR files in folder - /home/mkoch/Downloads/SAP] ***********************************
fatal: [localhost]: FAILED! => changed=true 
  cmd: |-
    ls IMDB*.SAR
  delta: '0:00:00.003031'
  end: '2022-09-16 09:57:56.364165'
  msg: non-zero return code
  rc: 2
  start: '2022-09-16 09:57:56.361134'
  stderr: 'ls: cannot access ''IMDB*.SAR'': No such file or directory'
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```

rhmk:
From the decription so far, I have expected, that I can give the role a directory where I have downloaded things and it just detects what it has downloaded. I think I misunderstood it. It seems it sorts a non-sorted download directory, so maybe the role name is misleading a bit.
Then I would suggest  to give it a readonly-source and it sorts everything in a target directory 
This role should never require root permissions, as it is not installing anything

rhmk:
How would the role deal with multiple versions of the same software?
The should not fail if some software is not available

sean-freeman:
Merging and raising GH Issues for additional features

Issue #94, opened 2022-03-23
sap_hana_install: hana_exists.yml false detection of existing installations


waorb:
hana_exists.yml checks for directories like `/hana/shared/SID` and `/usr/sap/SID` and fails if those directories aready exist.
This leads to false detections in enviroments where the SID is used as part of the filesystem mountpoint, e.g.

```
# df -h | egrep "HC1|Filesystem"
Filesystem                            Size  Used Avail Use% Mounted on
/dev/mapper/hc1usrsapvg-hc1usersaplv   32G   66M   32G   1% /usr/sap/HC1
/dev/mapper/hc1logvg-hc1loglv          48G  6.4G   42G  14% /hana/log/HC1
/dev/mapper/hc1sharedvg-hc1sharedlv    64G   50G   15G  77% /hana/shared/HC1
/dev/mapper/hc1datavg-hc1datalv        64G   18G   47G  28% /hana/data/HC1
v-srv1:/srv/backup/hana/HC1           2.4T 1021G  1.4T  43% /hana/backup/HC1
```
This is for example needed when you want to install more than one HANA system on a server.

Checking the existence of an HANA installation based on directory names is also somewhat problematic. During a manual installation, one can also override the suggested defaults for various directories. This isn't really recommended, but it is not forbidden either.

An alternative could be to check for the existence of `/usr/sap/hostctrl/exe/saphostctrl` and then use `saphostctrl` to check the already installed instances:

```
# /usr/sap/hostctrl/exe/saphostctrl -function ListInstances
 Inst Info : HC1 - 00 - hc1na1 - 753, patch 625, changelist 2007209
```

berndfinger:
What if we used `./extracted/SAP_HANA_DATABASE/hdblcm --list_systems` from the software directory (e.g. /software/hana)? Then we would not have to rely on saphostctrl. However, we would have to extract the files from the IMBD*.SAR file first not yet done.
BTW I just tested on my freshly installed HANA server:
`# /usr/sap/hostctrl/exe/saphostctrl -function ListInstances`
and got:
`Error: NIECONN_REFUSED (Connection refused), NiRawConnect failed in plugin_fopen()`
using the `root` user and also using the `<sid>adm` user. I have not tried to find out what is causing it but the other method at least appears to be more reliable.

DominikDunst:
We have the same problem with the existing directories when installing HANA on NFS.
[...]

PR #230, opened 2022-11-01
Dev: improved hana_exists

rhmk:
First enhancement towards https://github.com/sap-linuxlab/community.sap_install/issues/94

The improvements:
- uses saphostctrl to detect running and/or conflicting SAP HANA installation
- skips installation if sap_hana_install_check_sidadm_user: false is set and hana is properly installed
- added `sap_hana_install_force: no` default variable. Set to yes to disable checking for hana
- does not fail anymore if directory .../SID exists, but is empty. This check is skipped if saphostctrl is found

Not implemented: 
- hdblcm --list_systems, but the command is available after installation at "/hana/shared/{{ hana_sid }}/hdblcm/hdblcm" 
please let me know if you want this alternative . Can also be detected with find module in /hana/shared

Tests needed: 
- what happens if hostagent is installed, but no HANA
- what happens if other HANA SID or Instance is installed (=> role should fail)

Side effects: added pre-commit  config incl. working .ansible.lint to skip directories that should be checked


ja9fuchs:
@rhmk I think this PR sat here for too long and now has too many conflicts after the recent major updates.
Is it possible to re-create a PR with only the still valid changes?

PR #236, opened 2022-11-09
sap_ha_install_pacemaker: add SBD as fencing method

yeoldegrove:
This PR adds:
- [make stonith-timeout configurable (900 default)](https://github.com/sap-linuxlab/community.sap_install/commit/9a5b3aa522a399fe688f17c1b5699b03f231d294)
- [clarify task names](https://github.com/sap-linuxlab/community.sap_install/commit/e63ca849ff83d7e75e4fab8e0e8985db46ebf0da)
- [rename check_properties to check_status](https://github.com/sap-linuxlab/community.sap_install/commit/1fa88d884aa6680a42172d31ec6b9932e64a8490)
- [add fencing via SBD](https://github.com/sap-linuxlab/community.sap_install/commit/af0642f9e74a4782f7b69673ae4a110ae79904af)

The README is extended to reflect the new feature ans also now gives more context on how to configure the different fencing methods.


yeoldegrove:
@ja9fuchs I understand that you might want to close this for the same reason as #237... and implement this via https://github.com/linux-system-roles/ha_cluster.
`ha_cluster` has SBD implemented in it but is e.g. missing the code to actually load a watchdog kernel module. Also `sap_ha_pacemaker_cluster` would need adapted `meta/argument_specs.yml` to be able to deploy the SBD use case.
Is this something that you're already working on or shall I give it a try?

ja9fuchs:
@yeoldegrove You are correct, SBD can as well be implemented using the `ha_cluster` native parameters. I had not yet closed this PR because I was taking some time to save some of your good input for the stonith resource config on various platforms.

I also want to highlight that your contribution is very much appreciated! 
The deprecation of the original HA roles is not related to your input and is just a case of unfortunate timing and, on our side, a lack of visible clarification at an earlier stage. :)

Regarding the change of `meta/argument_specs.yml`, are you referring to the `ha_cluster` dictionary that is listed here? For that variable you are right, and I have planned to adjust the entry for this argument such that any option the `ha_cluster` role supports can be defined accordingly without causing any argument check issue, plus providing an enhanced example.

Regarding the gap of loading a watchdog kernel module I suggest you try and contribute your code directly to https://github.com/linux-system-roles/ha_cluster.

If you could test a cluster setup with the new `sap_ha_pacemaker_cluster` role, including the functionality you were missing in the deprecated roles, this would be much appreciated. 
Of course, please feel free to add improvements to this role as well, or open GitHub issues to track anything we could follow up on at a later stage. :)

PR #247, opened 2022-11-16
sap_ha_pacemaker_cluster: update argument def ha_cluster

ja9fuchs:
Tiny adjustment to the `ha_cluster` argument definition (native `ha_cluster` role parameter) to  allow any option instead of a limited set.


rhmk (approved):
LFTM

PR #248, opened 2022-11-17
sap*preconfigure: Solve issue 246

berndfinger:
(Prepare the collection for Ansible 2.14, by removing all args: and warn: lines from all tasks which are using the ansible.buitin.command module)


ja9fuchs (changes requested):
Looks good to me - except for the vi swap file oopsie. :)

ja9fuchs (approved):
Looks good now

PR #253, opened 2022-11-18
Fix invalid collection version

ja9fuchs:
'latest' is not a valid version key for galaxy collections and causes dependency resolutions to fail during checks.
Version syntax reference: https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#collections-older-version

Added a valid minimum version. This minimum is required for the ha_cluster role.


berndfinger:
lgtm

PR #255, opened 2022-11-18
deleted roles: ansible-lint config and workflows update

ja9fuchs:
This change 
* removes the deleted HA roles from .ansible-lint exclude config
* deletes the roles' corresponding GitHub workflows

+ adds a new workflow for role `sap_ha_pacemaker_cluster`
+ removes trailing whitespaces from all other workflow files


berndfinger (approved):
lgtm

Issue #246, opened 2022-11-16
Prepare the collection for Ansible 2.14


berndfinger:
For example, the warn: feature is no longer supported for the command module, see https://docs.ansible.com/ansible/latest/collections/ansible/builtin/command_module.html (retrieved 16-Nov-2022). Tasks using this feature in the command module will abort.

berndfinger:
Removed feature list is here: https://github.com/ansible/ansible/blob/v2.14.0/changelogs/CHANGELOG-v2.14.rst#removed-features-previously-deprecated

berndfinger:
The removed feature list contains:
`command/shell - remove deprecated warn module param`
but the https://docs.ansible.com/ansible/latest/collections/ansible/builtin/shell_module.html page still contains the warn parameter.
This is explained here: https://github.com/ansible/ansible/issues/79379#issuecomment-1314228747 , where sivel explained that the correct URL for the Ansible documentation for 2.14 is https://docs.ansible.com/ansible-core/2.14/collections/ansible/builtin/shell_module.html .

berndfinger:
This should help:
`# find . -name "*.yml" -exec awk '/shell:/{a=$1;x=1}/args:/{b=$1;if(x==1){y=1}}/warn:/{if(y==1){print NR, a, b, $0; y=0}}/^$/{x=0}' {} \; -print | awk '/^[0-9]/{print; a=1}/^\./{if (a==1){print; a=0}}'`

PR #268, opened 2022-11-28
preconfigure roles: Prepare for Ansible 2.14

berndfinger:
Solves issue #246, by removing all warn: false arguments from tasks which call the shell and command modules


ja9fuchs (approved):
lgtm

Issue #263, opened 2022-11-23
sap_netweaver_preconfigure: Add link to libstdc++.so.6 in rpath directory other than /usr/lib64


berndfinger:
As per SAP note 3119751, we need to create a link to /usr/sap/lib/libstdc++.so.6 from /opt/rh/SAP/lib64/compat-sap-c++-10.so, for SAP kernels 754 and for SAP kernels 788 and higher (which are built with GCC10).
See also https://bugzilla.redhat.com/show_bug.cgi?id=2143913 . 
For greater flexibility, for example if the executable's rpath will be extended to also search for libraries in /opt/rh/SAP/lib64, the directory in which the link is to be created needs to be configurable by a role variable.

PR #271, opened 2022-11-29
sap_netweaver_preconfigure: Add link to libstdc++.so.6

berndfinger:
Solves issue #263.


rhmk (approved):
Works like a charm and I am ok with merging. Although I would have put the variable `sap_netweaver_preconfigure_rpath`into the var section instead of the defaults section

berndfinger:
Thanks!
BTW We want the variable to be user configurable so we don't have to modify the code after the rpath of future SAP kernels have been updated to include `/opt/rh/SAP/lib64`, that's why I used defaults/main.yml in this case.

Issue #275, opened 2022-12-01
sap_hana_preconfigure: Update RHEL 8 vars file


berndfinger:
See also https://github.com/sap-linuxlab/community.sap_install/pull/274 .

PR #276, opened 2022-12-01
sap_hana_preconfigure: Update RHEL 8 repo vars file

berndfinger:
Solves issue #275.


rhmk (approved):
LGTM -> but I don't like the repo check in this file anyway. I prefer to use a separate role for that

berndfinger:
Sure, why not have a separate role for RHEL repo handling for SAP systems? As soon as we have one, we might remove the repo handling from the preconfigure roles.

Issue #63, opened 2022-02-17
sap_swpm: scrambles /etc/hosts


rhmk:
For some reason the the update of /etc/hosts fails. After running the role, here is how my  /etc/hosts looks like:

10.103.15.16	rha-nw.rha-nw.c.deft-axon-212715.internal	rha-nw
	rha-hana.rha-nw.c.deft-axon-212715.internal	rha-hana

while it has to be 

10.103.15.16	rha-nw.c.deft-axon-212715.internal	rha-nw
10.103.15.61	rha-hana.c.deft-axon-212715.internal	rha-hana

Investigating where this happens ....

rhmk:
The code that is affected is located here: 
roles/sap_swpm/tasks/pre_install/update_etchosts.yml

maybe we can update this with the code from sap_general_preconfigure

rhmk:
first investigation ansible_fqdn is misunderstood -- use of ansible_domain required. Need to rewoek code for use with sap_ip, sap_domain, sap_hostname.

rhmk:
Variables that need to be reviewed and its usage:
sap_swpm_fqdn:
sap_swpm_set_fqdn: "true"
sap_swpm_db_ip:
sap_swpm_db_fqdn:
sap_swpm_db_host:


sap_swpm_update_etchosts:

berndfinger:
Can you skip modifying /etc/hosts in sap_swpm and just do the modification in sap_general_preconfigure? Or should we modify role sap_general_preconfigure to better support NetWeaver installation?

rhmk:
sap_swpm_fqdn is misused and should be sap_swpm_domain

rhmk:
maybe in different places as well

rhmk:
sap_swpm_fqdn is used in two places: 
1. It sets the variable `NW_getFQDN.FQDN` in the configfile which expects the **domain name**
2. it is used to update the /etc/hosts

It needs to be the domain and not the FQDN, so very misleading usage by SAP

Suggested fix:
- Document this in defaults/main.yml above the variable and the demo playbooks
- run several tests in tasks/pre_install/update_etchosts.yml to ensure the variables are set correctly prior to updating /etc/hosts

please comment

berndfinger:
[...]

PR #285, opened 2022-12-06
sap_swpm: Ensure correct /etc/hosts entries

berndfinger:
Solves issue #63.


rhmk (approved):
Still not sure that this works, but the code looks much better than before

PR #303, opened 2023-01-13
simplification to sap_hana_preconfigure

kksat:
This is just a suggestion how to optimise code for sap_hana_preconfigure.

This code is not working, I just want to get maintainers opinion before full fledged code optimisation can be started. 

I've tried to keep task names from sap_hana_preconfigure, so mapping between original and optimised version is easy.

If you have any questions about particular lines of code and how they are working, feel free to ask.


kksat:
@berndfinger feel free to add more reviewers, I've just added ones recommended by github.

berndfinger:
@kksat I like some of your proposals, especially the much more compact way of installing the minimum package versions. Some others won't work. For example, we need to keep the installation and configuration tasks for RedHat and SLES separate. We do not need to touch the repo handling code in this role any more, as we have already more sophisticated repo handling code in sap_general_preconfigure.
Optimizing code which is known to work and which runs quickly enough (I don't think there will be a noticeable performance improvement by using the new code) should not be top priority. But on the other hand, there are really some good ideas in this pull request. Let's meet soon and discuss how and when we could implement it. Some of it could certainly make it into the next version of this role.

Issue #304, opened 2023-01-17
sap_netweaver_preconfigure: Do not create compat-sap-c++ link for s390x on RHEL 8.2 and earlier


berndfinger:
Package compat-sap-c++-10 is not needed and not available for RHEL 8.2 and earlier on hardware platform s390x. So we must not attempt to create a link in file tasks/sapnote/3119751.yml for these systems.

berndfinger:
We could just make the link creation dependent on the existence of the compat-sap-c++-10.so file. Otherwise, we would need to maintain a matrix with the RHEL 8 minor versions and the architecture.

berndfinger:
We could also create the link using the `file` module's `force` parameter but a link pointing to nowhere is nothing an admin should expect from an Ansible role.

PR #305, opened 2023-01-17
sap_netweaver_preconfigure: Solve issue #304

berndfinger:


rhmk:
The change looks good to me, but do not know where the failure is coming from

berndfinger:
> The change looks good to me, but do not know where the failure is coming from
It's explained in the related issue, #304, which says, "Package compat-sap-c++-10 is not needed and not available for RHEL 8.2 and earlier on hardware platform s390x. So we must not attempt to create a link in file tasks/sapnote/3119751.yml for these systems.".

PR #307, opened 2023-01-20
Solve issue 306 (avoid the `nobest` parameter in the `dnf` module)

berndfinger:
Tested with several RHEL 8.1...9.1 versions, on x86_64, ppc64le, and s390x. All systems booted successfully and had the Server software environment group installed afterwards.

### DIFF
--- a/roles/sap_general_preconfigure/tasks/RedHat/installation.yml
+++ b/roles/sap_general_preconfigure/tasks/RedHat/installation.yml
@@ -110,6 +110,9 @@
   when: ansible_distribution_major_version == '7'
 
 # Note: We do not want package updates, see also Red Hat bug 1983749.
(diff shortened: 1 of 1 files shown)


ja9fuchs (approved):
Agreed to return to using the command module for compatibility, while keeping the dnf module task as preparation for later releases.
Both task versions have been tested and are known to work in general.

rhmk:
Why are you using dnf or yum module and not package?  I would prefer  platform independence. Other than that I am fine with the change.

berndfinger:
> Why are you using dnf or yum module and not package? I would prefer platform independence. Other than that I am fine with the change.
We don't need to be platform independent, as the file ` RedHat/installation.yml` is in a platform dependent subdirectory already. But we might still use the package module because it acts as a proxy to the underlying package manager module and should forward all parameters to that one.

PR #319, opened 2023-02-02
sap_swpm: NWAS AAS fix and prettify params

sean-freeman:
sap_swpm: NWAS AAS fix and prettify params (alphabetical, move commented-out params to bottom of each segment etc).


berndfinger:
lgtm

PR #331, opened 2023-02-20
Added installation of tuned profile sap-hana-kvm-guest 

newkit:
When running on KVM, install the tuned profile.


sean-freeman:
We have left this PR too long, we need a decision on whether to merge. Arguably this contribution from @newkit belongs in `sap_vm_preconfigure` scope (but that Ansible Role is experimental/incomplete).

newkit:
Yeah the idea here is to put the tuned profile into the sap tuned packages and only activate that profile in case that we run on KVM. That even could be done in `sap_hana_preconfigure` when KVM is detected.

Issue #321, opened 2023-02-02
sap_netweaver_preconfigure: use meta/argument_specs.yml for all external variables


berndfinger:
See also:
- [specs](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-argument-validation)
- [example](https://github.com/sap-linuxlab/community.sap_install/blob/59e64f7706a1cb5fa633b2453842fba2f992e9bc/roles/sap_ha_pacemaker_cluster/meta/argument_specs.yml)

berndfinger:
The argument_specs mechanism does not work for variables which are set at run time.

We want to be able to set variable defaults, in some cases based on RHEL versions, and then allow for overriding these defaults by users. Example:
```
sap_netweaver_preconfigure_min_swap_space_mb: "{{ __sap_netweaver_preconfigure_min_swap_space_mb }}"
```
So we could use the opportunity to reduce the amount of run time dependent variables to the absolute minimum.

kksat:
Set at runtime you mean during role execution? Opposite to variables that are set by user before role execution starts? 
I believe you can run module https://docs.ansible.com/ansible/latest/collections/ansible/builtin/validate_argument_spec_module.html
to dynamically to validate arguments that are set during role execution. 

Do you have an example where argument_spec files does not work / fails? I never encountered unreliable behaviour. 

Why setting up default values for variables in `defaults/main.yml` is not an option? These variables can be overwritten by users, or dynamically set during role execution. 

For instance put this is in `defaults/main.yml`

```yaml
min_swap_space_mb_mapping:
   "7": 100
    "8": 200
    "9": 300
min_swap_space_mb: "{{ min_swap_space_mb_mapping[( ansible_facts['major_distribution']| default("7") )]}}"
```
You have to add default because at argument validation ansible facts are not available and validation will fail. 

As all above in defaults file, users can change `min_swap_space_mb` variable

```yaml
- name: Include role
   ansible.builtin.include_role:
      name: role_name
   vars:
      min_swap_space_mb: 150
```

berndfinger:
Here is another example, from role `sap_general_preconfigure`:
For RHEL 7, applicable SAP notes are requiring to set two RHEL kernel parameters: https://github.com/sap-linuxlab/community.sap_install/blob/3300ed080a5db494cf57086ce6139838610ccf4c/roles/sap_general_preconfigure/vars/RedHat_7.yml#L145

[...]

PR #341, opened 2023-03-10
sap_netweaver_preconfigure: Use meta/argument_specs.yml

berndfinger:
At the same time, remove OS dependent vars where possible (`sap_netweaver_preconfigure_min_swap_space_mb` in this case).

Solves issue https://github.com/sap-linuxlab/community.sap_install/issues/321.

The role's file README.md will be updated as the next step.

### DIFF
--- a/roles/sap_netweaver_preconfigure/README.md
+++ b/roles/sap_netweaver_preconfigure/README.md
@@ -25,70 +25,97 @@ Please check the SAP NetWeaver installation guide for swap space requirements.
 Do not run this role against an SAP NetWeaver or other production system. The role will enforce a certain configuration on the managed
 node(s), which might not be intended.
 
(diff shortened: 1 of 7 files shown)


ja9fuchs (approved):
Lgtm

rainerleber:
Lgtm - It seams there is no breaking change. But one thing from in my opinion. If documentation says 'default: true' the value shouldn't yes/no.

The Ansible decision is adapting true/false instead of yes/no

berndfinger:
I agree. The next step is to create a new README.md. And I think we should now switch to `true`/`false` for all the roles in this collection.

rhmk (approved):
lgtm

rainerleber (approved):

PR #346, opened 2023-03-17
sap_swpm: Add nw_config_webdisp_generic for WDP installation

slubb:
This PR makes it possible to install the web dispatcher in default mode:

```
 - ansible.builtin.include_vars: ./vars/wdp.sample.yml
      - ansible.builtin.include_role:
          name: community.sap_install.sap_swpm
      vars:
        sap_swpm_ansible_role_mode: default
```

It can use the default `sample-variables-sap-swpm-default-mode-webdisp-install.yml` from the `playbooks/vars` directory.

### DIFF
--- a/roles/sap_swpm/defaults/main.yml
+++ b/roles/sap_swpm/defaults/main.yml
@@ -85,7 +85,7 @@ sap_swpm_inifile_list:
   - nw_config_ports
 #  - nw_config_java_ume
 #  - nw_config_java_feature_template_ids
-#  - nw_config_webdisp_generic
+  - nw_config_webdisp_generic
 #  - nw_config_webdisp_gateway
   - nw_config_host_agent
 #  - nw_config_post_load_abap_reports


sean-freeman:
The Ansible Role default variables will not be altered, they are currently set to a generate an `inifile.params` which will perform an SAP S/4HANA installation by default - but could also be used for SAP S/4HANA System Copy or for SAP NetWeaver + SAP HANA without causing catastrophic failures.

It is documented that the user should configure their own variables, using either this "Default mode" or any of the others "Default Templates mode", "Advanced mode", "Advanced Templates mode", "Reuse inifile mode". The user only needs to append their own variables within an Ansible Playbook and this default is immediately overwritten - see Ansible Vars precedence > https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

This PR cannot be accepted for the above reasoning. If you need further assistance please reach out again

PR #349, opened 2023-04-14
sap_ha_pacemaker_cluster: fixed custom stonith resources list of >1 elements

ja9fuchs:
**Before**
Any custom stonith resources definition list of more than 1 element would result in only the last stonith resource being created due to the resource variable being overwritten, instead of appended to.

**After**
The elements in the custom stonith definition list are appended correctly to a list of resources to be configured.
All defined resources are now actually configured in the cluster.


rhmk (approved):
LGTM

Issue #363, opened 2023-05-02
sap_ha_pacemaker_cluster - tasks/construct_vars_stonith.yml - fencing does not work


waorb:
After the execution of the role, the defined Stonith device looks like: 
```
pcs stonith config res_fence_ibm_powervs
 Resource: res_fence_ibm_powervs (class=stonith type=fence_ibm_powervs)
  Attributes: api-type=private crn=crn:v1:bluemix:public:power-iaas:lon06:a/719e669b325b4d847c6f45401fef8bc1:0e6322f9-1b4c-4489-a5db-a9ddd73c7107:: instance=0e6482f9-1b4c-4489-a5db-a9ddd83c8107 pcmk_host_map="cl-h01-1:542127c0-fc27-4495-a62b-af81dc7bc2e6;cl-h01-2:b1068096-ade4-47e3-89fb-5a0de7091bc0" pcmk_monitor_timeout=600 pcmk_reboot_retries=4 pcmk_reboot_timeout=600 pcmk_status_timeout=60 power_timeout=240 proxy=http://10.30.10.4:3128 region=eu-gb token=**************
  Operations: monitor interval=60s (res_fence_ibm_powervs-monitor-interval-60s)
```

Please note the quotes around the pcmk_host_map string. With this pcmk_host_map string, the fencing operations do not work. 

I use the pcs stonith command to update the stonith device:
```
pcs stonith update res_fence_ibm_powervs pcmk_host_map="wo-cl-h01-1:542127c0-fc27-4495-a62b-af81dc7bc2e6;wo-cl-h01-2:b1068096-ade4-47e3-89fb-5a0de7091bc"
```

Now the device looks like:
```
pcs stonith config res_fence_ibm_powervs
 Resource: res_fence_ibm_powervs (class=stonith type=fence_ibm_powervs)
[...]

PR #367, opened 2023-05-02
Fixing issue #363: removed conflicting quotes from pcmk_host_map construction

ja9fuchs:
Removed explicit quotes from the variable construction task, they were added by mistake.

Tested a custom stonith resource definition using the pcmk_host_map parameter - with and without quotes around the host mapping string. 
Both resulted in correct stonith resource configuration and working fencing.

This fixes issue #363.


sean-freeman (approved):
lgtm 👍

Issue #365, opened 2023-05-02
sap_ha_pacemaker_cluster - SAPHanaTopology shows as promotable resource


waorb:
Next issue observed during the current tests.
The defined SAPHanaTopology resource has the promotable attribute set.

````

Node List:
  * Online: [ cl-h01-1 (1) cl-h01-2 (2) ]

Full List of Resources:
  * res_fence_ibm_powervs	(stonith:fence_ibm_powervs):	 Started cl-h01-1
  * vip_H01_00_primary	(ocf::heartbeat:IPaddr2):	 Started cl-h01-1
  * Clone Set: SAPHanaTopology_H01_00-clone [SAPHanaTopology_H01_00] (promotable):
    * SAPHanaTopology_H01_00	(ocf::heartbeat:SAPHanaTopology):	 Slave cl-h01-1
    * SAPHanaTopology_H01_00	(ocf::heartbeat:SAPHanaTopology):	 Slave cl-h01-2
  * Clone Set: SAPHana_H01_00-clone [SAPHana_H01_00] (promotable):
    * SAPHana_H01_00	(ocf::heartbeat:SAPHana):	 Master cl-h01-1
    * SAPHana_H01_00	(ocf::heartbeat:SAPHana):	 Starting cl-h01-2
````
Here is the resource definition as it is created by the cluster role:
````
# pcs resource config SAPHanaTopology_H01_00-clone
 Clone: SAPHanaTopology_H01_00-clone
  Meta Attrs: clone-max=2 clone-node-max=1 promotable=true
  Resource: SAPHanaTopology_H01_00 (class=ocf provider=heartbeat type=SAPHanaTopology)
   Attributes: InstanceNumber=00 SID=H01
   Operations: methods interval=0s timeout=5 (SAPHanaTopology_H01_00-methods-interval-0s)
               monitor interval=10 timeout=600 (SAPHanaTopology_H01_00-monitor-interval-10)
               reload interval=0s timeout=5 (SAPHanaTopology_H01_00-reload-interval-0s)
               start interval=0s timeout=600 (SAPHanaTopology_H01_00-start-interval-0s)
               stop interval=0s timeout=600 (SAPHanaTopology_H01_00-stop-interval-0s)
````

The code in construct_vars_hana_common.yml actually defines a _**promotable: "no"**_ parameter.
I don't know where this is lost.

````
- name: "SAP HA Prepare Pacemaker - Add resource clone: SAP HANA Topology"
[...]

PR #368, opened 2023-05-03
sap_ha_pacemaker_cluster: fixed issue #365 - resource meta attributes…

ja9fuchs:
Resource config before the fix:

```
Clone: SAPHanaTopology_DB1_00-clone
  Meta Attrs: clone-max=2 clone-node-max=1 promotable=true
```

Resource config after the fix (different SID on different cluster):
```
 Clone: SAPHanaTopology_DB2_00-clone
  Meta Attrs: clone-max=2 clone-node-max=1 interleave=true
```

The cause was the wrong placement of a meta-attribute definition and a bad indentation.


sean-freeman (approved):
lgtm 👍

PR #372, opened 2023-05-03
gh actions: install fixed ansible-compat version to keep compatibility

ja9fuchs:
2 days ago ansible-compat version 4.0.1 has been released. This version is automatically installed by the ansible-lint dependencies, but does not seem to be compatible with ansible-lint versions <6.12.2.

In order to keep 6.8.6 ansible-lint for the time being, this update of the workflows makes sure that ansible-compat version 3.0.2 is explicitly installed.

Tested here:
https://github.com/ja9fuchs/community.sap_install/actions/runs/4874282848/jobs/8695013042

### DIFF
--- a/.github/workflows/ansible-lint-sap_general_preconfigure.yml
+++ b/.github/workflows/ansible-lint-sap_general_preconfigure.yml
@@ -24,15 +24,17 @@ jobs:
 
     steps:
       - name: Check out the code
-        uses: actions/checkout@v2
+        uses: actions/checkout@main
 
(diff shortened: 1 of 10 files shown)


sean-freeman (approved):
I hesitate to lock to repos on `main` branch. However, both are merging to main at same time of a release (checkout could use tag `v3` for latest release and setup-python could use `v4` for latest release). As the main branch matches to latest version releases anyway, I am good with the amendment to the GH Action workflows.

PR #377, opened 2023-05-08
sap_storage_setup: configure local disks, swap and NFS mounts

ja9fuchs:
The role will now configure 
- LVM volumes on unused disk devices (discovered)
- swap on LVM volume *or* a swap file
- default directory structure and pre-installation permissions on NFS share(s), based on the service role(s) the target system is purposed for


sean-freeman (approved):
I have extensively tested this Ansible Role, it is brilliant code

Approved and merging

PR #378, opened 2023-05-09
sap_storage_setup: fixes in NFS task logic and parameter rename

ja9fuchs:
NFS handling for individual fs parameters was not working as expected and always failed back to using global variables instead.
Now the definition is resolved in the correct format and individual settings applied correctly.

The parameter `sap_storage_setup_host_role` has been renamed to `sap_storage_setup_host_purpose`.


sean-freeman (approved):
Tested and approved. Merging

PR #380, opened 2023-05-15
sap_ha_pacemaker_cluster: NetWeaver ASCS/ERS cluster setup

ja9fuchs:
New functionality to configure a pacemaker cluster for a pair of ASCS/ERS instances.

Deployment tested for existing HANA functionality, as well as ASCS/ERS nodes.

To be worked on after merge with `dev`:
- documentation
- default parameter review
- argument_specifications update


sean-freeman (approved):
Tested and reviewed. Agree with further activities to make after merge to `dev`, and will block further releases until these activities are concluded.

PR #383, opened 2023-06-01
sap_storage_setup: updated README and minor improvements

ja9fuchs:
... and a sample playbook for running the storage setup role.


sean-freeman (approved):
Reviewed and tested, approving merge to dev 👍

PR #394, opened 2023-06-21
Update prepare_storage.yml

wuftymerguftyguff:
Allow size of a lv / mount point to be specified and fall back to total size of VG if not specified

### DIFF
--- a/roles/sap_storage_setup/tasks/generic_tasks/prepare_storage.yml
+++ b/roles/sap_storage_setup/tasks/generic_tasks/prepare_storage.yml
@@ -15,7 +15,7 @@
   lvol:
     vg: "{{ item.value.vg }}"
     lv: "{{ item.value.lv }}"
-    size: 100%VG
+    size: "{{ item.value.size  | default('100%VG', true) }}"
   when: 
     - "item.value.numluns == '1'"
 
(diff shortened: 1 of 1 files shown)


wuftymerguftyguff:
Hi,

This scratched my itch to have more control, but preserves default behavior for generic storage.

ja9fuchs:
Hi @wuftymerguftyguff,

Scratching an itch is always welcome. :)

There have been made major changes in this role which are currently available in the 'dev' branch and being tested.
The code that is currently in the 'main' branch has mostly been replaced by the changes.

Please give the role in 'dev' a try and verify if it works for your use-case.
It probably has the same itch. However, if we want to improve it, we should do it in the latest version of the code right away.

I'm closing this PR as we will only merge new code with 'dev' first.

sean-freeman:
Good first issue @wuftymerguftyguff 😃  As said, the SAP Storage setup Ansible Role has just received a major upgrade and is in `dev` branch, for planned next release and I think setting the LVM Logical Volumes to use 100% of the underlying LVM Volume Groups is done.

Would really welcome additional feedback on this new Ansible Role too, it was a major update to that Ansible Role to permit more dynamic and complex setups for either SAP HANA or SAP NetWeaver in different scenarios.

PR #396, opened 2023-06-28
Issue 213

berndfinger:
This change:
- sets `log_mode` to `overwrite` only if the new role parameter `sap_hana_install_log_mode` is set to `overwrite`
- sets `log_mode` to `overwrite` only for the initial tenant database if `sap_hana_install_create_initial_tenant` had been set to `y`.


berndfinger:
We want to go for dev first, hence closing this PR now.

Issue #213, opened 2022-09-30
sap_hana_install: Set log_mode to overwrite always executed for tenant-database, also if sap_hana_install_create_initial_tenant is set to n


DominikDunst:
If sap_hana_install_create_initial_tenant is set to n, the task "SAP HANA Post Install - Set log_mode to overwrite" (post_install/log_mode.yml) fails because the "ALTER SYSTEM ALTER CONFIGURATION"-Command for the tenant could not be executed.
`SQLSTATE: HY000\n* 488: invalid database name: xxx`

berndfinger:
@DominikDunst I could not find a variable with the name `sap_hana_install_create_initial_tenant` in the current dev branch.
Are you referring to the following variable (from defaults/main.yml)?
`# The first tenant database is using a port range not within the range of the ports of additional tenant databases.
# In case this is not desired, you can set the following parameter to `yes` to recreate the initial tenant database.
sap_hana_install_recreate_tenant_database: no
`
Are you using an older version of the role?
In my tests when using default role variables, I never got the error you reported.
If the above does not help, can you please post your role variables here (omitting any customer specific content like e.g. `sap_hana_install_sid`)?

DominikDunst:
@berndfinger: I am reffering to variable "sap_hana_install_create_initial_tenant" from defaults/main.yml in the dev-branche.

```
# If the following variable is undefined or set to `y`, the role will create an initial tenant database.
# corresponding entry in the hdblcm configfile:
# create_initial_tenant, default: 'y'
sap_hana_install_create_initial_tenant: 'y'
```

berndfinger:
OK, thanks. Just a moment...

berndfinger:
OK, when using the defaults for `sap_hana_install_create_initial_tenant`, which is `y`, for installing a SAP HANA rev64 system on RHEL 8.4 on platform x86_64, I am not getting the error you reported. So I'd now like to check the role variables you are using and also the OS level and platform on which you are experiencing the symptom.

DominikDunst:
we set `sap_hana_install_create_initial_tenant=n` because we use the HANA as MDC and create the tenants with the installation of the ABAP-Systems.

berndfinger:
OK, sorry. You had mentioned that already. I will check and find a way to fix this.

[...]

PR #397, opened 2023-06-28
sap_hana_install: Solve issue #213

berndfinger:
This change:

- sets `log_mode` to `overwrite` only if the new role parameter `sap_hana_install_log_mode` is set to `overwrite`
- sets `log_mode` to `overwrite` only for the initial tenant database if `sap_hana_install_create_initial_tenant` had been set to `y`.


sean-freeman (approved):
Agree on approach taken. As this changes default behaviour of the Ansible Role (was always set to overwrite), this fix should be specifically mentioned in Git Tag/Release notes.

rhmk (approved):
LGTM - but have not actually tested

PR #398, opened 2023-07-03
Issue 385: Re-implement rar file renaming; improve rar file handling; add tags

berndfinger:
This is the first part of adding NFS support. No change yet in the logic for moving SAP archive files or for expanding files.


sean-freeman (approved):
LGTM, although two comments:
- `sap_install_media_detect_file_server_only` in defaults, should be set to false? this Ansible Role should assume it is being run on the host which SAP Software is being installed to, only if instructed otherwise should it behave differently _(and there are read-only detection to trigger fail if detect use of file server / remote storage when local storage is expected)_
- `sap_install_media_detect_source_directory` is the new end-user configurable var and replaces `sap_install_media_detect_directory` - this will have a regression impact and require a statement in the update release note.
    - `sap_install_media_detect_directory` has become a dynamically set / Role internal var. I would suggest renaming this to avoid conflict/confusion, something like `__sap_install_media_work_directory`?

rhmk (approved):
I trust Bernd.

Issue #402, opened 2023-07-12
sap_install_media_detect: Initialize certain variables


berndfinger:
Variables which for which elements are added with the `+` operator will be appended if the role is called more than once in a playbook. So they need to be initialized before the first element is added.

berndfinger:
Solved in PR #406

PR #403, opened 2023-07-12
sap_install_media_detect: Initialize fact variables for supporting multiple calls of the role in one playbook

berndfinger:
Solves issue #402.


berndfinger:
Closing; will be covered in the next pull request - the one which also solves #405

Issue #405, opened 2023-07-13
sap_install_media_detect: Also rename any zip file without extension


berndfinger:
We are renaming rar files but we better also rename zip files so that all files have proper endings.

berndfinger:
and also rename and self-extracting rar file (=the first file in a rar series) to .exe.

sean-freeman:
Although it breaks common practice for self-extracting RAR to use `.exe`, would it be "better practice" in our situation to use `.rar.exe` ?

It should have no usable impact even if the file was transferred to a Windows OS - but would be clearer.

berndfinger:
See https://github.com/sap-linuxlab/community.sap_install/issues/408 .

berndfinger:
With #408 and #406, we can close this issue now.

PR #406, opened 2023-07-13
sap_install_media_detect: Initialize fact variables; also rename zip and self-extracting rar files

berndfinger:
Solves issues #402 and #405


roles/sap_install_media_detect/tasks/rename/add_zip_extension_loop_block.yml
```diff
@@ -0,0 +1,15 @@
...
+  register: __sap_install_media_detect_register_zip_file_stat
+
+- name: SAP Install Media Detect - Rename - Loop Block - Fail if '{{ line_item.file }}.zip' already exists
+  ansible.builtin.fail:
+    msg: "File {{ line_item.dir }}/{{ line_item.file }}.zip already exists!"
+  when: __sap_install_media_detect_register_zip_file_stat.stat.exists
+
+- name: SAP Install Media Detect - Rename - Loop Block - Rename '{{ line_item.file }}' to '{{ line_item.file }}.zip'
+  ansible.builtin.command: mv {{ line_item.dir }}/{{ line_item.file }} {{ line_item.dir }}/{{ line_item.file }}.zip
+  changed_when: true
```
kksat, line 15: Do we really want to fail if file already exists? Failing if file exists breaks idempotency in a way. 
May be we should not do anything? Or just save all files that were skipped to report them later at the end of the role execution? 

Parameter creates looks like a way to go to me. See example below

```suggestion
- name: SAP Install Media Detect - Rename - Loop Block - Move '{{ source }}' to '{{ destination }}'
  ansible.builtin.command: mv {{ source }}/{{ destination }}
  vars:
    source: "{{ line_item.dir }}/{{ line_item.file }}"
    destination: "{{ line_item.dir }}/{{ line_item.file }}.zip"
  args:
    creates: "{{ destination }}"
```

berndfinger: Thanks for these hints @kksat.
In the meantime, we have discussed the topic again among the other reviewers and we think we need a different approach: If there are two files with the same prefix but one with the correct suffix and one without suffix (e.g. `51050708_part4.rar` and `51050708_part4`, we behave as follows:
- If these files have the same checksum, we assume that the file without suffix is not required and just remove it unconditionally.
- If the files have a different checksum but are of the file type self-extracting rar (`exe`), `rar`, or `zip`, we use the parameter `sap_install_media_detect_rename_target_file_exists` to determine the behavior:
  - `skip`: Don't do anything - leave both files in place
  - `fail`: The role will simply fail
  - `overwrite`: The file with the correct suffix will be overwritten by the file without suffix. In other words, the file with suffix will have the content of the file without suffix.
- If the files have a different checksum but the file type of the file without extension is not of type self-extracting (`exe`), `rar`, or `zip`, the file will not be part of the list of files to be renamed, so these files will be left in place.

roles/sap_install_media_detect/tasks/prepare/create_file_list.yml
```diff
@@ -41,13 +54,20 @@
   ansible.builtin.debug:
     var: __sap_install_media_detect_fact_files_without_extension_file_results
 
+- name: SAP Install Media Detect - Prepare - Set fact with file names without extension and of type 'Zip archive data'
```
kksat, line 57: From my point of view repetition is not needed here. One can process all the file types in one nested loop.

One of the ways to do so might be:

```ansible
- name: SAP Install Media Detect - Prepare - Set fact with file names without extension for all supported file types 
  ansible.builtin.set_fact:
   __files_without_extension:
     extension: "{{ __files_without_extension[extension] + [filename] }}"
  when: filename.file_type is search filetype_mapping[extension]
  loop: "{{ __sap_install_media_detect_fact_files_without_extension_file_results | product(filetype_mapping.keys() )| list }}" 
  vars:  
     filetype_mapping: # this variable should be in role vars to make it configurable, here just for reference
       zip: 'Zip archive data'
       exe: 'RAR self-extracting archive'
       rar: 'RAR archive data'
     filename: "{{ item.0 }}"
     extension: "{{ item.1 }}"
       

 # __files_without_extension - dictionary of lists
# __files_without_extension['rar'] - list of all rar files
# __files_without_extension['exe'] - list of all exe files
# this should also simplify later loop for any number of extensions, for instance if sar will be added. 
# __files_without_extension should be initialised as 
# __files_without_extension:
#     zip: []
#     exe: []
#.    rar: []
```

Code should "perform" the same as three separate checks, because number of actual checks is still the same. Benefit it is one task. 

To be honest I did not try code above, but it should work from my point of view. If not, it is simpler to remove nested loop and include task file with set_facts in a loop for filetype_mapping

roles/sap_install_media_detect/tasks/prepare/create_file_list.yml
```diff
@@ -29,7 +42,7 @@
 
 - name: SAP Install Media Detect - Prepare - Set fact with the results of the file command
   ansible.builtin.set_fact:
-    __sap_install_media_detect_fact_files_without_extension_file_results: "{{ __sap_install_media_detect_fact_files_without_extension_file_results | d([]) + [__new_dict] }}"
+    __sap_install_media_detect_fact_files_without_extension_file_results: "{{ __sap_install_media_detect_fact_files_without_extension_file_results + [__new_dict] }}"
```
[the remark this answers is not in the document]
berndfinger: @kksat As I have tested the above code and know that it is working and I do not want to spend much time to test a different code, you could maybe do me a favor and provide a link to a document which shows what I could replace here for getting the dictionary added to the variable on the left side?

roles/sap_install_media_detect/tasks/prepare/create_file_list.yml
berndfinger: @kksat Many thanks for these suggestions, but I do not want to spend more time right now on trying to fully understand and re-test the code you provided. I think I will do this - if time permits - if I finished the basic rework of the role.

ja9fuchs (approved):
This fixes a failure of the media role which a test deployment runs into using the `dev` branch, but succeeds now with the `issue-405` code. :+1:

sean-freeman (approved):
lgtm 👍

Issue #353, opened 2023-04-17
sap_swpm - grep returns return code 1 if no match


slubb:
Hi,
I tried to install SAP Content Server 6.5 using the sap_swpm role. 

The role checks for des25 strings in the `inifile.params` and produces an error:

```
TASK [community.sap_install.sap_swpm : SAP SWPM inifile_reuse mode - Check inifile for des25] **********************************************************
fatal: [sapcsdap1]: FAILED! => changed=false
  cmd: set -o pipefail && cat "/tmp/ansible.gnncas99swpmconfig/inifile.params" | grep des25 | wc -l
  delta: '0:00:00.006686'
  end: '2023-04-14 13:09:59.649851'
  msg: non-zero return code
  rc: 1
  start: '2023-04-14 13:09:59.643165'
  stderr: ''
  stderr_lines: <omitted>
  stdout: '0'
  stdout_lines: <omitted>
```

This seems to be caused by the rc of the grep command. If it finds no matches it returns 1. This is the task that checks it:

```
# Check inifile for des25
- name: SAP SWPM inifile_reuse mode - Check inifile for des25
  ansible.builtin.shell: set -o pipefail && cat "{{ sap_swpm_tmpdir.path }}/inifile.params" | grep des25 | wc -l
  register: sap_swpm_inifile_read_file
  changed_when: false
  tags: sap_swpm_generate_inifile
```

Here I demonstrate the issue:
```
[root@sapcsdap1 ~]# set -o pipefail && cat "/tmp/ansible.gnncas99swpmconfig/inifile.params" | grep des25 | wc -l
0
[root@sapcsdap1 ~]# echo $?
1
```

The next task in the role checks on the output of the grep command which is oke. But with nothing found it will not get there.

I created the following fix/workaround by adding the `ignore_errors` flag.:

```
# Check inifile for des25
- name: SAP SWPM inifile_reuse mode - Check inifile for des25
  ansible.builtin.shell: set -o pipefail && cat "{{ sap_swpm_tmpdir.path }}/inifile.params" | grep des25 | wc -l
  register: sap_swpm_inifile_read_file
  changed_when: false
  tags: sap_swpm_generate_inifile
  ignore_errors: True
```

sean-freeman:
@slubb The expected behaviour when using `sap_swpm_ansible_role_mode: inifile_reuse` is to check for `des25` as any reused `inifile.params` will contain hashed passwords.

As an example:
```ini
###############
[...]

PR #415, opened 2023-07-19
sap_swpm: Solve issue #353

berndfinger:
The awk method with a pattern to not look into comments appears to be the most robust method in this case.


sean-freeman (approved):
lgtm 👍 without an integration test I approve merging this minor fix, as it can be proved with a unit test on a plaintext file

PR #423, opened 2023-07-25
fix paths

crysaki:
´sap_hostagent_sar_remote_path´ and ´sap_hostagent_sapcar_remote_path´ was wrong way round.


ja9fuchs (approved):
Good catch! :+1: 

Vielen Dank @crysaki

PR #424, opened 2023-07-25
sap_storage_setup: improved idempotency

ja9fuchs:
The LVM setup modules would be idempotent, but the order in which devices from ansible facts are parsed is not reliable and breaks idempotency.

For now, any used devices are excluded from potential configuration.
If there are no volume/filesystem definitions left due to lack of devices, the local filesystem creation tasks are skipped.


sean-freeman (approved):
I agree with the alteration, this will effectively provide:
> Detect devices, `NEW: if not already used in LVM then` compare to given specification `else skip`, setup LVM PV/VG/LV

The Ansible Role will continue to focus on initialisation of a host's storage, and does not handle changes/expansion to an existing LVM VG/LV, it is beyond the scope.

PR #427, opened 2023-08-03
Solve issue 426

berndfinger:
- add ` -o` as an additional option to the `unar` command for specifying the destination extraction directory 
- remove all of `set -o pipefail` for all shell command sequences which contain `grep -q`
- add ` # noqa risky-shell-pipe` to the affected tasks and add `Reason for noqa: ...` in a comment right before the task


ja9fuchs (approved):
Tested successfully for NW AS ABAP (which failed before the fixes). :+1:

sean-freeman (approved):
Tested successfully with NWAS JAVA and IBM Db2, approved for merge

PR #433, opened 2023-08-04
sap_netweaver_preconfigure: Update RedHat_7.yml

adamaze:
according to SAP Note 2215355, ADS needs some 32 bit packages as well. it seems that this was addressed for RHEL8, but not RHEL7 in this role.


berndfinger:
@adamaze Many thanks for reporting this issue and also the relevant SAP note number.

According to version 7 (the current one, dated 30-Jun-2020) of https://me.sap.com/notes/2135057, the packages are either x86_64 or i686, so we better modify the currently incorrect list in https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_netweaver_preconfigure/vars/RedHat_7.yml to match the SAP note 1:1 and not add x86_64 packages if they are not mentioned in the SAP note.
Could you maybe test this?

adamaze:
It looks like the 64 bit versions of these packages are already system defaults, or installed by something else in this, or another sap system role, so i think simply removing the 64bit packages from this list is the right move. thanks!

berndfinger:
@adamaze I recently got a reply from my SAP contacts that with the current version (version 11, dated 30-Jun-2020) of SAP note https://me.sap.com/notes/2135057, there are no known problems with installing ADS on RHEL 7. My assumption is that one or a combination of the following is the case:
- Maybe ADS has not been used with the latest RHEL 7 version(s)
- Maybe customers had solved the problem with the RHEL 6 packages on their own without notifying
- Maybe customers had been using the RHEL System Roles for SAP or the community.sap_install or redhat.sap_install collection despite the deviation from the SAP note https://me.sap.com/notes/2135057, either without problems or they solved any problems on their own without notifying us. 

Because of this, I will now close this pull request without merging. We could also change the code of the role as per your suggestion but I would not do so without further investigation. Given the fact that [RHEL 7 will reach its Maintenance Support 2 phase](https://access.redhat.com/support/policy/updates/errata#Life_Cycle_Dates) during the next months, it's probably not worth spending more time on investigating this issue further.

berndfinger (commented):
Won't fix, see my [comment](https://github.com/sap-linuxlab/community.sap_install/pull/433#issuecomment-1985848270).

Issue #388, opened 2023-06-12
sap_install_media_detect: `find` module does not support `contains` with `file_type: directory`


berndfinger:
as per https://docs.ansible.com/ansible/latest/collections/ansible/builtin/find_module.html.
The related task is https://github.com/sap-linuxlab/community.sap_install/blob/60f5f40497d54c46b0bd3f79ef1fef4ada148119/roles/sap_install_media_detect/tasks/detect_sapanydb_sapase.yml#L78 .

berndfinger:
A solution would be to replace the following lines in all affected `find` modules (for example):
```
contains: "LINUXX86_64"
```
by:
```
patterns: ".*LINUXX86_64.*"
use_regex: true
```

berndfinger:
Solved by [07641a2](https://github.com/berndfinger/community.sap_install/commit/07641a25be9db562200b5e88a6c34f59382c4c73).

PR #435, opened 2023-08-05
sap_install_media_detect: use 'patterns' for find

berndfinger:
... instead of contains.
Also use regex patterns, and use only true/false for booleans in the find tasks.
Solves issue #388


sean-freeman (approved):
lgtm

PR #437, opened 2023-08-09
sap_install_media_detect: Implement remote_dir

sean-freeman:
The mechanism is as follows:
- If the variable 'sap_install_media_detect_target_directory' is definied and if this directory is writable, then the variable sap_install_media_detect_source is set to 'remote_dir', and this directory is used for copying and extracting files and also for moving the files as the last step.
- In all cases, as the first step, we create the required destination directories on the local, writable file system.
- Then we are creating a list of files for the product by checking the source directory.
- In case of remote_dir, we are then copying the identified files to the local directory.
- All remaining steps (extracting, moving and populating the final list of files) are done on the local directory.

### DIFF
--- a/roles/sap_install_media_detect/tasks/detect_backup_saphana.yml
+++ b/roles/sap_install_media_detect/tasks/detect_backup_saphana.yml
@@ -2,6 +2,26 @@
 
 - name: Identify SAP HANA Backup files
   ansible.builtin.find:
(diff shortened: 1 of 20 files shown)


sean-freeman:
Closing PR, in favour of an updated detection method (which uses detection of all files at the beginning).

PR #439, opened 2023-08-10
sap_ha_pacemaker_cluster: AWS improvements

ja9fuchs:
**Fixes of stonith resource configuration:**
- line break in pcmk_host_map variables caused fencing not working
- commented out AWS credentials from default stonith definition, it will automatically use the aws cli credentials that are configured anyway for the VIP resource
- cosmetical fix of a duplicate pattern in the resource name

**Added:**
AWS cloud-init services are disabled and stopped.
--> Required to avoid removal of the floating Overlay IP.


sean-freeman (approved):
Agree with fixes, approved for merge

PR #447, opened 2023-09-06
sap_ha_pacemaker_cluster: fix: GCP vip resource force cidr 32

ja9fuchs:
In Google cloud platform the primary instance IP is by default /32 and the VIP must also be /32 to avoid conflicts with the routing.

Adding the cidr_netmask to the IPaddr2 resources with a hard-coded value of 32 for the time being, until further detection functionality has been added to make this dynamic for all platforms.


sean-freeman (approved):
For time being this lgtm 👍 Will need to be addressed further with NIC selection

PR #448, opened 2023-09-07
sap_storage_setup: enh: begin integration of multipathing support

ja9fuchs:
The first steps to also work with multipathed devices.
This is disabled by default.


sean-freeman (approved):
Extensively tested on multiple platforms, with and without multipath. Approved for merge 👍

PR #454, opened 2023-09-18
Issue 385 new detection method

berndfinger:
- At the beginning of the execution of the role, a new tool `sapfile` is pushed to a temporary directory on the managed node.
- Also a package which contains a command for extracting and listing content of files of type `rar` is installed.
- The next step is to check if source and/or target directories exist. If role parameter `sap_install_media_detect_target_directory` is defined, files will later be copied from `sap_install_media_detect_source_directory`. This is the `remote_dir` case.
- If the system on which the `sap_install_media_detect_source_directory` is not writable, the role would normally fail because one or both of the following conditions are not met:
  - The SAPCAR EXE file is not executable.
  - There are one or more zip or rar files without extension.
- In this `remote_dir` case, to make sure the role does not fail, it needs to be run first on the node on which `sap_install_media_detect_source_directory` is writable, with role parameter `sap_install_media_detect_file_server_only` set to `true` so the role will not perform and further file detection activities.
- After the SAPCAR EXE file is executable and there are no more `zip` or `rar` files without extension, the role can be called on a managed node where `sap_install_media_detect_source_directory` is not writable.
- A new list of all files with the correct final file names will then be created, and for each of the files, the SAP file types are determined using the `sapfile` tool, either using the file names or - if this information is not sufficient - from information inside the file.
- We then assert that there is at least (or exactly, depending on the file type) one file available for each of the `sap_install_media_detect_*` parameters. For example, if `sap_install_media_detect_kernel_db` is set to `saphana`, then there must be one SAP Kernel DB dependent file for SAP HANA.
- In case of `remote_dir`, the next step is to copy all files from `sap_install_media_detect_source_directory` to `sap_install_media_detect_target_directory`.
- Then we extract files which are configured in `sapfile` to be extracted, and copy or move files which are configured in `sapfile` to be copied or moved. Certain files like SAPCAR*.EXE and SAP Host Agent will be copied to two different directories.
- Once all necessary files have been extracted and all files are copied or moved to where we want them to be, we are using the Ansible find module to identify the different file types by using file or directory name patterns.
- The last step is to fill all required `sap_swpm` parameters from the result of the previous find step, and display all the variables.


sean-freeman (approved):
This is a complete re-write of the Ansible Role, and I have tested 10+ different scenarios for any regression impacts.

From inspection, the outcomes _(extract files and execute set Ansible Variable values for other Ansible Roles within the sap_install Ansible Collection)_ appear unchanged. Therefore I approve merge to `dev` branch for further integration testing.

PR #458, opened 2023-09-19
sap_ha_pacemaker_cluster: fixes to the VIP address parameters

ja9fuchs:
- strip the (optional) /xx cidr suffix from the IP address parameters
- IP address parameters combined for configuration must correspond to the individual host's host type - this is needed to support defining e.g. HANA and NetWeaver IP parameters in the same config file, but avoid VIP resources to be configured in the wrong cluster


sean-freeman (approved):
I understand the purpose behind these commits. I will need to test separately, but these are relatively benign alterations that improve long-term usability. Merge approved

PR #460, opened 2023-09-20
Issue 385 new detection method - bug fixes

berndfinger:
- fix for incorrect handling of `SIGNATURE.SMF` for HANA SAR files in role `sap_install_media_detect`
- variable name change from `sap_swpm_cd_rdms_path` to `sap_swpm_cd_rdbms_path` in all affected places in the collection
- file `sap_install_media_detect/defaults/main.yml` now with two different variable settings, one for SAP S/4HANA and one for SAP ERP 6.0 EHP8 with SAP HANA database, the former one active (= not commented out)


sean-freeman (approved):
Agree with all changes, including new var for RDBMS Client install media search to create separate execution path for any SAP NetWeaver hosts. Approved and merging

PR #461, opened 2023-09-20
Issue 385 new detection method - further changes

berndfinger:
This PR contains the following changes:
- set all SAP product related vars in `defaults/main.yml` to empty or, in case of boolean, to false
- no longer set vars to defaults in comparisons (because all vars are already set in `defaults/main.yml`)
- add more DB client file handling code, to support cases where only the DB client files are needed and not all the DB installation files
- use different variable names of `sap_swpm_kernel_dependent_file_name_get` for db unspecific and for db specific cases
- ensure the unzip package is present so the `sapfile` utility will work even on systems where the `sap*preconfigure` roles have not been executed before
- replace double quotes by single quotes in string comparisons

### DIFF
--- a/roles/sap_install_media_detect/README.md
+++ b/roles/sap_install_media_detect/README.md
@@ -44,6 +44,7 @@ This role does not depend on any other Ansible Role.
 ## Tags
 
 With the following tags, the role can be called to perform certain activities only:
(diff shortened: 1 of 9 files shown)


sean-freeman (approved):
lgtm 👍

PR #465, opened 2023-09-21
sap_general_preconfigure: Improve SELinux handling

### LINKED ISSUE #355: sap*preconfigure: verify that the roles work by default on selinux disabled systems

### LINKED ISSUE #464: sap_general_preconfigure: Allow more directories to be created and SELinux labeled
The role should be enabled to create and relabel other directories than just `/usr/sap`.

berndfinger:
- Fixes #464: Allow more directories than just `/usr/sap` to be created and SELinux labeled
- Also fixes #355. The role will now skip calling the `lsr/selinux` role if the current SELinux state is `disabled` and a relabeling is requested. It will also display a warning message about this fact. After a reboot, once the SELinux state is `permissive` or `enforcing`, the role can be called again and then perform the SELinux relabeling.

### DIFF
--- a/roles/sap_general_preconfigure/README.md
+++ b/roles/sap_general_preconfigure/README.md
@@ -237,11 +237,27 @@ Can be useful if you want to implement your own reboot handling.<br>
 
 One of the SELinux states to be set on the system.<br>
 
+### sap_general_preconfigure_create_directories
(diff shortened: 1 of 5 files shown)


rhmk (approved):
LGTM

Issue #467, opened 2023-09-22
sap_install_media_detect: Separate the handling of database and database client files


berndfinger:
... for supporting database and app instance systems without copying database **and** database client files

berndfinger:
Solved by #469

PR #469, opened 2023-09-22
sap_install_media_detect: Handle db and db client separately

berndfinger:
... for solving issue #467.


sean-freeman (approved):
Resolves errors observed during multi-host installations. Code is lgtm 👍

PR #471, opened 2023-09-22
sap_ha_pacemaker_cluster: enhanced pre-checks and improved variable handling

ja9fuchs:
- new: pre-flight checks for instance numbers
- new: all parameters are transformed into 'ha_cluster' native variables at play level
- new: the 'ha_cluster' role include now relies on the defined native variables, no more static list of vars in the include task

- simplification: removed required/type entries from argument_specifications that are default anyway
- simplification: changed the output vars to display the used 'ha_cluster' vars in the optional configuration output file

- house-keeping: added prefix to internal vars and pre-defined them in vars/main.yml
- house-keeping: default parameter re-sorting into categories and applied the same sort order to the argument_specifications file


sean-freeman (approved):
lgtm 👍  approved for merge

Issue #470, opened 2023-09-22
sap_install_media_detect: Fix SAPCAR file extraction error for SAP ASE Client


berndfinger:

berndfinger:
Solved by #472

PR #472, opened 2023-09-22
sap_install_media_detect: Fix SAPCAR extraction error for SAP ASE Client

berndfinger:
Solves issue #470 - found no deviation for any SAPCAR file which are to be extracted, with the exception of the SAP ASE client which now is handled properly by the role.
I also added two comments which explain why we handle SAPCAR file extraction differently for different types of SAPCAR files.


sean-freeman (approved):
lgtm 👍  merging

PR #477, opened 2023-09-26
sap_install_media_detect: Make the role (nearly) idempotent

berndfinger:
Solves issue 474. Except that the `*_extracted` directories will remain in place. But these directories will not be searched for SAP archive files.
Note: After running the role with the following four tags, the SAP archive files will be in the same place as before running the role the first time, and no further processing will be done. The directories with pattern `*_extracted` will remain in place.
`sap_install_media_detect_provide_sapfile_utility,sap_install_media_detect_check_directories,sap_install_media_detect_create_file_list_phase_1,sap_install_media_detect_move_files_to_main_directory`
This is useful when the role is run with different settings of `sap_install_media_detect_target_directory` so that the `sap_install_media_detect_source_directory` will contain the same SAP archive files as before the role had been run the first time.


sean-freeman (approved):
Test executed successfully when Ansible Role is executed twice in sequence during SAP SolMan Sandbox, lgtm 👍

Issue #475, opened 2023-09-25
sap_install_media_detect: Recognize more file types with sapfile


berndfinger:
We also want to detect maintenance planner files and certain saplanguage files without extension

berndfinger:
Solved by https://github.com/sap-linuxlab/community.sap_install/pull/480 .

PR #480, opened 2023-09-27
sap_install_media_detect: Also recognize MP Stack XML files and correctly handle VCH AFL files

berndfinger:
Fixes issue #475.


sean-freeman (approved):
lgtm 👍

PR #482, opened 2023-09-28
Fix issue #479

berndfinger:
Also fix the `ansible-lint` `jinja[spacing]` violation next to the `ternary` filter (remove space character before the open parenthesis)


sean-freeman (approved):
lgtm

PR #486, opened 2023-09-29
sap_ha_pacemaker_cluster: improved ASCS/ERS check logic

ja9fuchs:
- using `WaitforStarted` to allow instances to finish starting up
- type change: `sap_ha_pacemaker_cluster_aws_vip_update_rt` must be a string, not a list
- improved and aligned search conditionals


sean-freeman (approved):
lgtm 👍

PR #509, opened 2023-11-04
sap_swpm: variable corrections

sean-freeman:
### sap_swpm: fix ibmdb2 unicode enforce

By default `#NW_getUnicode.isUnicode =` exists in inifile.params output, and defaults to true. In certain IBM Db2 executions, the automatic detection does not work and requires `NW_getUnicode.isUnicode = true` contained in the inifile.params. This value should already be true except for legacy NUC install files, which are no longer provided by SAP, therefore appending this into the default inifile block for IBM Db2 installations.

### sap_swpm: improve var handling for advanced exec

When executing advanced or inifile reuse, variable detection can fail. Added improved handling and consistent output (i.e. header of inifile.params).

### sap_swpm: improve vars for System Copy and Export activities

For future, appended System Copy and Export variables to Jinja template as reference, and fixed DDIC password parameter which was in the incorrect section and could lead to SAP ECC Installation Export file failure to import to the target SAP AnyDB.

### sap_swpm: fix IBM Db2 incorrect var population

Database Schema and Database Connect User for ABAP is default `sap<sapsid>` but previous code incorrectly used `sap<anydbsid>`.

### DIFF
--- a/playbooks/vars/sample-variables-sap-swpm-advanced-mode-s4hana-onehost-install.yml
+++ b/playbooks/vars/sample-variables-sap-swpm-advanced-mode-s4hana-onehost-install.yml
@@ -10,12 +10,14 @@ sap_install_media_detect_kernel: true
 sap_install_media_detect_webdisp: false
 sap_install_media_detect_db: "saphana"
 
+# Manual set critical software paths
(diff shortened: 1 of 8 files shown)


berndfinger (approved):
LGTM

Issue #500, opened 2023-10-25
sap_hana_preconfigure: No longer fail by default if the RHEL version is not in the list of supported RHEL versions


berndfinger:
Due to (necessary) delays needed for QA, the list of supported RHEL versions for SAP HANA can typically not be updated quickly enough after the SAP HANA validation has completed and the relevant SAP notes (mainly [2235581](https://me.sap.com/notes/2235581)). Also we should make it easier for customers to test new RHEL versions with SAP HANA.
So let's just display a warning message if the RHEL version is not in the list.

berndfinger:
Solved by #511 .

PR #511, opened 2023-11-06
sap_hana_preconfigure: Be less restrictive with new RHEL versions

berndfinger:
Solves issue #500.

The new behavior if a system is running a RHEL version which is not in the list `sap_hana_preconfigure_supported_rhel_minor_releases` is as follows:

1. role running in **normal** mode:

   a. `sap_hana_preconfigure_min_rhel_release_check: false` (default)
The debug task `Report if the system is running a RHEL release which is supported for SAP HANA` displays a warning message.

    b. `sap_hana_preconfigure_min_rhel_release_check: true`
The assert task `Fail if the system is running a RHEL release which may not be supported for SAP HANA` fails with a failure message.

2. role running in **assert** mode, `sap_hana_preconfigure_assert: true`:

    a. `sap_hana_preconfigure_min_rhel_release_check: false` (default)
The assert task `Report if the RHEL release is supported for SAP HANA` displays a warning message but the warning(error) is always ignored, so the playbook continues.

    b. `sap_hana_preconfigure_min_rhel_release_check: true`
The assert task `Assert that the system is running a RHEL release which is supported for SAP HANA` fails with a failure message.

### DIFF
--- a/roles/sap_hana_preconfigure/README.md
+++ b/roles/sap_hana_preconfigure/README.md
@@ -159,10 +159,11 @@ For the RHEL System Roles for SAP, or for Red Hat Automation Hub, use 'redhat.rh
 
 ### sap_hana_preconfigure_min_rhel_release_check
 - _Type:_ `bool`
-- _Default:_ `true`
+- _Default:_ `false`
 
-Check the RHEL release against a predefined list of known SAP HANA supported RHEL minor releases.<br>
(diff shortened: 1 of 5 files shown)


ja9fuchs (approved):
LGTM :+1:

rhmk (approved):
LGTM

sean-freeman (approved):
lgtm :+1:

Issue #512, opened 2023-11-06
Role sap_ha_pacemaker_cluster and __sap_ha_pacemaker_cluster_vip_resource_id variable


Baboulinet-33:
Hello,

I want to play role `sap_ha_pacemaker_cluster` for ASCS and ERS.
What is the purpose of the variable `__sap_ha_pacemaker_cluster_vip_resource_id` ?
Here link: [__sap_ha_pacemaker_cluster_vip_resource_id](https://github.com/sap-linuxlab/community.sap_install/blob/5b920752aa818fcd9fe96ce740bffaa314baf99e/roles/sap_ha_pacemaker_cluster/tasks/construct_vars_vip_resources_default.yml#L19C39-L19C39)

The playbook failed because the variable is undefined

ja9fuchs:
Thanks for reporting! This bad variable reference slipped through testing after the previous VIP resource logic improvements.

PR #514, opened 2023-11-07
sap_ha_pacemaker_cluster: fix: bad var reference - issue #512

ja9fuchs:
Fixes issue #512 - bad reference from previous VIP resource logic adjustments.


sean-freeman (approved):
lgtm 👍 it is clear it should be the loop item var

Issue #525, opened 2023-11-14
sap_install_media_detect: No longer loop over all sapfile results for removing or creating directories


berndfinger:
This will increase execution speed and reduce the amount of output.

berndfinger:
Solved in #532.

PR #532, opened 2023-11-23
sap_install_media_detect: Don't loop too much

berndfinger:
By defining variables for useful subsets of the sapfile results, we can use those variables for looping, thus limiting the output for several tasks significantly, especially in those cases where there are many SAP archive files in the software directory.

Solves issue #525.


sean-freeman (approved):
lgtm 👍

Issue #507, opened 2023-11-02
collection: EOL of ansible-core 2.13 on November 26, 2023


berndfinger:
So we have to change our `meta/runtime.yml` files to `2.14`.
At the same time, we should also change move from `ansible-lint` version `6.8.6` to the lastest version, which currently is `6.21.1`. Depending on the amount of changes required for `ansible-lint`, we can maybe do these changes in one PR.

berndfinger:
Let's use `2.12` for the community version, to increase the amount of supported releases for control nodes and use `2.14` only in the downstream version which is uploaded to Automation Hub.

berndfinger:
Just FYI - I have used:
```
cd (collection root)
(modified meta/runtime.yml)
find roles -name runtime.yml -exec echo cp meta/runtime.yml {} \; -print
find roles -name runtime.yml -exec cp meta/runtime.yml {} \; -print
find . -name runtime.yml -exec cat {} \; -print
```
to make the changes and verify the correct settings.

berndfinger:
Solved with #540.

PR #533, opened 2023-11-24
collection: Enforce minimum ansible version 2.12

berndfinger:
Solves issue #507.

Note: We are using version 2.12 and not 2.14 to extend the range of possible control nodes to older releases (e.g. RHEL 8.6).


rhmk (approved):
LGTM

Issue #534, opened 2023-11-24
collection: Automatically install required collection(s)


berndfinger:
In the meantime, certain roles require certain other roles from other collections (e.g. [lsr/selinux](https://github.com/linux-system-roles/selinux) or [lsr/ha_cluster](https://github.com/linux-system-roles/ha_cluster)). We have documented a manual dependency resolution mechanism by requesting users to run:
```
ansible-galaxy install -vv -r meta/collection-requirements.yml
```
in each role's `README.md` file.

The roles are also used by users without reading all of the documentation first, and we want to make it as easy as possible to use the software. So we should modify the role and the collection to install all dependencies automatically.

berndfinger:
Resources:
- collection dependencies:
  - https://docs.ansible.com/ansible/latest/collections_guide/collections_installing.html
  - https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_shared.html#listing-collection-dependencies
  - https://docs.ansible.com/ansible/latest/dev_guide/collections_galaxy_meta.html#collections-galaxy-meta
  - https://ansible.readthedocs.io/projects/builder/en/latest/collection_metadata/
- role dependencies:
  - https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#using-role-dependencies

berndfinger:
Thanks @richm . However, we would like to get the collections of dependent roles to be installed when installing the `sap_install` collection.

Specifying dependencies in `requirements.yml` does not cause the dependent collection(s) to be installed. However, when adding the dependent collection name(s) to `galaxy.yml` before creating the collection tar file (the resulting file does not contain `galaxy.yml` BTW), this will cause the dependent collection to be installed, as well as collections which depend on the dependencies.

Sample entry in `galaxy.yml`:
```
dependencies: {"fedora.linux_system_roles": "*"}
```

berndfinger:
Alternative syntax:
```
dependencies:
  fedora.linux_system_roles: "*"
```
Or, in our case, because we want a minimum version:
```
dependencies:
  fedora.linux_system_roles: '>=1.13.0'
```

PR #552, opened 2023-12-13
collection: Automatically install in dependent collection

berndfinger:
Solves issue #534.


berndfinger:
If we automatically install a dependent collection in this upstream version (GitHub, Galaxy), the installation will be performed even on those platforms which do not use roles from the dependent collection.
Modifying this PR to draft.

ja9fuchs (approved):
lgtm

PR #553, opened 2023-12-18
sap_hypervisor_node_preconfigure: Minor fixes

newkit:


berndfinger:
LGTM.
@newkit: What was the reason for the double quotes around `true` in https://github.com/sap-linuxlab/community.sap_install/pull/553/commits/df5b58604a06c83ec48701ded90015d70f5bb7b7 (just for the records)?

newkit:
@berndfinger Simply the fact that the double quotes are needed there in order to work.

Wabri:
@berndfinger it is the definition of a k8s configuration and in this case `feature.node.kubernetes.io/network-sriov.capable` takes a string and not a boolean.

You can find the documentation for this component in here https://docs.openshift.com/container-platform/4.9/networking/hardware_networks/about-sriov.html

And also a configuration example in here https://docs.openshift.com/container-platform/4.8/networking/hardware_networks/configuring-sriov-device.html#nw-sriov-networknodepolicy-object_configuring-sriov-device

berndfinger:
@Wabri Many thanks!

Issue #535, opened 2023-11-24
collection: Cleanup for ansible-lint 6.22.0


berndfinger:
We should prepare for a version bump of ansible-lint.

berndfinger:
I tried to solve:
```
risky-shell-pipe: Shells that use pipes should set the pipefail option.
tasks/RedHat/installation.yml:167 Task/Handler: Create a list of minimum required package versions to be installed
```
by using the package module instead of the shell command sequence for ensuring a minimum package version.
However, when testing the following code on a RHEL 8.8 system with `kernel` `4.18.0-477.10.1.el8_8`, only the package `kernel-core` gets installed to `4.18.0-477.27.1.el8_8` (and the system will boot into `kernel-4.18.0-477.27.1.el8_8`) but neither the package `kernel-modules` nor the package `kernel` will be installed. So we can unfortunately not replace the complex `awk` command sequence of the task mentioned above by the much simpler code:
```
- name: Define a new variable containing package name and version
  ansible.builtin.set_fact:
    __sap_hana_preconfigure_register_minpkglist: "{{ __sap_hana_preconfigure_register_minpkglist | d([]) + [ line_item[0] + ' >= ' + line_item[1] ] }}"
  loop: "{{ __sap_hana_preconfigure_min_pkgs }}"
  loop_control:
    loop_var: line_item

- name: Install minimum packages if required
  ansible.builtin.package:
    name: "{{ __sap_hana_preconfigure_register_minpkglist }}"
    state: present
```

Issue #555, opened 2023-12-19
The conditional check 'XX <= 15' failed. The error was: Conditional is marked as unsafe, and cannot be evaluated.


audiolomb:
### Summary

When running sap_general_preconfigure, getting the error above.

### Issue Type

Bug Report

### Component Name

sap_general_preconfigure

### Ansible Version

```console (paste below)
$ ansible --version

```
ansible [core 2.16.2]
  config file = /home/user/ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.11/site-packages/ansible
  ansible collection location = /home/user/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.11.5 (main, Sep  7 2023, 00:00:00) [GCC 11.4.1 20230605 (Red Hat 11.4.1-2)] (/usr/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True

### Community.sap_install Version

```console (paste below)
$ ansible-galaxy collection list community.sap_install
community.sap_install 1.3.2  

```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```
COLLECTIONS_PATHS(env: ANSIBLE_COLLECTIONS_PATH) = ['/home/user/ansible/collections']
CONFIG_FILE() = /home/user/ansible/ansible.cfg
DEFAULT_ACTION_PLUGIN_PATH(/home/user/ansible/ansible.cfg) = ['/usr/share/ansible/plugins/action']
DEFAULT_FORKS(/home/user/ansible/ansible.cfg) = 50
DEFAULT_REMOTE_USER(/home/user/ansible/ansible.cfg) = ansible
DEFAULT_ROLES_PATH(env: ANSIBLE_ROLES_PATH) = ['/home/user/ansible/roles']
DEFAULT_VAULT_PASSWORD_FILE(/home/user/ansible/ansible.cfg) = /home/user/vault-pass
HOST_KEY_CHECKING(/home/user/ansible/ansible.cfg) = False

### OS / Environment

RHEL 9 (VMware)

### Steps to Reproduce

```yaml (paste below)

```
      - import_role:
          name: community.sap_install.sap_general_preconfigure

VARS:
sap_general_preconfigure_reboot_ok: true
sap_general_preconfigure_selinux_state: disabled
sap_general_preconfigure_modify_selinux_labels: false
[...]

PR #557, opened 2023-12-20
various roles: ansible-lint 6.22.0 cleanup

berndfinger:
Solves issue #535 for the following roles:
- sap_netweaver_preconfigure
- sap_hana_preconfigure
- sap_hana_install
- sap_swpm
- sap_install_media_detect
- sap_hostagent


roles/sap_hostagent/tasks/deploy_bundle.yml
```diff
@@ -1,15 +1,15 @@
...
 
 - name: Install SAPHOSTAGENT
-  command: "{{ sap_hostagent_agent_tmp_directory }}/global/hdb/saphostagent_setup/saphostexec -install"
+  ansible.builtin.command: "{{ sap_hostagent_agent_tmp_directory }}/global/hdb/saphostagent_setup/saphostexec -install"
   register: installagent
   args:
     chdir: "{{ sap_hostagent_agent_tmp_directory }}/global/hdb/saphostagent_setup/"
-  changed_when: "'Install service' in installagent.stdout"
\ No newline at end of file
+  changed_when: "'Install service' in installagent.stdout"
```
rhmk, line 15: Is the double quote irrelevant?

rhmk (approved):
LGTM

roles/sap_hostagent/tasks/deploy_bundle.yml
berndfinger: Probably not. We have the same syntax in all other roles as per output of:
```
$ cd (collection root)
$ grep -rn "changed_when" | grep " in "
```
Also, while working on fixing https://github.com/sap-linuxlab/community.sap_install/issues/555, I got the following error message when removing the double quotes from the `that:` statement in one of the `assert` tasks:
```
  ansible.builtin.assert:
    that: 'bind-utils' in ansible_facts.packages
                       ^ here
This one looks easy to fix. It seems that there is a value started
with a quote, and the YAML parser is expecting to see the line ended
with the same kind of quote. For instance:

    when: "ok" in result.stdout

Could be written as:

   when: '"ok" in result.stdout'

Or equivalently:

   when: "'ok' in result.stdout"
```

rhmk: Thanks for the explanation.

Issue #556, opened 2023-12-20
collection: Ensure compatibility with Ansible 2.16.1


berndfinger:

berndfinger:
See also:
- https://github.com/sap-linuxlab/community.sap_install/issues/555 - relates to: 
  - https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_9.html#playbook - relates to:
    - https://access.redhat.com/security/cve/cve-2023-5764 ([Red Hat bug 2247629](https://bugzilla.redhat.com/show_bug.cgi?id=2247629); [entry on cve.org](https://www.cve.org/CVERecord?id=CVE-2023-5764))

berndfinger:
According to https://github.com/advisories/GHSA-7j69-qfc3-2fq9, the versions of `ansible-core` in which CVE 2023-5764 is fixed are:
- 2.16.1
- 2.15.8
- 2.14.12.
With these versions of `ansible-core`, the roles will fail in assertion tasks which need (but do not have) the fix for issue #556.

For roles not running in *assert* mode, at least the following tasks are affected (`grep -rn "that:" | awk '$1!~/assert-/&&/{{/{print}'` - This command however will not catch multiline `that:` statements):
- [sap_general_preconfigure: "Ensure that the length of the hostname is not longer than 'sap_general_preconfigure_max_hostname_length'"](https://github.com/sap-linuxlab/community.sap_install/blob/5b920752aa818fcd9fe96ce740bffaa314baf99e/roles/sap_general_preconfigure/tasks/RedHat/generic/configure-hostname.yml#L53)
- [sap_hana_preconfigure: Ensure that the system is running a RHEL release which is supported for SAP HANA](https://github.com/sap-linuxlab/community.sap_install/blob/5b920752aa818fcd9fe96ce740bffaa314baf99e/roles/sap_hana_preconfigure/tasks/RedHat/installation.yml#L9)
- [sap_hana_install: SAP HANA Checks - In case there is a group 'sapsys', assert that its group ID is identical to 'sap_hana_install_groupid'](https://github.com/sap-linuxlab/community.sap_install/blob/5b920752aa818fcd9fe96ce740bffaa314baf99e/roles/sap_hana_install/tasks/hana_exists.yml#L145)
[...]

PR #562, opened 2023-12-22
various roles: Fix issue sap-linuxlab#556 for non-assert runs

berndfinger:
This commit makes the roles compatible with the following versions of `ansible-core`:
- 2.16.1
- 2.15.8
- 2.14.12 when running in normal (=non-assert) mode.

The preconfigure roles also support an extended check mode, called assert mode, in which the roles do not change anything but verify if all settings are correct. This commit will not cover the assert mode. So several tasks of the preconfigure roles in assert mode will fail. By using the `_assert_ignore_errors` role parameters in assert mode, the roles will not fail but the affected tasks will not be executed so those settings will not be validated.

Relates to issues #555 and #556.


ja9fuchs (approved):
lgtm

Issue #589, opened 2024-01-12
collection: Assign all non-role specific ("global") variables to role local variables


berndfinger:
We should no longer use "global" variables directly in a role but always assign them to local variables which follow the variable naming scheme, as in:
```
sap_ha_pacemaker_cluster_hana_sid: "{{ sap_hana_sid | d('') }}"
```
Note: When running `ansible-lint` in the **collection root**, we get `var-naming[no-role-prefix]` violations which are not detected when running `ansible-lint` in the **role root**.

berndfinger:
Solved for now in https://github.com/sap-linuxlab/community.sap_install/pull/591.

PR #591, opened 2024-01-12
sap_general_preconfigure: skip var-naming[no-role-prefix]

berndfinger:
... for the three variables sap_hostname, sap_domain, sap_ip.

Those are used in many places in the role and the effort for changing and testing the code is too high given that we plan to replace this part of the role by the new role sap_maintain_etc_hosts.

Relates to #589.


ja9fuchs (approved):
lgtm, no ansible-lint complaints spit out anymore

Issue #566, opened 2023-12-27
sap_ha_pacemaker_cluster - AWS instances detection improvement


marcelmamula:
Current platform detection is checking ansible_system_vendor and ansible_product_name for "amazon" string, but it is not present in every AWS EC2 instance.

Example from my test environment:
m5.large (APP):
```
"ansible_product_name": "m5.large",
"ansible_system_vendor": "Amazon EC2",
```

r4.2xlarge (HDB):
```
"ansible_product_name": "HVM domU",
"ansible_system_vendor": "Xen",
```

https://github.com/sap-linuxlab/community.sap_install/blob/f41c5fe469568a5db084438dde571c69042ddfd0/roles/sap_ha_pacemaker_cluster/tasks/platform/ascertain_platform_type.yml#L47

Solution:
Add verification for ansible_product_version as it contains amazon string, because it is not present anywhere else. Optional fact ansible_bios_version also contains amazon string.
```
when:
    - '"amazon" in ansible_system_vendor | lower
      or "amazon" in ansible_product_name | lower
      or "amazon" in ansible_product_version | lower'
```

Example:
m5.large:
```
"ansible_product_name": "m5.large",
"ansible_product_version": "NA",
"ansible_system_vendor": "Amazon EC2",
```

r4.2xlarge:
```
"ansible_product_name": "HVM domU",
"ansible_product_version": "4.2.amazon",
"ansible_system_vendor": "Xen",
```

Hostvars filtered out for amazon on r4 instance:
```
nw750ahahdb1:~ # cat /tmp/hostvars.json | grep -i Amazon
    "ansible_bios_version": "4.2.amazon",
        "bios_version": "4.2.amazon",
                "device": "fs-redacted.efs.eu-central-1.amazonaws.com:/",
        "product_version": "4.2.amazon",
            "device": "fs-redacted.efs.eu-central-1.amazonaws.com:/",
    "ansible_product_version": "4.2.amazon",
```

marcelmamula:
Pull request #596 created with proposed change.

sean-freeman:
@marcelmamula This code design should not attempt to resolve or workaround data quality issues (it is requirement for Cloud IaaS Certification that certain SMBIOS fields are used by a Cloud Vendor, which are absorbed into Ansible Facts and we're reusing).

In addition, `r4` series is from 2016 and while still SAP Certified - really should not be used. The latest release is `r7`.

[...]

PR #596, opened 2024-01-15
sap_ha_pacemaker_cluster: #566 AWS instances detection improvement

marcelmamula:
Adds proposed detection from Issue #566 to cover for HVM instances with different facts.


marcelmamula:
Commit is already included in new https://github.com/sap-linuxlab/community.sap_install/pull/597 with other fixes.

Issue #600, opened 2024-01-15
collection: No longer use separate yamllint workflows


berndfinger:
`yamllint` is part of `ansible-lint` workflows already, due to `yaml` being part of the `enable_list` in all `.ansible-lint` files.

berndfinger:
Closed with #601.

PR #601, opened 2024-01-15
collection: Remove superfluous yamllint workflows

berndfinger:
Solves issue #600.


sean-freeman (approved):
Agree with approach taken, lgtm 👍

PR #620, opened 2024-01-23
fix(sap_ha_pacemaker_cluster): update with new condition

### LINKED ISSUE #615: sap_ha_pacemaker_cluster: sap_ha_pacemaker_cluster_hacluster_user_password ignored
Current workflow is not accepting variable sap_ha_pacemaker_cluster_hacluster_user_password for HA cluster password because:
1. ha_cluster_hacluster_password is always defined in defaults as ""
2. __sap_ha_pacemaker_cluster_hacluster_user_password i always emptied because ha_cluster_hacluster_password  is always defined.
https://github.com/sap-linuxlab/community.sap_install/blob/5ee5cbe030ad152edfe96567155812f9abcdc25f/roles/sap_ha_pacemaker_cluster/tasks/import_hacluster_vars_from_inventory.yml#L68
3. Construct common vars is always skipped because __sap_ha_pacemaker_cluster_hacluster_user_password  is defined in previous step, so sap_ha_pacemaker_cluster_hacluster_user_password  is never used.
https://github.com/sap-linuxlab/community.sap_install/blob/5ee5cbe030ad152edfe96567155812f9abcdc25f/roles/sap_ha_pacemaker_cluster/tasks/construct_vars_common.yml#L21
4. ha_cluster_hacluster_password is "" because __sap_ha_pacemaker_cluster_hacluster_user_password is defined and empty ""
https://github.com/sap-linuxlab/community.sap_install/blob/5ee5cbe030ad152edfe96567155812f9abcdc25f/roles/sap_ha_pacemaker_cluster/tasks/construct_final_hacluster_vars.yml#L74

Solution: Update import_hacluster_vars file with conditional for empty string for ha_cluster_hacluster_password  from defaults.
```
# ha_cluster_hacluster_password
- name: "SAP HA Prepare Pacemaker - (ha_cluster) Register parameter 'ha_cluster_repos'"
  when:
    - ha_cluster_hacluster_password is defined
    - ha_cluster_hacluster_password | length > 0
  ansible.builtin.set_fact:
    __sap_ha_pacemaker_cluster_hacluster_user_password: "{{ ha_cluster_hacluster_password }}"
  no_log: true  # handle credentials with care
```

Wabri:
Closes #615

### DIFF
--- a/roles/sap_ha_pacemaker_cluster/tasks/import_hacluster_vars_from_inventory.yml
+++ b/roles/sap_ha_pacemaker_cluster/tasks/import_hacluster_vars_from_inventory.yml
@@ -63,7 +63,9 @@
 
 # ha_cluster_hacluster_password
 - name: "SAP HA Prepare Pacemaker - (ha_cluster) Register parameter 'ha_cluster_repos'"
-  when: ha_cluster_hacluster_password is defined
+  when:
+    - ha_cluster_hacluster_password is defined
+    - ha_cluster_hacluster_password | length > 0
   ansible.builtin.set_fact:
     __sap_ha_pacemaker_cluster_hacluster_user_password: "{{ ha_cluster_hacluster_password }}"
   no_log: true  # handle credentials with care


ja9fuchs (changes requested):
I suggest to instead remove the password variables from `defaults/main.yml`. There is no default password and it should not be defined empty by the role either. 
This is a remainder from before the variable import was done through these dedicated tasks for better control.

The conditional won't harm, of course. If it should be kept for sanity reasons, please also fix the older copy-paste typo   referring to "ha_cluster_repos" in the task title, while you're at it. :)

Wabri:
@ja9fuchs do you mean remove this assignation? https://github.com/sap-linuxlab/community.sap_install/blob/a7c348d99a4e707f2766adf1158ce5b0e1153feb/roles/sap_ha_pacemaker_cluster/defaults/main.yml#L99

marcelmamula:
@ja9fuchs @sean-freeman I was going through it along with @Wabri and there are few ways how to solve password issue beside this workaround.

There can be 2 use cases:
1. sap_ha_pacemaker_cluster executed as normal user or new user - No idea about ha_cluster or its variables
2. sap_ha_pacemaker_cluster executed as advanced user - Already have ha_cluster variables ready and imported in wrapper

Problem happens only for first (hopefully bigger) group, because:
1. Import of ha_cluster role/collection imports its defaults and define ha_cluster_hacluster_password: ""
https://github.com/linux-system-roles/ha_cluster/blob/6b53f9d487c0021e3990a46e318566fee1793f1b/defaults/main.yml#L22
2. All futher workflow during import_hacluster_vars_from_inventory.yml will just invalidate all further password handling, replacing user input sap_ha_pacemaker_cluster_hacluster_user_password

Solutions:
1. Workaround from this PR to expect defined ha_cluster_hacluster_password: "" and avoid bad password handling
2. Add assert for sap_ha_pacemaker_cluster_hacluster_user_password and force it, with conditional in import_hacluster_vars_from_inventory.yml to avoid replacement
3. Keep it without assert, but enhance import_hacluster_vars_from_inventory.yml to avoid execution if ha_cluster variables are not provided.

PR #640, opened 2024-02-06
sap_*_preconfigure: disable and stop sapconf when saptune run

### LINKED ISSUE #333: sap*preconfigure: SLES saptune enable & start
Hello,
I have a fresh installation of SLES4SAP SP15 SP4 and there is the sapconfig service active and enabled. This leads that saptune cannot be enabled and started.

## Problem

```
andsse1d01:~ # saptune_check

This is saptune_check v0.2.2.

It verifies if saptune is set up correctly.
Please keep in mind:
 - This tool does not check, if the tuning itself works correctly.
 - Follow the hints from top to down to minimize side effects.


Checking saptune
================
[NOTE] saptune package has version 3.0.2
[WARN] System is in status "degraded". Failed services are: saptune.service     -> Check the cause and reset the state with 'systemctl reset-failed'!
[ OK ] configured saptune version is 3
[FAIL] sapconf.service is active        -> Run 'systemctl stop sapconf.service' or 'saptune service takeover'.
[FAIL] sapconf.service is enabled       -> Run 'systemctl disable sapconf.service' or 'saptune service takeover'.
[FAIL] saptune.service is failed        -> Run 'systemctl start saptune.service', 'saptune service start' or 'saptune service takeover'.
[ OK ] saptune.service is enabled

1 warning(s) have been found.
3 error(s) have been found.
Saptune will not work properly!
```

I have had this problem also on another machine, where I did the following:
```
systemctl stop sapconfig.service
systemctl disable sapconf.service
```
But nevertheless saptune_check was showing same output like above. I have then restarted the saptune service. Same output again. Only a reboot from the whole machine worked.

But the command `saptune service takeover` did the trick without restarting anything.

## Idea
It should be checked if sapconf.service is active and enabled, if yes, do the `saptune service takeover` and then proceed with the tasks like they are.

Wabri:
Closes #333

### DIFF
--- a/roles/sap_hana_preconfigure/tasks/SLES/configuration.yml
+++ b/roles/sap_hana_preconfigure/tasks/SLES/configuration.yml
@@ -2,6 +2,12 @@
 - name: Takover saptune and enable
   when: __sap_hana_preconfigure_run_saptune
   block:
+    - name: Ensure sapconf is stopped and disabled
+      ansible.builtin.systemd:
+        name: sapconf
+        state: stopped
+        enabled: false
+
     - name: Make sure that sapconf and tuned are stopped and disabled
       ansible.builtin.command: "saptune service takeover"
       register: __sap_saptune_takeover
--- a/roles/sap_netweaver_preconfigure/tasks/SLES/configuration.yml
+++ b/roles/sap_netweaver_preconfigure/tasks/SLES/configuration.yml
@@ -2,6 +2,12 @@
 - name: Takover saptune and enable
   when: __sap_netweaver_preconfigure_run_saptune
   block:
+    - name: Ensure sapconf is stopped and disabled
+      ansible.builtin.systemd:
+        name: sapconf
+        state: stopped
+        enabled: false
+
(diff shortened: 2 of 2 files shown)


sean-freeman (approved):
lgtm 👍

sean-freeman:
@Wabri I checked Zuul, I cannot access that job's log either because the error is itself related to the logs :) `POST_FAILURE` appears to mean, the container did not successfully upload the logs back to the CI/CD Pipeline host.

This particular job is a prerequisite for the auto-upload to Ansible Galaxy upon a release. The check is the same each time so, you can look at the job logs [here](https://ansible.softwarefactory-project.io/zuul/build/d1c3cb69eae0494b815a0c9079ebdbce/logs) from a different build run yesterday (after yours) that passed immediately. I tried to re-trigger for this PR, but this check is done via webhook so unless a bit of code is pushed into the PR - it won't trigger again.

Best to ignore it and merge if you're ready to do so IMHO. Your PR contents clearly wouldn't break this check.

Wabri:
Lets ignore it then 💯

PR #651, opened 2024-02-12
fix(.github/workflows): remove github-workflow-test branch

### LINKED ISSUE #650: workflow: remove github-workflow-test branch
The branch github-workflow-test was added by @berndfinger to test the regression problem with the new reusable workflow: https://github.com/sap-linuxlab/community.sap_install/issues/621 the regression is now solved and that branch is no longer required.

There are reference to that branch in every github workflow .github/workflows/ansible-lint-*yml for example here: https://github.com/sap-linuxlab/community.sap_install/blob/main/.github/workflows/ansible-lint-sap_ha_pacemaker_cluster.yml

Wabri:
Closes #650

### DIFF
--- a/.github/workflows/ansible-lint-sap_general_preconfigure.yml
+++ b/.github/workflows/ansible-lint-sap_general_preconfigure.yml
@@ -9,7 +9,6 @@ on:
     branches:
       - main
       - dev
-      - github-workflow-test
     paths:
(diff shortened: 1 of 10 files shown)


berndfinger (approved):
I agree that there should be no need any more to keep the `github-workflow-test` branch in the workflows. The workflows appear to be stable, and also all workflows can be started manually.

PR #652, opened 2024-02-12
fix(ansible-test-sanity): update workflow to run in every pr

Wabri:
Closes #648

PR #653, opened 2024-02-12
sap_general_preconfigure: Use tags for limiting the role scope

berndfinger:
Please see the updates of the `README.md` file for examples for using tags with this role (currently only for RHEL systems).
Solves #342 but leaves the related role parameters (e.g. `sap_general_preconfigure_configuration`, `sap_general_preconfigure_2772999_02`) in place and functional because no deprecation note has been published yet.

### DIFF
--- a/roles/sap_general_preconfigure/tasks/RedHat/configuration.yml
+++ b/roles/sap_general_preconfigure/tasks/RedHat/configuration.yml
@@ -3,6 +3,8 @@
 - name: Configure - List required SAP Notes
   ansible.builtin.debug:
     var: __sap_general_preconfigure_sapnotes_versions | difference([''])
+  tags:
+    - always
 
 - name: Configure - Set directory variables for setting SELinux file contexts
   ansible.builtin.set_fact:
@@ -16,11 +18,19 @@
       target: "{{ line_item }}(/.*)?"
       setype: 'usr_t'
   when: sap_general_preconfigure_modify_selinux_labels
+  tags:
+    - sap_general_preconfigure_3108316_02
+    - sap_general_preconfigure_2772999_02
+    - sap_general_preconfigure_selinux
 
 - name: Configure - Display directory variable
   ansible.builtin.debug:
     var: sap_general_preconfigure_fact_targets_setypes
   when: sap_general_preconfigure_modify_selinux_labels
+  tags:
+    - sap_general_preconfigure_3108316_02
+    - sap_general_preconfigure_2772999_02
+    - sap_general_preconfigure_selinux
 
 - name: Configure - Create directories
   ansible.builtin.file:
@@ -33,6 +43,11 @@
   loop_control:
     loop_var: line_item
   when: sap_general_preconfigure_create_directories or sap_general_preconfigure_modify_selinux_labels
+  tags:
+    - sap_general_preconfigure_create_directories
+    - sap_general_preconfigure_3108316_02
+    - sap_general_preconfigure_2772999_02
+    - sap_general_preconfigure_selinux
 
 - name: Configure - Include configuration actions for required sapnotes
   ansible.builtin.include_tasks:
@@ -41,5 +56,4 @@
   loop_control:
     loop_var: sap_note_line_item
   tags:
-    - sap_general_preconfigure_configuration
-    - sap_general_preconfigure_configuration_all_steps
+    - always
--- a/roles/sap_general_preconfigure/tasks/RedHat/configuration.yml
+++ b/roles/sap_general_preconfigure/tasks/RedHat/configuration.yml
@@ -35,7 +35,8 @@
   when: sap_general_preconfigure_create_directories or sap_general_preconfigure_modify_selinux_labels
 
 - name: Configure - Include configuration actions for required sapnotes
-  ansible.builtin.include_tasks: "sapnote/{{ sap_note_line_item.number }}.yml"
+  ansible.builtin.include_tasks:
+    file: "sapnote/{{ sap_note_line_item.number }}.yml"
   with_items: "{{ __sap_general_preconfigure_sapnotes_versions | difference(['']) }}"
   loop_control:
     loop_var: sap_note_line_item
--- a/roles/sap_general_preconfigure/tasks/RedHat/configuration.yml
+++ b/roles/sap_general_preconfigure/tasks/RedHat/configuration.yml
@@ -39,3 +39,6 @@
   with_items: "{{ __sap_general_preconfigure_sapnotes_versions | difference(['']) }}"
   loop_control:
     loop_var: sap_note_line_item
+  tags:
+    - sap_general_preconfigure_configuration
+    - sap_general_preconfigure_configuration_all_steps
--- a/roles/sap_general_preconfigure/tasks/main.yml
+++ b/roles/sap_general_preconfigure/tasks/main.yml
@@ -3,6 +3,9 @@
 - name: Display the role path
   ansible.builtin.debug:
     var: role_path
+  tags:
+    - sap_general_preconfigure_installation
+    - sap_general_preconfigure_configuration
 
 - name: Include OS specific vars, specific
   ansible.builtin.include_vars: '{{ item }}'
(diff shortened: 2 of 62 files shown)


roles/sap_general_preconfigure/tasks/RedHat/configuration.yml
```diff
@@ -35,7 +35,11 @@
...
 - name: Configure - Include configuration actions for required sapnotes
-  ansible.builtin.include_tasks: "sapnote/{{ sap_note_line_item.number }}.yml"
+  ansible.builtin.include_tasks:
+    file: "sapnote/{{ sap_note_line_item.number }}.yml"
   with_items: "{{ __sap_general_preconfigure_sapnotes_versions | difference(['']) }}"
   loop_control:
     loop_var: sap_note_line_item
+  tags:
+    - sap_general_preconfigure_configuration
+    - sap_general_preconfigure_configuration_all_steps
```
Wabri, line 45: Is the tag `sap_general_preconfigure_configuration_all_steps` really necessary?

roles/sap_general_preconfigure/tasks/main.yml
```diff
@@ -52,45 +61,75 @@
...
   ansible.builtin.package_facts:
+  tags:
+    - sap_general_preconfigure_installation
 
 - name: Include tasks from 'installation.yml'
-  ansible.builtin.include_tasks: '{{ item }}/{{ __sap_general_preconfigure_fact_assert_filename_prefix }}installation.yml'
+  ansible.builtin.include_tasks:
+    file: '{{ item }}/{{ __sap_general_preconfigure_fact_assert_filename_prefix }}installation.yml'
+    apply:
+      tags: sap_general_preconfigure_installation
```
Wabri, line 102: There is already a tag in this task, why adding the same one?

roles/sap_general_preconfigure/tasks/main.yml
```diff
@@ -52,45 +61,75 @@
...
+  tags:
+    - sap_general_preconfigure_installation
+    - sap_general_preconfigure_configuration
 
 - name: Include tasks from 'configuration.yml'
-  ansible.builtin.include_tasks: '{{ item }}/{{ __sap_general_preconfigure_fact_assert_filename_prefix }}configuration.yml'
+  ansible.builtin.include_tasks:
+    file: '{{ item }}/{{ __sap_general_preconfigure_fact_assert_filename_prefix }}configuration.yml'
+    apply:
+      tags: sap_general_preconfigure_configuration_all_steps
```
Wabri, line 121: Same question done in the line 102

roles/sap_general_preconfigure/tasks/RedHat/configuration.yml
berndfinger: > Is the tag `sap_general_preconfigure_configuration_all_steps` really necessary?

Thanks! I tested again and it found no longer a combination of tags which would require both tags. So it seems it's really not required. I'll update the code accordingly.

roles/sap_general_preconfigure/tasks/main.yml
berndfinger: This code will apply a tag to all tasks of the included file(s). And without the tags: line below, in the same task, this current task will not be run when using the tag. See also: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_tasks_module.html .

roles/sap_general_preconfigure/tasks/main.yml
berndfinger: See above.

roles/sap_general_preconfigure/tasks/RedHat/configuration.yml
berndfinger: I was a bit too fast. Further testing revealed that only with both tags, `sap_general_preconfigure_configuration` and  `sap_general_preconfigure_configuration_all_steps`, it is possible to execute just one or a few of the configuration steps.
With just the tag `sap_general_preconfigure_configuration`, it is possible to run all configuration steps and skip some of them. But I want more. ;-)
See also the examples in README.md.

roles/sap_general_preconfigure/tasks/main.yml
rhmk: LGTM

rhmk (approved):
The changes look good to me and really improve usability

Wabri (approved):
LGTM

berndfinger:
After discussing an alternative solution with a colleague (for avoiding two parameters when running all or most configuration steps), I will test another approach. Moving to draft now.

berndfinger:
This version is now easier to use and eliminates the need for specifying two tags when running all configuration steps. The only little disadvantage is that now, when using the tag `sap_general_preconfigure_configuration` only, also some header lines of configuration tasks (displaying the SAP note numbers and versions) are displayed. But as using the tags should not be a default, I think this can be accepted.
See the additional lines in `README.md` for examples. Thanks to @Wabri and @ja9fuchs for triggering this change!

berndfinger:
BTW We now know that https://github.com/sap-linuxlab/community.sap_install/pull/652 (ansible-test-sanity for every PR) works as expected. Run time was 1m 32s. Thanks @Wabri !

rainerleber:
LGTM. This PR is a really cool improvement.

rainerleber (approved):

roles/sap_general_preconfigure/tasks/sapnote/2002167.yml
ja9fuchs, line 1: Would it make sense to apply a tag of `sap_general_preconfigure_2002167` to all the imported task files of this specific note? 
As I understand it right now, applying only all tasks for 2002167 requires to list 5 tags. Is that intended?

roles/sap_general_preconfigure/tasks/sapnote/2772999.yml
ja9fuchs, line 1: Same as above, what is the chance of a use-case to only apply all steps related to note 277999? If this is a realistic scenario I suggest to add a tag for the entire note to all the related imported tasks.

roles/sap_general_preconfigure/tasks/sapnote/assert-2772999.yml
```diff
@@ -4,39 +4,59 @@
...
   ansible.builtin.import_tasks: 2772999/08-assert-linux-kernel-parameters.yml
   when: sap_general_preconfigure_config_all | d(true) or sap_general_preconfigure_2772999_08 | d(false)
+  tags:
+    - sap_general_preconfigure_2772999_08
 
 - name: Import tasks from '2772999/09-assert-process-resource-limits.yml'
   ansible.builtin.import_tasks: 2772999/09-assert-process-resource-limits.yml
   when: sap_general_preconfigure_config_all | d(true) or sap_general_preconfigure_2772999_09 | d(false)
+  tags:
+    - sap_general_preconfigure_nproc_limits
```
ja9fuchs, line 56: I'm a little bit confused here by the inconsistency. Why not (also) add a tag `sap_general_preconfigure_2772999_09` to this task, similar to the others? And why are there no "theme" tags like `..._kernel_parameters`, etc., here, similar to the other note tasks?

ja9fuchs (changes requested):

roles/sap_general_preconfigure/tasks/sapnote/2002167.yml
berndfinger: Thanks for the idea! With this mechanism we can simplify implementing just one of the notes! I will implement and test and then update this PR.

roles/sap_general_preconfigure/tasks/sapnote/2772999.yml
berndfinger: (See above)

roles/sap_general_preconfigure/tasks/sapnote/assert-2772999.yml
berndfinger: This is a coding error which I need to fix.

berndfinger: @ja9fuchs BTW The coding error only affects line 56. Regarding the "theme" tags, I am defining those in the files in which the low level files are included, so the tags are in the import tasks of the same name. This should make it easier to understand the code and the related tags.

ja9fuchs (approved):
Great stuff. :+1:

PR #657, opened 2024-02-13
feat(workflows): create a workflow to open issue when outdate deps

Wabri:
Refs https://github.com/sap-linuxlab/community.sap_install/discussions/421#discussioncomment-8457445

### DIFF
--- a/.github/workflows/check_outdated_dependencies.yml
+++ b/.github/workflows/check_outdated_dependencies.yml
@@ -1,0 +1,20 @@
+---
+
+name: Check outdated dependencies
+
+on: workflow_dispatch
+
+jobs:
+  check_outdate_dependencies:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out the code
+        uses: actions/checkout@main
+
+      - name: Check workflow dependencies 
+        uses: ./workflows/check_outdate_deps
+        env:
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ secrets.TOKEN }}
+          REQUIREMENT_FILE: "requirements-workflow.txt"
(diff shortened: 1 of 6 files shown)


berndfinger (approved):
LGTM!

rainerleber:
Thanks for this PR :-) as of this announcement https://forum.ansible.com/t/devtools-projects-transitioning-to-calver-releases/3444 the dev tools are migrated to caller instead of server maybe we should test this with the new version layout as well ?

Wabri:
The actual version we are using are:

```
ansible==9.1.0
ansible-compat==4.1.10
ansible-core==2.16.2
ansible-lint==6.22.1
```

With the CalVer Releases the workflow not change and the reason is for this `       delta=$(( ${version1_splitted[i]} - ${version2_splitted[i]} ))` that you can find [here](https://github.com/Wabri/community.sap_install/blob/5b2fa12a222b244e4afb103d8b015b52fcd94d2b/workflows/check_outdate_deps/check_outdate_deps.sh#L15). The delta is the distance between the current version and the latest and if this value is less than 0 the package must be update.

We should not worry about this new versioning 😃

Wabri:
This is not better than dependabot, is simply different and we can manage by ourself if we need some kind of specific feature.

This is https://github.com/Wabri/community.sap_install/pull/20 an example of PR that dependabot open, if we choose that is the one we want we can go for it.

I've open the dependabot pr to choose what is the best solution for us #659

ja9fuchs (approved):
This is a great idea to get notified of such updates.
I like that it opens an issue and allows us to follow up later or ignore a change. Especially useful to first review the effect of frequent changes with impact like ansible-lint updates across the entire collection.

berndfinger (approved):
LGTM (however, I didn't test).

rhmk (approved):
LGTM

Wabri:
To test this, we should merge to main. is it ok for all of you?

PR #659, opened 2024-02-14
feat(dependabot): create dependabot configuration

Wabri:
Refs https://github.com/sap-linuxlab/community.sap_install/discussions/421#discussioncomment-8457445

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -25,10 +25,7 @@ jobs:
 
       - name: Install test dependencies
         run: |
-          pip3 install ansible==9.1.0
-          pip3 install ansible-compat==4.1.10
-          pip3 install ansible-core==2.16.2
-          pip3 install ansible-lint==6.22.1
+          pip3 install -r requirements-workflow.txt
 
 #      - name: Install collection dependencies
 #        run: ansible-galaxy collection install community.general
--- a/requirements-workflow.txt
+++ b/requirements-workflow.txt
@@ -1,0 +1,4 @@
+ansible==9.1.0
+ansible-compat==4.1.10
+ansible-core==2.16.2
+ansible-lint==6.22.1


kksat (approved):
This is much better than other PR from my point of view.

rhmk:
Is this an alternative to creating the issue or is it complementing it?

Wabri:
No this is an alternative solution to the pr#657, the main difference are:
- pr#657 can be maintain by us with feature provided by us
- pr#657 can now open only issues, but I can improve to open pr
- this pr is about dependabot that is a github product that open pr only without issues
I think that there are no best solution, but dependabot is the one that we can leave as it is and works because is maintain by github itself.

berndfinger:
@Wabri I prefer having issues opened instead of PRs, so this appears to be an advantage over dependabot. Also the feature of creating an issue only once, even if an issue has been closed without solving the dependency, is very useful. It's because we may decide in some cases to stick to a specific outdated version longer.

@kksat Are aware of any advantages of dependabot we have not covered yet in this issue?

berndfinger:
So far, I found https://github.com/renovatebot/renovate/commit/5960b5339ac485796ac8d05ad2ad9b037f0fbc5b as comparison between dependabot and renovatebot.

kksat:
@berndfinger I see opening issues instead of PRs as disadvantage. If PR is opened to update version of ansible-lint, and ansible-lint is running as CI. One can immediately see all the new failures and estimate how long will it take to fix them. 
With issues, just more work to manually create PR, same work to fix all the issues anyway. 
I have really good experience with dependabot, it is really a pleasure merging automatically created PRs.

ja9fuchs:
Getting a PR opened for every dependency change will most likely cause a lot of overhead for role maintainers. Considering the frequency at which ansible-lint changes and the effect it usually has, it would be annoying to increase the frequency at which we have to work through all roles' new lint complaints. 

I suggest we rather let it open issues, test the impact on the collection and then plan the related work ahead.

rainerleber:
Both approaches are fine. But I think we should use the dependabot over a script because of the following reasons.

- Openening a PR for new dependencies is a good practices. The maintainers should always know the latest versions of e.g. ansible-lint and make sure everything works with newer versions.
- Ignoring new Versions could be achieved by closing the pr as well.
- We should more likely define workflows which runs automatically on new PRs to test with the new dependency. If there are no errors the PR could merged either automatically or manually without any issue.
- The script and the container where the script runs, must be maintained as well. The Bot is maintained by GitHub.
- I think the integration effort for a created issue to raise new Versions is more than just have a PR where all packages are updated.

There are way more reasons for the dependeabot instead this script but I think my preference is clear :-)

berndfinger:
I really do not want to have many PRs opened by dependabot. We will probably have more dependent tools in GitHub actions over the next weeks and months and getting too many PRs just for this purpose will delay our development. We better focus on the real work and do a version bump of dependent tools for example every 3 or 6 months.

I think we could give dependabot a try if:
- we let it run just once per week
- if we only run it against the `dev` tree (similar to https://github.com/sap-linuxlab/community.sap_install/blob/7c36fea31c695f4a5443bf806d71e613275c4170/.github/dependabot.yml#L10), so it seems that it easyly possible
- all changes go into one PR which is always updated using the latest `dev` tree until we finally (e.g. after 3 or 6 months) merge it into the `dev` tree

@Wabri Do you know if/think the above is possible?

Wabri:
@berndfinger if the question is: could dependabot use only one pr to do the dependency update request? the answer is no. Dependabot will open one pr for every outdated package, I dont know how it acts when it open a pr that remain pending for a long time and a new version cames out. I'll investigate more about this.

If we want to try dependabot we can merge this pr and let it run for a while, I dont think it will be a problem if it open a couple of pr.

berndfinger:
> @berndfinger if the question is: could dependabot use only one pr to do the dependency update request? the answer is no. Dependabot will open one pr for every outdated package, I dont know how it acts when it open a pr that remain pending for a long time and a new version cames out. I'll investigate more about this.
> 
> If we want to try dependabot we can merge this pr and let it run for a while, I dont think it will be a problem if it open a couple of pr.

I agree - We could start trying out. However, if we already know that we will get many PRs, we have created some unnecessary additional work. So we better spend a little bit more time to get it right initially as much as possible.

In the meantime, I found in the [dependabot documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file) that we can use the parameter [open-pull-requests-limit](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit) to limit the amount of open PRs. If we set this limit to two, we would quickly see after the first PR that a new dependency update has been detected and then we could either close it and wait for the next one to be opened, or if we think it's critical start working on both.

It seems however that after a PR is closed without merging, dependabot will not include the previously detected updates in the consecutive runs. I found a related discussion [here](https://github.com/dependabot/dependabot-core/issues/2980) - the original post also mentioned a PR limit of 2. But it seems this problem has not yet been solved.

Wabri:
Update with the PR limit 👍

Wabri:
Dependabot cannot be trigger manually, I will do more tests to find a solution!

PR #660, opened 2024-02-16
sap_general_preconfigure: Use the role sap_maintain_etc_hosts - RHEL systems

berndfinger:
We want to use a central role for all `/etc/hosts` related manipulation (it is part of this collection already), so let's start with `sap_general_preconfigure`.

At the same time, we can use new role variables which include the role prefix, replacing the previous "global" variables `sap_ip`, `sap_hostname`, and `sap_domain` in `defaults/main.yml` (while still being compatible).

I expect merge conflicts with https://github.com/sap-linuxlab/community.sap_install/pull/653 but it should not be too difficult to solve them, just annoying.


rhmk (approved):
lgtm

rainerleber (approved):
lgtm - in the next step we should do the same for sles

Issue #663, opened 2024-02-22
sap_swpm: prepare_software changes ownership of files every time it runs


rnozaki:
In the task prepare_software.yml there are 3 subtasks that will run to change the ownership of the SAP media directory and files, every time it runs.

This adds extra time on the execution (in one customer the media is from NFS, so it takes 15+ minutes more to run).

When running the automation multiple times on the same servers, or also on different servers, it gets to a point where we know that ownership was already set.

Can we make these subtasks optional? So if the user knows that the ownership was already validated/adjusted, they can run the automation with a flag that will prevent the execution of the redundant ownership change.

berndfinger:
@rnozaki I agree, this behavior needs to be changed.
FYI - somehow similar issue but for role sap_hana_install: https://github.com/sap-linuxlab/community.sap_install/issues/199

rnozaki:
OK, with the PR that I sent, the default behavior would not be changed.

So if someone knows for sure that they do not need to redo the ownership change (after having done it initially), then they can intentionally set the flag to false to avoid the redundant processing.

berndfinger:
Solved in #665.

PR #664, opened 2024-02-22
sap_swpm: Add optional flag to prevent redoing ownership change

rnozaki:
This is a suggestion for issue #663


berndfinger:
@rnozaki Thanks for the PR. I am currently working on a different solution, which would be:
- Change the ownership to `root:root` and permissions **of the directories only**, and of all files with file names starting with **SAPCAR**, to `0755`. Use a role parameter named `sap_swpm_set_directory_sapcar_permissions` to optionally **skip** this step.
- Change the ownership to `root:root` and permissions **of all files only, except SAPCAR** to `0644`. Use a role parameter named `sap_swpm_set_file_permissions` to optionally **skip** this step. If we find out during the testing phase that the permissions of the files have never been wrong, we can skip this step by default and use this role parameter to optionally **enable** this step.

What do you think about this?

rnozaki:
Looks like a viable approach.

We just need to make sure that any ownership change that must be done, be done at least once (initially).

berndfinger:
> We just need to make sure that any ownership change that must be done, be done at least once (initially).

@rnozaki Please have a look at https://github.com/sap-linuxlab/community.sap_install/pull/665 .

rnozaki:
> > We just need to make sure that any ownership change that must be done, be done at least once (initially).
> 
> @rnozaki Please have a look at #665 .

Yes, that looks good, Bernd. That will cover what I requested in issue #663 .

When do you think this will be released?

berndfinger:
> When do you think this will be released?
Now ;-) (as part of dev)

PR #665, opened 2024-02-28
sap_swpm: optionally skip setting file permissions

berndfinger:
This PR supports optionally skipping setting permission and ownership of SAP installation files. It also introduces new role parameters for setting the mode, owner, and group of the software directory, the directory for SAPCAR, and the directory for SWPM, as well as the files within.

Solves issue #663.

### DIFF
--- a/roles/sap_swpm/tasks/swpm/prepare_software.yml
+++ b/roles/sap_swpm/tasks/swpm/prepare_software.yml
@@ -32,14 +32,18 @@
 
     - name: SAP SWPM Pre Install - Create list of absolute directory names from the find result
       ansible.builtin.set_fact:
-        __sap_swpm_fact_directories: "{{ __sap_swpm_fact_directories | d([]) + [item.path] }}"
+        __sap_swpm_fact_directories: "{{ __sap_swpm_fact_directories | d([]) + [line_item.path] }}"
       loop: "{{ __sap_swpm_register_find_result_directories.files }}"
+      loop_control:
+        loop_var: line_item
       when: __sap_swpm_register_find_result_directories is defined
 
     - name: SAP SWPM Pre Install - Create list of absolute file names for non-SAPCAR files from the find result
       ansible.builtin.set_fact:
-        __sap_swpm_fact_files_non_sapcar: "{{ __sap_swpm_fact_files_non_sapcar | d([]) + [item.path] }}"
+        __sap_swpm_fact_files_non_sapcar: "{{ __sap_swpm_fact_files_non_sapcar | d([]) + [line_item.path] }}"
       loop: "{{ __sap_swpm_register_find_result_files_non_sapcar.files }}"
+      loop_control:
+        loop_var: line_item
       when: __sap_swpm_register_find_result_files_non_sapcar is defined
 
     - name: SAP SWPM Pre Install - Ensure correct permissions and ownership of all directories
--- a/roles/sap_swpm/tasks/swpm/prepare_software.yml
+++ b/roles/sap_swpm/tasks/swpm/prepare_software.yml
@@ -32,13 +32,13 @@
 
     - name: SAP SWPM Pre Install - Create list of absolute directory names from the find result
       ansible.builtin.set_fact:
-        __sap_swpm_fact_directories: "{{ __sap_swpm_fact_directories|d([]) + [ item.path ] }}"
+        __sap_swpm_fact_directories: "{{ __sap_swpm_fact_directories | d([]) + [item.path] }}"
       loop: "{{ __sap_swpm_register_find_result_directories.files }}"
       when: __sap_swpm_register_find_result_directories is defined
 
     - name: SAP SWPM Pre Install - Create list of absolute file names for non-SAPCAR files from the find result
       ansible.builtin.set_fact:
-        __sap_swpm_fact_files_non_sapcar: "{{ __sap_swpm_fact_files_non_sapcar|d([]) + [ item.path ] }}"
+        __sap_swpm_fact_files_non_sapcar: "{{ __sap_swpm_fact_files_non_sapcar | d([]) + [item.path] }}"
       loop: "{{ __sap_swpm_register_find_result_files_non_sapcar.files }}"
       when: __sap_swpm_register_find_result_files_non_sapcar is defined
 
--- a/roles/sap_swpm/tasks/swpm/prepare_software.yml
+++ b/roles/sap_swpm/tasks/swpm/prepare_software.yml
@@ -56,7 +56,7 @@
         - __sap_swpm_fact_directories is defined
         - __sap_swpm_register_find_result_directories is defined
 
-    - name: SAP SWPM Pre Install - Create argument list for chmod of non-SAPCAR*EXE files
+    - name: SAP SWPM Pre Install - Create argument list for chown and chmod of non-SAPCAR*EXE files
       ansible.builtin.set_fact:
         __sap_swpm_fact_files_non_sapcar_chown_arg_list: "{{ __sap_swpm_fact_files_non_sapcar | map('quote') | join(' ') }}"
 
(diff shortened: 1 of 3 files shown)


rhmk (approved):
LGTM

berndfinger:
@Wabri Do you want to have a look and comment/approve this PR?

Wabri:
Yes, I'll do it!

roles/sap_swpm/tasks/swpm/prepare_software.yml
```diff
@@ -12,49 +12,105 @@
...
+    - name: SAP SWPM Pre Install - Create argument list for chown and chmod of non-SAPCAR*EXE files
+      ansible.builtin.set_fact:
+        __sap_swpm_fact_files_non_sapcar_chown_arg_list: "{{ __sap_swpm_fact_files_non_sapcar | map('quote') | join(' ') }}"
+
+# Reasons for noqa:
+# - command-instead-of-module: Shorter execution time compared to looping over a list when using the file module
+# - no-changed-when: Not worth checking permissions and ownership before this task and comparing afterwards
+    - name: SAP SWPM Pre Install - Ensure correct permissions and ownership of all non-SAPCAR files # noqa command-instead-of-module no-changed-when
+      ansible.builtin.shell: >
+         set -o pipefail && chown {{ sap_swpm_files_non_sapcar_owner }}:{{ sap_swpm_files_non_sapcar_group }} \
```
Wabri, line 68:
```suggestion
         chown {{ sap_swpm_files_non_sapcar_owner }}:{{ sap_swpm_files_non_sapcar_group }} \
```

There are no pipeline command so the set pipefail is not needed!

roles/sap_swpm/tasks/swpm/prepare_software.yml
```diff
@@ -81,6 +137,15 @@
...
   failed_when: not sap_swpm_sapcar_file_name_stat.stat.exists
 
+- name: SAP SWPM Pre Install - Ensure correct permissions and ownership of the SAPCAR*EXE file
+  ansible.builtin.file:
+    path: "{{ sap_swpm_sapcar_path }}/{{ sap_swpm_sapcar_file_name }}"
+    recurse: no
+    mode: "{{ sap_swpm_files_sapcar_mode }}"
+    owner: "{{ sap_swpm_files_sapcar_owner }}"
+    group: "{{ sap_swpm_files_sapcar_group }}"
+  when: sap_swpm_set_file_permissions
```
Wabri, line 147: I miss the sap_swpm_sapcar_path definition, I guess is somewhere else, but is it not necessary to check that it is defined? or do we take it for granted? (same on task 85)

Wabri (changes requested):

roles/sap_swpm/tasks/swpm/prepare_software.yml
```diff
@@ -12,49 +12,105 @@
...
+    - name: SAP SWPM Pre Install - Ensure correct permissions and ownership of all directories
+      ansible.builtin.file:
+        path: "{{ line_item }}"
+        recurse: no
+        mode: "{{ sap_swpm_software_directory_mode }}"
+        owner: "{{ sap_swpm_software_directory_owner }}"
+        group: "{{ sap_swpm_software_directory_group }}"
+      loop: "{{ __sap_swpm_fact_directories }}"
+      loop_control:
+        loop_var: line_item
```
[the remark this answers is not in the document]
berndfinger: Yes, defining a separate loop var is best practices. I forgot to define it in the previous task. See here:
https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#defining-inner-and-outer-variable-names-with-loop-var and here:
https://github.com/linux-system-roles/selinux/issues/216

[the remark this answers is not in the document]
berndfinger: BTW We could define a loop variable with the role name as a prefix as per https://github.com/linux-system-roles/selinux/commit/389a55eca4128ede751bf317530f90b849cb0e43. However, I did not have any variable name conflict when using line_item, and the main purpose for defining something else than the default of `item` is that there are some Ansible loops which are using `item` implicitly.

roles/sap_swpm/tasks/swpm/prepare_software.yml
berndfinger: From looking at the code, I agree that it doesn't make much sense because there are no pipelines.
If I remember correctly, the task does not fail without pipefail when one of the commands (maybe the first one?) fails (maybe a side effect).
Maybe I should run some more tests and then either:
- document why it is useful, or
- remove it

roles/sap_swpm/tasks/swpm/prepare_software.yml
berndfinger: The parameter `sap_swpm_sapcar_path` is contained in defaults/main.yml and needs to be defined in the playbook or inventory.
It's also checked in the task named `SAP SWPM Pre Install - Check availability of SAPCAR path` (line 83). In the roles `sap_hana_install` and `sap_install_media_detect`, we search for the SAPCAR file but a similar mechanism has not yet been implemented for `sap_swpm`. So we better check if the file and directory is available and modify the permissions of both.
If the `sap_swpm_sapcar_path` (the directory where sapcar is located is not available, the task `Check availability of SAPCAR path` (line 83) will fail.
And if the SAPCAR file is not there, the previous task, `Check availability of SAPCAR` (line 138), will fail.

roles/sap_swpm/tasks/swpm/prepare_software.yml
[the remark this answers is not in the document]
Wabri: Ok, so should we use something else than line_item? something like `__sap_swpm_item`?

roles/sap_swpm/tasks/swpm/prepare_software.yml
Wabri: if it fails there are something wrong in the ansible builtin shell module, it doesn't make sense in the scope of the content that set.

berndfinger: I tested and fixed it in https://github.com/sap-linuxlab/community.sap_install/pull/665/commits/538a41073b8b708f1003bdccdf0b7ad85790e734 . I could not find a problem with the shell command again.

Wabri: In case that fails, we should consider use **pipelines** `||` instead of **logical and** `&&` to be more consistent in what we have inside that:
```
set -o pipefail 
chown {{ sap_swpm_files_non_sapcar_owner }}:{{ sap_swpm_files_non_sapcar_group }} \
{{ __sap_swpm_fact_files_non_sapcar_chown_arg_list }} || 
chmod {{ sap_swpm_files_non_sapcar_mode }} \
{{ __sap_swpm_fact_files_non_sapcar_chown_arg_list }}
```

roles/sap_swpm/tasks/swpm/prepare_software.yml
Wabri: Ok, makes sense!

roles/sap_swpm/tasks/swpm/prepare_software.yml
Wabri: Ok, so this is a resolved conversation!

roles/sap_swpm/tasks/swpm/prepare_software.yml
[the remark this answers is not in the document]
berndfinger: In https://github.com/sap-linuxlab/community.sap_install/pull/665/commits/b76b655a0d9fc218462e5139859fba525a6e7325, I used `line_item`. I am open to using `__<role_name>_item` but if we start using this variable name, we should better do it with all the roles in the collection. But this would mean a large change.

Before agreeing on using `__<role_name>_item` vs. using `line_item`, I'd like to have a good justification. 

@Wabri Do you have some examples, or do you have time to do some investigation on this topic? If not, I suggest we merge first and continue investigation during the next weeks. And then we could do the change from line_item to `__<role_name>_item` for the roles in a few PRs.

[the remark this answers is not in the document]
Wabri: My question is not a request, I don't want to change a thousand of lines just for good practice. For me is good like this, I was just asking!

Wabri (approved):

ja9fuchs (approved):
:+1:

PR #668, opened 2024-02-29
sap_maintain_etc_hosts: Fix wrong assert messages

berndfinger:
Also fixes some typos.

Solves issue #662.

Issue #667, opened 2024-02-29
sap_maintain_etc_hosts: no longer use the ansible.utils.ip test


berndfinger:
We should avoid using non-`ansible.builtin` code wherever possible so we reduce the number of dependencies.

berndfinger:
The source of the ansible.utils.ip test is here: https://github.com/ansible-collections/ansible.utils/blob/main/plugins/test/ip.py

berndfinger:
Closed by https://github.com/sap-linuxlab/community.sap_install/pull/669.

PR #669, opened 2024-02-29
sap_maintain_etc_hosts: no longer use ansible.utils.ip

berndfinger:
Checking the ip address against a regex pattern should be sufficient to ensure that valid IP addresses are passed to the role.

Solves issue #667.

Should not cause merge conflicts with #668 but probably worth checking.


rhmk (approved):
LGTM

Issue #671, opened 2024-03-05
sap_*_preconfigure: Tests with ansible parameter -i instead of -l


streitd:
Switching from using the `-l` parameter to `-i` in tests scripts running ansible-playbook command. No more need to specify hosts in `/etc/ansible/hosts` each time.

berndfinger:
@streitd Sorry, I found your commit only after I provided mine. But mine also has the fix for the test programs in `sap_hana_install`. ;-)

berndfinger:
Solved in https://github.com/sap-linuxlab/community.sap_install/pull/680. Also in https://github.com/sap-linuxlab/community.sap_install/pull/672 but that one had been opened against the main tree.
@streitd Thanks for reporting and fixing this!

PR #672, opened 2024-03-05
sap_*_preconfigure: Tests with ansible parameter -i instead of -l #671

streitd:
This PR is changing ansible-playbook parameter `-l` to `-i` in tests scripts.
No more need to specify hosts in the file `/etc/ansible/hosts` each time.

### DIFF
--- a/roles/sap_general_preconfigure/tests/run-sap_general_preconfigure-tests.py
+++ b/roles/sap_general_preconfigure/tests/run-sap_general_preconfigure-tests.py
@@ -143,6 +143,7 @@ for par1 in __tests:
     command = (
         'ansible-playbook sap_general_preconfigure-default-settings.yml '
         + par1['command_line_parameter']
+        + '-u root '
         + '-i '
         + _managed_node
         + ', '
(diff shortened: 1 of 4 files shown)


Wabri (approved):

berndfinger:
Hi @streitd,
We should also add `-u root` to the `ansible-playbook` command (so that a normal user can also run the test).
We also should fix the same issue for the role `sap_hana_install` .
Feel free to use the code from https://github.com/berndfinger/community.sap_install/commit/8c40e91ec3d881533d25843efd1af16448a8ba16 and add it as a new commit to your branch so that it will be part of this PR.

berndfinger (approved):
LGTM!

berndfinger (changes requested):
Wait! We need to merge into dev, not in main.

berndfinger (changes requested):
I tried to modify the destination branch to `main` but this would lead to lots of other changes to be reverted.
@streitd Can you please create a new PR with all the changes but based on the upstream dev tree?

PR #678, opened 2024-03-06
Fix automate open issue action

Wabri:
The issue #677 should not be there because is not in the requirement file.

This pr fix this problem via parsing the requirement file and open the issue only if the package is listed in that file.


Wabri:
Close as #690 resolve all of this problems

Issue #681, opened 2024-03-09
sap_swpm: incorrect label in loop in swpm.yml


rnozaki:
1. In the file swpm.yml, under the role sap_swpm one of the tasks is pointing to an incorrect name for a label in a loop_control.

```
- name: SAP SWPM - Display status from sapcontrol file/s
  ansible.builtin.debug:
    msg: "{{ ['GetInstanceProperties'] + file_output.stdout_lines }}"
  loop: "{{ swpm_sapcontrol_file_contents.results }}"
  loop_control:
    loop_var: file_output
    label: "{{ file_output.item }}"       # << This should be line_item, as this was added recently to a loop in a previous task
```

2. After I used the dev branch (from PR $665), I got the following error:

```
###
TASK [community.sap_install.sap_swpm : SAP SWPM - Display status from sapcontrol file/s] ***15:12:00
failed: [xxxxxx] (item=None) => {
"msg": "Failed to template loop_control.label: 'dict object' has no attribute 'item'"
}
fatal: [xxxxxx]: FAILED! => {"msg": "One or more items failed"}
```

3. I noticed that 2 tasks above there was a recent change adding the loop_var under loop_control.

4. So I updated the code:
a. From:  label: "{{ file_output.item }}"   
b. To:  label: "{{ file_output.line_item }}"

5. With this update, the playbook was re-executed without error.

marcelmamula:
Did you check previous step in execution log that it registered that variable? It is possible it did not find any files, therefore item was not defined.

I have been running it in current state and I never had this issue with .item.

```
- name: SAP SWPM - Find sapcontrol log file/s
  ansible.builtin.find:
    paths: "{{ swpm_last_install_path.stdout }}"
    recurse: false
    file_type: file
    patterns: '.*sapcontrol.*log'
    use_regex: true
  register: swpm_output_sapcontrol_files
```

berndfinger:
@rnozaki I had the same error when testing an S/4HANA Foundation 2022 installation last week but I could not finish my investigation. Trying to reproduce it again now and also testing the solution you proposed.

rnozaki:
> Did you check previous step in execution log that it registered that variable? It is possible it did not find any files, therefore item was not defined.
> 
> I have been running it in current state and I never had this issue with .item.
> 
> ```
> - name: SAP SWPM - Find sapcontrol log file/s
>   ansible.builtin.find:
>     paths: "{{ swpm_last_install_path.stdout }}"
[...]

PR #682, opened 2024-03-09
sap_swpm: Fix item label in swpm.yml

rnozaki:
This is to address issue #681, which was created because an error was thrown complaining about the name in the label.

I noticed that the new name had been added 2 tasks above, and then I used the same label name on the last task and re-executed the playbook, which then finished successfully.


berndfinger (approved):
LGTM!

PR #686, opened 2024-03-18
sap_netweaver_preconfigure-RHEL: Sync with SAP note 3119751 v.13

berndfinger:
We need to sync the role `sap_netweaver_preconfigure` with the latest status of SAP note 3119751:

For RHEL 8, the following needs to be done to be prepared for all SAP kernels which are supported on RHEL 8 so that the SAP kernel can find the GCC10 symbols:

- In directory `/opt/rh/SAP/lib64` (the directory where the shared libraries of the `compat-sap-c++-<version>` package(s) are placed), a symlink named `libstdc++.so.6` needs to be present, pointing to file `compat-sap-c++-<version>.so` in the same directory, with `<version>` being `10` or higher.
- For certain older versions of the SAP kernel, there needs to be an additional symlink in directory `/usr/sap/lib`, named  `libstdc++.so.6` and pointing to file `/opt/rh/SAP/lib64/libstdc++.so.6`. This is because `RPATH` for older kernels did not yet include `/opt/rh/SAP/lib64`.

For simplicity, maintainability and supportability, we always create both links for RHEL 8.

### DIFF
--- a/roles/sap_netweaver_preconfigure/tasks/sapnote/3119751.yml
+++ b/roles/sap_netweaver_preconfigure/tasks/sapnote/3119751.yml
@@ -7,49 +7,52 @@
 
 # Note: This file is only included for RHEL 8, so no further when condition is required here.
 
-# Reason for noqa: We do not want to fail if there is no output from the ls command.
-- name: Identify the latest compat-sap-c++-NUM.so file with NUM >= 10 # noqa risky-shell-pipe
-  ansible.builtin.shell:
-  args:
-    cmd: ls compat-sap-c++-1* | sort | tail -1
-    chdir: /opt/rh/SAP/lib64
-  register: __sap_netweaver_preconfigure_register_ls_compat_sap_cpp
-  changed_when: false
+- name: Indentify all compat-sap-c++-NUM.so files with NUM >= 10
+  ansible.builtin.find:
+    paths: '/opt/rh/SAP/lib64'
+    patterns: "compat-sap-c++-1*"
+  register: __sap_netweaver_preconfigure_register_find_compat_sap_cpp
 
-# Note: The file compat-sap-c++-NUM.so file with NUM >= 10 will be available if the role sap_general_preconfigure has been run before.
-- name: Fail if there is no file compat-sap-c++-NUM.so file with NUM >= 10
+# Note: The file compat-sap-c++-NUM.so with NUM >= 10 will be available if the role sap_general_preconfigure has been run before,
+# which is a requirement.
+- name: Fail if there is no file 'compat-sap-c++-NUM.so' file with NUM >= 10
   ansible.builtin.fail:
-    msg: |
-      - There is no file /opt/rh/SAP/lib64/compat-sap-c++-NUM.so file with NUM >= 10.
-      - Make sure that package compat-sap-c++-10 or later is installed.
-  when: __sap_netweaver_preconfigure_register_ls_compat_sap_cpp.stdout is undefined or
-        __sap_netweaver_preconfigure_register_ls_compat_sap_cpp.stdout is none or
-        __sap_netweaver_preconfigure_register_ls_compat_sap_cpp.stdout | length == 0
+    msg: There is no file '/opt/rh/SAP/lib64/compat-sap-c++-NUM.so' file with NUM >= 10!
+  when: __sap_netweaver_preconfigure_register_find_compat_sap_cpp.matched == 0
+
+# Note: All following tasks depend on the previous task not having failed, so no further when condition is used below.
+- name: Set fact for file names from find result
+  ansible.builtin.set_fact:
+    __sap_netweaver_preconfigure_fact_compat_sap_cpp_filenames: "{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_filenames | d([]) + [ __sap_netweaver_preconfigure_item.path | basename ] }}"
+  loop: "{{ __sap_netweaver_preconfigure_register_find_compat_sap_cpp.files }}"
+  loop_control:
+    loop_var: __sap_netweaver_preconfigure_item
+    label: __sap_netweaver_preconfigure_item
+
+- name: Set fact for the latest compat-sap-c++.NUM.so file name
+  ansible.builtin.set_fact:
+    __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest: "{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_filenames | sort | last }}"
 
 - name: Display the identified compat-sap-c++-NUM.so file name
   ansible.builtin.debug:
-    msg: "File /opt/rh/SAP/lib64/{{ __sap_netweaver_preconfigure_register_ls_compat_sap_cpp.stdout }} is present."
+    msg: "File /opt/rh/SAP/lib64/{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest }} is present."
 
-- name: Ensure necessary symlinks for RHEL 8 are available
-  when: __sap_netweaver_preconfigure_register_ls_compat_sap_cpp.stdout | length > 0
-  block:
+- name: Ensure there is a symlink in directory '/opt/rh/SAP/lib64' named 'libstdc++.so.6', pointing to '{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest }}'
+  ansible.builtin.file:
+    src: "{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest }}"
+    dest: "/opt/rh/SAP/lib64/libstdc++.so.6"
+    state: link
 
-    - name: Ensure there is a symlink in directory '/opt/rh/SAP/lib64' named 'libstdc++.so.6', pointing to '{{ __sap_netweaver_preconfigure_register_ls_compat_sap_cpp.stdout }}'
-      ansible.builtin.file:
-        src: "{{ __sap_netweaver_preconfigure_register_ls_compat_sap_cpp.stdout }}"
-        dest: "/opt/rh/SAP/lib64/libstdc++.so.6"
-        state: link
+- name: Ensure directory '{{ sap_netweaver_preconfigure_rpath }}' is present
+  ansible.builtin.file:
+    path: "{{ sap_netweaver_preconfigure_rpath }}"
+    state: directory
+    owner: root
+    group: root
+    mode: '0755'
 
-    - name: Ensure directory '{{ sap_netweaver_preconfigure_rpath }}' is present
-      ansible.builtin.file:
-        path: "{{ sap_netweaver_preconfigure_rpath }}"
-        state: directory
-        owner: root
-        group: root
-        mode: '0755'
-
-    - name: Ensure there is a symlink in directory '{{ sap_netweaver_preconfigure_rpath }}' named 'libstdc++.so.6' pointing to '/opt/rh/SAP/lib64/libstdc++.so.6'
-      ansible.builtin.file:
-        src: /opt/rh/SAP/lib64/libstdc++.so.6
-        dest: "{{ sap_netweaver_preconfigure_rpath }}/libstdc++.so.6"
-        state: link
+- name: Ensure there is a symlink in directory '{{ sap_netweaver_preconfigure_rpath }}' named 'libstdc++.so.6' pointing to '/opt/rh/SAP/lib64/libstdc++.so.6'
+  ansible.builtin.file:
+    src: /opt/rh/SAP/lib64/libstdc++.so.6
+    dest: "{{ sap_netweaver_preconfigure_rpath }}/libstdc++.so.6"
+    state: link
--- a/roles/sap_netweaver_preconfigure/tasks/sapnote/3119751.yml
+++ b/roles/sap_netweaver_preconfigure/tasks/sapnote/3119751.yml
@@ -7,7 +7,7 @@
 
 # Note: This file is only included for RHEL 8, so no further when condition is required here.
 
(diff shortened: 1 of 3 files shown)


roles/sap_netweaver_preconfigure/tasks/sapnote/3119751.yml
```diff
@@ -5,23 +5,51 @@
...
+
+# Reason for noqa: We do not want to fail if there is no output from the ls command.
+- name: Identify the latest compat-sap-c++-NUM.so file with NUM >= 10 # noqa risky-shell-pipe
+  ansible.builtin.shell:
+  args:
+    cmd: ls compat-sap-c++-1* | sort | tail -1
+    chdir: /opt/rh/SAP/lib64
+  register: __sap_netweaver_preconfigure_register_ls_compat_sap_cpp
+  changed_when: false
+
```
ja9fuchs, line 18: Is there a reason why `ansible.builtin.find` would not work? It also supports regex and powerful filter options, similar to the find command itself.

ja9fuchs (changes requested):

roles/sap_netweaver_preconfigure/tasks/sapnote/assert-3119751.yml
```diff
@@ -5,46 +5,77 @@
     msg: "SAP note {{ (__sap_netweaver_preconfigure_sapnotes_versions | selectattr('number', 'match', '^3119751$') | first).number }}
           (version {{ (__sap_netweaver_preconfigure_sapnotes_versions | selectattr('number', 'match', '^3119751$') | first).version }}): Linux Requirements for SAP Kernel 754 and for SAP Kernel 788 and higher"
 
-- name: Get info about the compat-sap-c++-10.so file
+# Note: This file is only included for RHEL 8, so no further when condition is required here.
+
+- name: Identify the latest compat-sap-c++-NUM.so file with NUM >= 10 # noqa risky-shell-pipe
+  ansible.builtin.shell:
+  args:
+    cmd: ls compat-sap-c++-1* | sort | tail -1
```
ja9fuchs, line 13: Same here, I suggest using the `file` module instead of shell.

rhmk (commented):
See comments from Janine, as it is good practise to avoid shell/command module

roles/sap_netweaver_preconfigure/tasks/sapnote/3119751.yml
berndfinger: No known reason, fixed now! Thanks for reporting this.

roles/sap_netweaver_preconfigure/tasks/sapnote/assert-3119751.yml
berndfinger: See above, now fixed.

roles/sap_netweaver_preconfigure/tasks/sapnote/3119751.yml
```diff
@@ -5,23 +5,54 @@
...
+  ansible.builtin.set_fact:
+    __sap_netweaver_preconfigure_fact_compat_sap_cpp_filenames: "{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_filenames | d([]) + [__sap_netweaver_preconfigure_item.path | basename] }}"
+  loop: "{{ __sap_netweaver_preconfigure_register_find_compat_sap_cpp.files }}"
+  loop_control:
+    loop_var: __sap_netweaver_preconfigure_item
+    label: "{{ __sap_netweaver_preconfigure_item.path }}"
+
+- name: Set fact for the latest 'compat-sap-c++.NUM.so' file name
+  ansible.builtin.set_fact:
+    __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest: "{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_filenames | sort | last }}"
```
ja9fuchs, line 34: Another suggestion to simplify the processing of the file query results by using map filters and combining 2 tasks since it got rid of the step going through the loop.

The following example gets a list of the `path` elements, sorts and reduces to the last file and strips the path from that file.
The outcome is of type `AnsibleUnsafeText` and not unicode. No idea if that makes any difference in further processing, hence mentioning.

Tested example:
```
- name: find files
  find:
    path: /etc
  register: etc_files

- set_fact:
    lastfile: >
      {{ etc_files.files
        | map(attribute = 'path')
	| sort | last | basename }}

- debug:
    var: lastfile
```


```suggestion
- name: Set fact for the latest 'compat-sap-c++.NUM.so' file name
  ansible.builtin.set_fact:
    __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest: >
      {{ __sap_netweaver_preconfigure_register_find_compat_sap_cpp.files
        | map(attribute = 'path')
        | sort | last | basename }}
```

roles/sap_netweaver_preconfigure/tasks/sapnote/assert-3119751.yml
```diff
@@ -5,46 +5,90 @@
...
+  loop: "{{ __sap_netweaver_preconfigure_register_find_compat_sap_cpp.files }}"
+  loop_control:
+    loop_var: __sap_netweaver_preconfigure_item
+    label: "{{ __sap_netweaver_preconfigure_item.path }}"
+  when: __sap_netweaver_preconfigure_register_find_compat_sap_cpp.matched > 0
+
+- name: Set fact for the latest 'compat-sap-c++.NUM.so' file name
+  ansible.builtin.set_fact:
+    __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest: "{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_filenames | sort | last }}"
+  when: __sap_netweaver_preconfigure_register_find_compat_sap_cpp.matched > 0
```
ja9fuchs, line 28: Same suggestion as above applies to these 2 tasks. They can be combined to one simpler task, since the whole purpose of the first large task is to parse the find results in a loop, which can be achieved by using the `map` filter in combination with the filters in the second task.

roles/sap_netweaver_preconfigure/tasks/sapnote/3119751.yml
berndfinger: Thanks for the suggestion. The proposed code does not work but with a little modification, I could solve this problem. Here's my test playbook:
```
    - name: Set fact for the latest 'compat-sap-c++.NUM.so' file name
      ansible.builtin.set_fact:
        __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest_1: >
          {{ __sap_netweaver_preconfigure_register_find_compat_sap_cpp.files
            | map(attribute = 'path')
            | sort | last | basename }}

    - name: debug 1
      debug:
        msg:
          - ">>>{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest_1 }}<<<"

    - name: Set fact for the latest 'compat-sap-c++.NUM.so' file name
      ansible.builtin.set_fact:
        __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest_2: "{{ __sap_netweaver_preconfigure_register_find_compat_sap_cpp.files
                                                                   | map(attribute = 'path') | sort | last | basename }}"

    - name: debug 2
      debug:
        msg:
          - ">>>{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest_2 }}<<<"
```

Output:
```TASK [Set fact for the latest 'compat-sap-c++.NUM.so' file name] ***************************************************************************
ok: [localhost]

TASK [debug 1] *****************************************************************************************************************************
ok: [localhost] => 
  msg:
  - |-
    >>>compat-sap-c++-10.so
    <<<

TASK [Set fact for the latest 'compat-sap-c++.NUM.so' file name] ***************************************************************************
ok: [localhost]

TASK [debug 2] *****************************************************************************************************************************
ok: [localhost] => 
  msg:
  - '>>>compat-sap-c++-10.so<<<'
```

ja9fuchs: Yes, my mistake with the suggestion syntax due to the rewrite.
It works when changing `>` to `>-` to strip trailing whitespace:

```
__sap_netweaver_preconfigure_fact_compat_sap_cpp_latest_1: >-
```

berndfinger: @ja9fuchs Thanks for the suggestion!

roles/sap_netweaver_preconfigure/tasks/sapnote/assert-3119751.yml
berndfinger: again, very useful, thanks!

roles/sap_netweaver_preconfigure/tasks/sapnote/3119751.yml
berndfinger: Solved with the latest commits.

roles/sap_netweaver_preconfigure/tasks/sapnote/assert-3119751.yml
```diff
@@ -5,46 +5,88 @@
...
+    use_regex: true
+  register: __sap_netweaver_preconfigure_register_find_compat_sap_cpp
+
+# Note: The symlink compat-sap-c++-NUM.so with NUM >= 10 will be available if the role sap_general_preconfigure has been run before,
+# which is a requirement.
+- name: Assert that there is at least one symlink '/opt/rh/SAP/lib64/compat-sap-c++-NUM.so' with NUM >= 10
+  ansible.builtin.assert:
+    that: __sap_netweaver_preconfigure_register_find_compat_sap_cpp.matched > 0
+    fail_msg: "FAIL: There is no symlink '/opt/rh/SAP/lib64/compat-sap-c++-NUM.so' with NUM >= 10!"
+    success_msg: "FAIL: There is at least one symlink '/opt/rh/SAP/lib64/compat-sap-c++-NUM.so' with NUM >= 10!"
```
ja9fuchs, line 24: Is the "FAIL:" in the _success_ message text intended or a copy-paste remainder?

roles/sap_netweaver_preconfigure/tasks/sapnote/assert-3119751.yml
```diff
@@ -5,46 +5,88 @@
...
+  ignore_errors: "{{ sap_netweaver_preconfigure_assert_ignore_errors | d(false) }}"
+  when: __sap_netweaver_preconfigure_register_stat_opt_rh_libstdc_assert.stat.exists
+
+- name: Assert that file '/opt/rh/SAP/lib64/libstdc++.so.6' is a symlink to '{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest }}'
+  ansible.builtin.assert:
+    that: __sap_netweaver_preconfigure_register_stat_opt_rh_libstdc_assert.stat.lnk_target == __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest
+    fail_msg: "FAIL: File '/opt/rh/SAP/lib64/libstdc++.so.6' is not a symlink to '{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest }}!'"
+    success_msg: "PASS: File '/opt/rh/SAP/lib64/libstdc++.so.6' is a symlink to '{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest }}'."
+  ignore_errors: "{{ sap_netweaver_preconfigure_assert_ignore_errors | d(false) }}"
+  when: __sap_netweaver_preconfigure_register_stat_opt_rh_libstdc_assert.stat.exists
```
ja9fuchs, line 63: I would actually combine these 2 checks. Is there really much value in separately displaying that libstdc++.so.6 is a symlink, compared to a combined output in the second task like:
- fail: is not a symlink or not pointing to `<compat-sap-cpp-lastest>`
- pass: is a symlink pointing to `<compat-sap-cpp-latest>`

roles/sap_netweaver_preconfigure/tasks/sapnote/assert-3119751.yml
berndfinger: The latter, thank for catching it!

roles/sap_netweaver_preconfigure/tasks/sapnote/assert-3119751.yml
ja9fuchs: Suggestion:

```
- name: Assert that file '/opt/rh/SAP/lib64/libstdc++.so.6' is a symlink to '{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest }}'
  ansible.builtin.assert:
    that:
      - __sap_netweaver_preconfigure_register_stat_opt_rh_libstdc_assert.stat.exists
      - __sap_netweaver_preconfigure_register_stat_opt_rh_libstdc_assert.stat.islnk
      - __sap_netweaver_preconfigure_register_stat_opt_rh_libstdc_assert.stat.lnk_target == __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest
    fail_msg: "FAIL: File '/opt/rh/SAP/lib64/libstdc++.so.6' does not exist or is not a symlink to '{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest }}!'"
    success_msg: "PASS: File '/opt/rh/SAP/lib64/libstdc++.so.6' is a symlink to '{{ __sap_netweaver_preconfigure_fact_compat_sap_cpp_latest }}'."
  ignore_errors: "{{ sap_netweaver_preconfigure_assert_ignore_errors | d(false) }}"

berndfinger: Great suggestion, thanks!

ja9fuchs (approved):
:+1: :smiley_cat:

rhmk (approved):
LGTM

Issue #696, opened 2024-04-11
sap_general_preconfigure: false duplicates in /etc/hosts due to partial matching


ja9fuchs:
Related to #132, but an issue on its own and while comments are not there on fresh end-to-end installations, the following is a problem that can occur anytime.

The `awk` search for duplicate IPs in `/etc/hosts` is not restricted to match the whole word. That leads to partial matches and a false error that ends the play and is an issue when IPs are randomly assigned during system deployment in a specified range, e.g. in cloud instances.

Example: IP check for `10.1.1.12` will conflict with an existing entry for `10.1.1.120`.

`roles/sap_general_preconfigure/tasks/RedHat/generic/assert-etc-hosts.yml`

Code: `awk 'BEGIN{a=0}/{{ sap_general_preconfigure_ip }}/{a++}END{print a}' /etc/hosts`

Suggestion to only match the whole IP of an actual entry (no comment):
Add `^` and `\s` to indicate the word ending.
`awk 'BEGIN{a=0}/^{{ sap_general_preconfigure_ip }}\s/{a++}END{print a}' /etc/hosts`

Edit: `\w` didn't work for other entries, but `\s` does.

berndfinger:
@ja9fuchs Thanks for reporting this. I already successfully tested with `\s`. See also https://github.com/sap-linuxlab/community.sap_install/blob/d36cc155dc82fdb1848b1f36d6dbbedb9d06a3ab/roles/sap_general_preconfigure/tasks/RedHat/generic/configure-etc-hosts.yml#L57 and https://github.com/sap-linuxlab/community.sap_install/blob/d36cc155dc82fdb1848b1f36d6dbbedb9d06a3ab/roles/sap_maintain_etc_hosts/tasks/update_host_present.yml#L86 .

berndfinger:
Fixed in #697, regression #736 fixed in #737.

PR #697, opened 2024-04-11
sap_general_preconfigure: Fix /etc/hosts check in assert mode

berndfinger:
Solves issue #696.

### DIFF
--- a/roles/sap_general_preconfigure/tasks/RedHat/generic/assert-etc-hosts.yml
+++ b/roles/sap_general_preconfigure/tasks/RedHat/generic/assert-etc-hosts.yml
@@ -98,7 +98,7 @@
     success_msg: "PASS: There is one line containing '{{ sap_general_preconfigure_hostname }}' in /etc/hosts!"
   ignore_errors: "{{ sap_general_preconfigure_assert_ignore_errors | d(false) }}"
 
-- name: Check hostname -s
+- name: Test hostname -s
   ansible.builtin.shell: test "$(hostname -s)" = "$(hostname)"
   register: __sap_general_preconfigure_register_hostname_s_assert
   ignore_errors: yes
@@ -111,7 +111,7 @@
     success_msg: "PASS: The output of 'hostname -s' matches the output of 'hostname'."
   ignore_errors: "{{ sap_general_preconfigure_assert_ignore_errors | d(false) }}"
 
-- name: Check hostname -f
+- name: Test hostname -f
   ansible.builtin.shell: test "$(hostname -f)" = "$(hostname).$(hostname -d)"
   register: __sap_general_preconfigure_register_hostname_f_assert
   ignore_errors: yes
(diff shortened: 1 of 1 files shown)


roles/sap_general_preconfigure/tasks/RedHat/generic/assert-etc-hosts.yml
```diff
@@ -20,17 +20,18 @@
     success_msg: "PASS: The line '{{ sap_general_preconfigure_ip }} {{ sap_general_preconfigure_hostname }}.{{ sap_general_preconfigure_domain }} {{ sap_general_preconfigure_hostname }}' is once in /etc/hosts."
   ignore_errors: "{{ sap_general_preconfigure_assert_ignore_errors | d(false) }}"
 
+# We allow more than one line containing sap_general_preconfigure_ip:
```
ja9fuchs, line 23: Why though? Only the first entry is used.

roles/sap_general_preconfigure/tasks/RedHat/generic/assert-etc-hosts.yml
```diff
@@ -20,17 +20,18 @@
...
 - name: Count the number of sap_general_preconfigure_ip ({{ sap_general_preconfigure_ip }}) entries in /etc/hosts
-  ansible.builtin.command: awk 'BEGIN{a=0}/{{ sap_general_preconfigure_ip }}/{a++}END{print a}' /etc/hosts
+  ansible.builtin.command: awk 'BEGIN{a=0}/^{{ sap_general_preconfigure_ip }}\s/{a++}END{print a}' /etc/hosts
   register: __sap_general_preconfigure_register_sap_ip_once_assert
   ignore_errors: yes
   changed_when: no
 
-- name: Assert that there is just one line containing {{ sap_general_preconfigure_ip }} in /etc/hosts
+- name: Assert that there is at least one line containing {{ sap_general_preconfigure_ip }} in /etc/hosts
```
ja9fuchs, line 30: As a sysadmin, I would not allow PASS-ing this check if one IP is configured more than once in /etc/hosts.

ja9fuchs (changes requested):
Questioning the tolerance to duplicate IP entries in `/etc/hosts`.
I cannot think of any valid reason for such a configuration to be considered healthy and a basis to run any application installation on.

roles/sap_general_preconfigure/tasks/RedHat/generic/assert-etc-hosts.yml
berndfinger: We currently allow more than one line with the same IP address in /etc/hosts in https://github.com/sap-linuxlab/community.sap_install/blob/42af6baab0310ec9c47994147d877f3d11ca4108/roles/sap_maintain_etc_hosts/tasks/update_host_present.yml#L85, so we now should discuss the impact of allowing such a configuration in a broader context.

rhmk: Hi, 
IIRC this was a user request. This user is putting host aliases to separate lines instead of appending to a single line. This was in the very beginning and I am not sure if this has been proper documented. 
So I am fine with approving this PR and wait for complaints. Especially because I haven't found any evidence, that this is a common practise in the net.

berndfinger:
After [the reply from @rhmk](https://github.com/sap-linuxlab/community.sap_install/pull/697#discussion_r1565554584) , let's no longer allow more than one line containing the same IPv4 address in /etc/hosts. This also affects the role `sap_maintain_etc_hosts`, see https://github.com/sap-linuxlab/community.sap_install/issues/708.

rhmk (approved):
LGTM:

Issue #699, opened 2024-04-12
sap_install_media_detect: incorrect detection of file JAVA export file


berndfinger:
The `sapfile` tool, which is part of role `sap_install_media_detect`, is detecting the file `SAP S/4HANA 2023 Java`, file name `51057035`, file type `ZIP`, as `sap_export_ecc_ides` instead of the correct file type `sap_export_nwas_java` or `sap_export_s4hana_java`.

Initially reported as PR https://github.com/sap-linuxlab/community.sap_install/pull/695.

berndfinger:
Solved in #701 .

PR #695, opened 2024-04-08
sap_install_media_detect: S4HANA Java stack installation

remrozk:
### Context
SAP S/4HANA Java stack is still using SWPM1 (https://help.sap.com/doc/6b11678926d3409bbfea8897cb34d10f/2023/en-US/INST_OP2023.pdf) and still needs DATA_UNITS to be installed.
s4hana_java

### Enhancement

- Create new value `saps4hana_java` for the var `sap_install_media_detect_export`
- Update sapfile to be able to detect `sap_export_s4hana_java`
- Update `roles/sap_install_media_detect/tasks/find_files_after_extraction.yml` in order to manage `sap_export_s4hana_java`
- Update `roles/sap_install_media_detect/tasks/prepare/create_file_list_phase_2.yml` in order to manage `sap_export_s4hana_java`
- Update `roles/sap_install_media_detect/tasks/set_global_vars.yml` in order to manage `saps4hana_java`

### Output
```log
TASK [community.sap_install.sap_install_media_detect : SAP Install Media Detect - Prepare - Set fact with the results of the sapfile command]
...
[...]

PR #701, opened 2024-04-12
sap_install_media_detect: Improve SAP file type detection in sapfile

berndfinger:
Relates to issue #699.

We don't introduce a new file type because the file named `51057035` (renamed by the role to `51057035.zip` before `sapfile` is called) will be detected as type `sap_export_nwas_java`, which should be sufficient for correct processing.


sean-freeman:
As per my comment in #695 , `sap_export_nwas_java` logic should allow for SAP S/4HANA JAVA export media detection. This PR resolves a bug where it was not being detected correctly.

In the future this could be altered, so that a distinct subdirectory `/<software_path>/sap_export_s4hana_java` is created. This may provide a minor additional benefit for end-users, although it would be functionally zero difference to existing `sap_export_nwas_java` logic.

I suggest this PR is merged and PR 695 closed.

PR #702, opened 2024-04-12
sap_install_media_detect: Detection of duplicate SAR files for EXE, EXEDB, igs, igshelper and webdisp should count only file with type=sapcar

rob0d:
This is a fix for issue #703

### DIFF
--- a/roles/sap_install_media_detect/tasks/prepare/create_file_list_phase_2.yml
+++ b/roles/sap_install_media_detect/tasks/prepare/create_file_list_phase_2.yml
@@ -120,7 +120,7 @@
       ansible.builtin.assert:
         that:
           - __sap_install_media_detect_fact_files_sapfile_results | selectattr('sap_file_type', 'equalto', 'sap_kernel') | length > 0
-          - __sap_install_media_detect_fact_files_sapfile_results | selectattr('file', 'search', 'SAPEXE_') | length == 1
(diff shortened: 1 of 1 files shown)


berndfinger:
I tested a download directory with S/4HANA Foundation 2022 with:
- `SAPEXE_201-70006642.SAR`
- `SAPEXEDB_201-70006641.SAR`
- `igsexe_4-70005417.sar`
- `igshelper_17-10010245.sar`
and some four dummy text files with the same prefix but without the suffix (e.g. `SAPEXE_201-70006642`).
When using the dev tree with the proposed change in this PR, the role finished with the expected entries for `sap_swpm_igs_file_name`, `sap_swpm_igs_helper_file_name`, `sap_swpm_kernel_dependent_file_name`, and `sap_swpm_kernel_independent_file_name`.
When using the current dev tree, the role fails (as described in https://github.com/sap-linuxlab/community.sap_install/issues/703) with:
```
  assertion: __sap_install_media_detect_fact_files_sapfile_results | selectattr('file', 'search', 'SAPEXE_') | length == 1
  evaluated_to: false
  msg: No, or more than one, DB independent SAP Kernel file found
```
When using the dev tree but without the four dummy text files, the resulting outputs of task `Test result for 'sap_s4hana_foundation_2022': Display _result_dict` are identical.

---> PASS

berndfinger (approved):
LGTM

Issue #719, opened 2024-04-18
sap_swpm: No longer require module community.general.pids


berndfinger:
We should be able to replace `community.general.pid` by `ps -ef` and `awk` processing using the `ansible.builtin.shell` module. The goal is mainly to avoid installing some Python related package and module installations and to simplify support for the downstream collection version.

Issue #716, opened 2024-04-18
Workflows: galaxy-importer fails due to missing module community.general.pids


berndfinger:
See https://ansible.softwarefactory-project.io/zuul/build/69ccd596b20c4ef5bc0789fcd70f4fc6 or https://ansible.softwarefactory-project.io/zuul/build/69ccd596b20c4ef5bc0789fcd70f4fc6/log/job-output.txt#672 .

The error message is:
`WARNING: roles/sap_swpm/tasks/swpm.yml:74:3: 'syntax-check[unknown-module]:' 'couldn'\''t' resolve module/action ''\''community.general.pids'\''.`

marcelmamula:
@berndfinger Wouldn't this become redundant change when https://github.com/sap-linuxlab/community.sap_install/pull/720 is implemented?

berndfinger:
> @berndfinger Wouldn't this become redundant change when #720 is implemented?

Yes, that was one of the justifications for opening #720. However, as per https://github.com/sap-linuxlab/community.sap_install/issues/709#issuecomment-2066181345, changing essential code in `sap_swpm` will trigger a lot of testing and thus take some time before we can integrate it, so both of #709 and #720 might take some time. Maybe fixing #716 can be done earlier.

ja9fuchs:
@berndfinger Is this still an issue? I thought it was fixed in https://github.com/sap-linuxlab/community.sap_install/pull/786?

PR #720, opened 2024-04-18
sap_swpm: Remove pids module

berndfinger:
We should better remove the `community.general.pids` module. Advantages:
- No dependency on `community.general` in the role sap_swpm, simplifying support for downstream versions
- No more Python related package installations at the beginning of the role
- Probably also solving https://github.com/sap-linuxlab/community.sap_install/issues/716.

Solves https://github.com/sap-linuxlab/community.sap_install/issues/719.

### DIFF
--- a/roles/sap_swpm/tasks/swpm.yml
+++ b/roles/sap_swpm/tasks/swpm.yml
@@ -23,20 +23,19 @@
 ### Async method
 
 # Required for Ansible Module pids
-- name: Install Python devel, Python pip and gcc to system Python
-  ansible.builtin.package:
-    name:
-      - python3-devel
-      - python3-pip
-      - gcc
-    state: present
+#- name: Install Python devel, Python pip and gcc to system Python
+#  ansible.builtin.package:
+#    name:
+#      - python3-devel
+#      - python3-pip
+#      - gcc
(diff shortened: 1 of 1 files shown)


marcelmamula (commented):
@berndfinger Tested and it works for sap_swpm, but it breaks recent changes in sap_ha_pacemaker_cluster and ha_cluster roles.

New ASCS/ERS post steps were added that include python3-pip with pexpect for ansible.builtin.expect module in https://github.com/sap-linuxlab/community.sap_install/blob/ef0636394cedc582ce4e4614b22b9da6e207f20f/roles/sap_ha_pacemaker_cluster/tasks/Suse/post_steps_nwas_abap_ascs_ers.yml#L20

@ja9fuchs We would need to take these removed python/pip steps and put them into sap_ha role instead.

sean-freeman (approved):

berndfinger:
> @berndfinger did you figure out your issues that you mentioned?

None of the issues I had seen were related to this code change but rather to wrong parameter settings.

marcelmamula (approved):
LGTM

berndfinger:
Aborting PR due to an error when resolving a merge conflict.

PR #722, opened 2024-04-19
sap_hana_install: update README.md

marcoschum:
Initial checks whether a system is already running or not will only be performed if "sap_hana_install_force" is set to false. If it is set to true if will not run and potentials overwrite the installation. Just a documentation error.

PR #723, opened 2024-04-22
sap_hana_install: Update README.md regarding parameter sap_hana_install_force

marcoschum:
Updated README as per suggestions in PR: sap_hana_install: update README.md #722.

Clearification on the description of section "Perform Initial Checks". Those checks will only be skipped if "sap_hana_install_force" is set to "true".


berndfinger (approved):
LGTM!

PR #727, opened 2024-04-26
sap_hana_install: Add support for fapolicyd

berndfinger:
This PR adds support for fapolicyd:
- The presence of the package `fapolicyd` is ensured.
- The integrity setting in `/etc/fapolicyd/fapolicyd.conf` will be `sha256` instead of the default `none`.
- The presence of a new fapolicyd rule file in `/etc/faplicyd/rules.d`, for protecting shell scripts, is ensured. 
- The presence of all executables under `/hana` and `/usr/sap` in fapolicyd trust files in `/etc/faplicyd/trust.d` is ensured.
- The service `fapolicyd` is enabled and running.

I introduced a new role parameter, `sap_hana_install_root_path` which is needed for this functionality, and replaced `/hana` by this variable and `/hana/shared` by its corresponding variable.

I also modified `yes/no` to `true/false` in `defaults/main.yml` so we over time only use those for all roles.

Solves https://github.com/sap-linuxlab/community.sap_install/issues/728.

### DIFF
--- a/roles/sap_hana_install/defaults/main.yml
+++ b/roles/sap_hana_install/defaults/main.yml
@@ -33,7 +33,7 @@ sap_hana_install_keep_copied_sarfiles: no
 sap_hana_install_use_fapolicyd: no
 
 # When using fapolicyd, the file name in /etc/fapolicyd/trust.d/ must be set:
-sap_hana_install_fapolicyd_trust_filename: '_hana'
+sap_hana_install_fapolicyd_trust_filename: 'hana'
 
 # File name of SAPCAR*EXE in the software directory. If the variable is not set and there is more than one SAPCAR executable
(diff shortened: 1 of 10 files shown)


roles/sap_hana_install/defaults/main.yml
rhmk, line 1: Line 28 should be 'true' instead of false -- but it is just a comment

rhmk (changes requested):
please correct the comments, they are misleading. The rest looks fine for me

berndfinger:
> please correct the comments, they are misleading. The rest looks fine for me
Thanks for your review! All good now (I hope).

rhmk (approved):
LGTM

sean-freeman (approved):
lgtm with backwards compatibility for the var names of directory paths (i.e. /hana)

Issue #729, opened 2024-04-26
sap_install_media_detect: allow separate detection of exe and exedb for installations where exedb isn't required


rob0d:
@sean-freeman  as per  your message to track this enhancement suggestion.
When set [sap_install_media_detect_kernel](https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_install_media_detect/defaults/main.yml#L73) the code in `sap_install_media_detect` currently detects both EXE and EXEDB parts of the SAP Kernel.
Some types of installations don't require EXEDB, only EXE.

Suggestion is to change the behaviour  to allow  the kernel detection to separately check EXE or EXEDB or both.
Options probably depend on how much we care about backwards compatibility as  `sap_install_media_detect_kernel` is used in many places.
Option 1:
Allow more than just true/false, but keep true for compatibility purposes:
 - to detect both parts of kernel  set this to: `true` or `all`
 - to detect part1/exe set this to: `exe` or `part1`
 - to detect part2/exedb set hit to: `exedb` or `part2`

Option2:
Introduce new control parameters: `sap_install_media_detect_kernel_part1` + `sap_install_media_detect_kernel_part2`
ans possibly for backward compatibility keep `sap_install_media_detect_kernel` which will override the two new parameters if set to true. This may lead to a slightly confusing behaviour is the three parameters are not inline.

berndfinger:
@rob0d @sean-freeman - According to the comments about `sap_install_media_detect_kernel_db` in `defaults/main.yml`, setting this parameter to the database type is only necessary if there is more than one SAPEXEDB file in the source directory. Taking into account the absence of this parameter in https://github.com/sap-linuxlab/ansible.playbooks_for_sap, I agree that we should be careful when modifying the meaning of the parameter `sap_install_media_detect_kernel` or `sap_install_media_detect_kernel_db`.

Wouldn't a new value for `sap_install_media_detect_kernel_db`, like for example `absent`, be sufficient? I wouldn't choose `false` because this could give the impression that this parameter is a boolean.

The new behavior would be as follows:

- If `sap_install_media_detect_kernel_db` is **not** defined, the behavior is unchanged and the role will check for the presence of a single database dependent kernel file (SAPEXEDB).
- If `sap_install_media_detect_kernel_db` **is** defined and if the value is **not** `absent` , the behavior is unchanged and the role will check for the presence of the correct database dependent kernel file (SAPEXEDB) for the specified database type (e.g. `saphana`).
[...]

PR #732, opened 2024-05-04
sap_install_media_detect: Allow detection of SAPEXE only

berndfinger:
Solves issue #729.

### DIFF
--- a/roles/sap_install_media_detect/tasks/find_files_after_extraction.yml
+++ b/roles/sap_install_media_detect/tasks/find_files_after_extraction.yml
@@ -147,8 +147,11 @@
 - name: SAP Install Media Detect - Find files after extraction - Find SAPEXEDB, database specific - block
   when:
     - sap_install_media_detect_kernel
-    - sap_install_media_detect_kernel_db is defined
-    - sap_install_media_detect_kernel_db | length > 0
+    - sap_install_media_detect_kernel_db == 'saphana' or
+      sap_install_media_detect_kernel_db == 'sapase' or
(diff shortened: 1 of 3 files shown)


berndfinger:
I successfully tested with S/4HANA 2022 and with `sap_install_media_detect_kernel_db` unset (1) and set to `absent` (2).
When unset (1) , the output showed correct content for the variables `sap_swpm_kernel_dependent_path` and `sap_swpm_kernel_dependent_file_name`. When set to `absent` (2), the two variables were empty.

sean-freeman:
To confirm, this PR is to allow detection of "only SAPEXE" and "only SAPEXEDB" indepedently?

The current AND/OR logic seems to not match the above statement.

berndfinger:
> To confirm, this PR is to allow detection of "only SAPEXE" and "only SAPEXEDB" indepedently?

This PR is meant to support the detection of "only SAPEXE". By setting `sap_install_media_detect_kernel_db` to `none`, the role will no longer look for a SAPEXEDB file and will not report the two related output parameters for SAPEXEDB. If `sap_install_media_detect_kernel` is set to `true` and `sap_install_media_detect_kernel_db` is not set, the behavior is the same as before, meaning that the role will detect SAPEXE and SAPEXEDB.

If `sap_install_media_detect_kernel_db` is set and `sap_install_media_detect_kernel` is not set, the role will not display the output parameters for SAPEXE and SAPEXEDB. Also this case is probably not applicable, and it is out of scope for issue https://github.com/sap-linuxlab/community.sap_install/issues/729.

sean-freeman (approved):
lgtm 👍

Issue #736, opened 2024-05-10
sap_general_preconfigure: Regression after PR #697


berndfinger:
The new code of https://github.com/sap-linuxlab/community.sap_install/pull/697 is not correctly matching the hostname as a full string separated by white space(s). It also matches the hostname as a substring of a longer string.

A possible solution is to replace the line
https://github.com/sap-linuxlab/community.sap_install/blob/a97844d8d7cfb8022d8f11ffff844ece62451949/roles/sap_general_preconfigure/tasks/RedHat/generic/assert-etc-hosts.yml#L92
by:
```
  when: __sap_general_preconfigure_line_item | regex_search('(\s)' + sap_general_preconfigure_hostname + '($|\s)' )
```
.

I tested this with a sample /etc/hosts file containing:
- a line with ipv4 adress, the FQDN, the hostname and an alias, all separated by a tab, one, or two spaces, and a space at the end or not space at the end of the line.
- and another line with a different ipv4 address, the FQDN with a hostname starting with the same characters as the previous one but with 3 more characters, and the hostname without the DNS domain, and an alias.

To be done: Find out why the original pattern incorrectly matched the second line as well.

berndfinger:
Also, an Ansible-only solution should limit the amount of output lines: When looping over a long /etc/hosts file, there will be many output lines, even when optimizing the output with the `label` variable in `loop_control`. So an Ansible-only solution should use Ansible filters like `count`. However, I do currently not have enough time to test and analyze any issues related to constructing `regex` patterns from multiple strings, which would be a requirement. So I'll now revert back to the `awk` method, which is easy to understand and analyze and also can easily produce lean output.

berndfinger:
Fixed in #737.

PR #737, opened 2024-05-13
sap_general_preconfigure: Revert to awk for asserting /etc/hosts

berndfinger:
Solves issue #736.

It's essentially a backout of PR #697, especially of commit 1fc1459.


sean-freeman (approved):
lgtm 👍

PR #754, opened 2024-06-14
sap_ha_pacemaker_cluster: Add python3-pip and NFS fix for Azure

marcelmamula:
Scope:
1. Moved python3-pip code from sap_swpm to sap_ha_pacemaker_cluster to enable change in https://github.com/sap-linuxlab/community.sap_install/pull/720 for @berndfinger 

2. Added throttling and retry loop for tasks touching files on cross-mounted NFS filesystem under ASCS ERS, which was not working correctly on MS Azure NFS filesystems. It will now do change on first host by using `throttle: 1` then it will retry-loop attempt to do change on second node. @ja9fuchs 
> Testing has shown that it takes 20-40 seconds for NFS to be accessible again.


berndfinger (approved):
LGTM.

ja9fuchs (approved):
looks good, thanks :)

berndfinger (approved):
LGTM

PR #757, opened 2024-06-17
sap_hana_install: Fix issue 756

berndfinger:
Two of the tasks were enabled even when `sap_hana_install_use_fapolicyd` was set to `false`. This PR fixes the issue and ensures that all the fapolicyd tasks are now inside blocks which are only active when `sap_hana_install_use_fapolicyd` is `true`.

### DIFF
--- a/roles/sap_hana_install/tasks/post_install.yml
+++ b/roles/sap_hana_install/tasks/post_install.yml
@@ -208,6 +208,7 @@
   when:
     - sap_hana_install_use_fapolicyd
     - '"fapolicyd" in ansible_facts.packages'
+  tags: sap_hana_install_use_fapolicyd
   block:
 
     - name: SAP HANA Post Install, fapolicyd - Ensure Ansible marker for 'integrity' is present in fapolicyd config file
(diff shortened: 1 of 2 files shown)


rhmk (approved):
LGTM

ja9fuchs:
@berndfinger Cosmetical, but I think you can apply the tag of the tasks to the block now, instead all individual tasks?
`tags: sap_hana_install_use_fapolicyd`

In any case, merging soon would be nice, since this issue currently breaks the HANA post-inst. :)

roles/sap_hana_install/tasks/post_install.yml
```diff
@@ -243,81 +241,71 @@
...
+  block:
+
+    - name: SAP HANA Post Install, fapolicyd - Process template for creating rule file '{{ sap_hana_install_fapolicyd_rule_file }}'
+      ansible.builtin.template:
+        src: fapolicyd-rules.j2
+        dest: "/etc/fapolicyd/rules.d/{{ sap_hana_install_fapolicyd_rule_file }}.rules"
+        owner: root
+        group: fapolicyd
+        mode: '0644'
+      tags: sap_hana_install_use_fapolicyd
```
ja9fuchs, line 265: line 265 is obsolete now :)

ja9fuchs (approved):
lgtm

sean-freeman (approved):
Tested, lgtm 👍

PR #759, opened 2024-06-20
sap_swpm: legacy ascs ha code

sean-freeman:
Legacy code for ASCS HA, traced to original IBM LSS code which does not contain any reasoning for this code to exist

The code appears to serve no function but does cause ASCS HA <sid>adm to not default to C Shell, which is problematic for runtime stability

Tested an execution of ASCS HA without this code, and it installed ASCS HA successfully (with correct use of C Shell default for <sid>adm)

Therefore, commented out the call that imports Ansible Tasks from file `create_os_user.yml` and in future suggest this code is removed entirely


berndfinger (approved):
LGTM

PR #760, opened 2024-06-20
sap_swpm: sap_swpm_db_schema_password must be set explicitly  for AAS install due to anomaly in password_facts.yml

rob0d:


sean-freeman (approved):
Sample file only, agree to changes without testing :+1:

PR #762, opened 2024-06-20
collection: prepare for v1.4.1

berndfinger:


berndfinger:
We need to integrate two more fixes, so closing this PR now. I'll update the change log and create a new one afterwards.

PR #763, opened 2024-06-20
sap_swpm: Set DDIC password in SWPM if sap_swpm_ddic_000_password is defined

rob0d:
rather just when doing system copy as this parameter is relevant for both new installs and copies.

### DIFF
--- a/roles/sap_swpm/templates/configfile.j2
+++ b/roles/sap_swpm/templates/configfile.j2
@@ -256,7 +256,7 @@ HDB_Userstore.useABAPSSFS = false
 # NW_HDB_DBClient.clientPathStrategy = LOCAL
 {% endif %}
 
-{% if 'credentials_syscopy' in sap_swpm_inifile_list %}
+{% if sap_swpm_ddic_000_password is defined %}
 ######
 # credentials_syscopy
 ######
--- a/roles/sap_swpm/templates/configfile.j2
+++ b/roles/sap_swpm/templates/configfile.j2
@@ -257,9 +257,6 @@ HDB_Userstore.useABAPSSFS = false
 {% endif %}
 
 {% if sap_swpm_ddic_000_password is defined %}
-######
-# credentials_syscopy
-######
 # Are the passwords for the DDIC users different from the default value?
 NW_DDIC_Password.needDDICPasswords = true
 NW_DDIC_Password.ddic000Password = {{ sap_swpm_ddic_000_password }}


sean-freeman:
While it can be used, is this common to use in new installations and worth doing? The `credentials` block with master password, <sapsid>adm password etc. are the the usual candidates for change; the DDIC can be altered after initial login?

I'm flexible either way. Although if this is made generic:

1. the section would be named `credentials_ddic` (and retaining the ## prettification so the end-user can see where it came from after generated),
2. this would be moved upwards in the Jinja template and directly underneath the `credentials` block.

> It's a breaking change, so we have to be purposeful here.

> Side note: Please prefix the commits with the Ansible Role name you're changing, such as _`sap_swpm: set DDIC password in SWPM if sap_swpm_ddic_000_password is defined`_

rob0d:
Hi @sean-freeman,
This is a tentative pull request as I am not sure how do you typically handle changes to the jinja template. I will review your comments and update accordingly.
To answer your initial question - yes. Some customers have a very strict security policies. Sometimes I don't even know that the SAP* and DDIC passwords are as they are  automatically generated and stored in a vault when executing the playbooks and the security people create user accounts before anyone can login. Having SAP* and DDIC password the same during install would mean an extra manual step that is not really necessary or productive.
The [ role defaults](https://github.com/sap-linuxlab/community.sap_install/blob/42af6baab0310ec9c47994147d877f3d11ca4108/roles/sap_swpm/defaults/main.yml#L190) say that this can be used for new installs, so from my point of view the current behaviour is a bug.
This can potentially make `credentials_syscopy` obsolete as the only place this is used was in jinja template and this request changes that.
So I need some guidance regarding this one from you guys,

sean-freeman:
@rob0d To be verbose, to facilitate closing this PR I would like to see the following:
- the code block in Jinja
- appending the parameter in the YAML List (i.e. the default SWPM inifile.params generation list) in the defaults file so the variable is handled automatically if it exists

This breaking change would be noted in the Release Notes, but the following should fix your circumstance

`templates/configfile.j2`
```jinja
{% if 'credentials_ddic_000' in sap_swpm_inifile_list %}
######
# credentials_ddic_000
######
# Are the passwords for the DDIC users different from the default value?
NW_DDIC_Password.needDDICPasswords = true
NW_DDIC_Password.ddic000Password = {{ sap_swpm_ddic_000_password | default(sap_swpm_master_password) }}
#NW_DDIC_Password.ddic001Password =
{% endif %}
```

`defaults/main.yml`
```yaml
sap_swpm_inifile_list:
...
#  - credentials_hdbuserstore
  - credentials_ddic_000
  - db_config_hana
....
```

sean-freeman:
As described, PR was replaced by PR #874

Issue #743, opened 2024-05-23
sap_general_preconfigure: dig command


sean-freeman:
sap_general_preconfigure: dig command

Ansible Task `Check resolv.conf settings` requires the FQDN to retrieve the A Record (i.e. `sap-hana.example.com`). Otherwise a CNAME Record is required if wanting to use short name only (i.e. `sap-hana`), which if I recall correctly is not recommended by SAP.

Ansible Task `Check dns reverse settings` requires a PTR Record.

In both cases I would suggest clearer Ansible Task names:
- `Check DNS Resolver in resolv.conf contains A Record for host`
- `Check DNS Resolver in resolv.conf contains reverse PTR Record for host`

I would suggest appending using the full `hostname -f` value for the first Ansible Task

berndfinger:
Solved by https://github.com/sap-linuxlab/community.sap_install/pull/765.

PR #765, opened 2024-06-20
sap_general_preconfigure: Don't ignore DNS A/PTR record inconsistencies, this should fail the check is negative

rob0d:
Also included a minor task description change to make it clearer what has failed.

### DIFF
--- a/roles/sap_general_preconfigure/tasks/RedHat/generic/check-dns-name-resolution.yml
+++ b/roles/sap_general_preconfigure/tasks/RedHat/generic/check-dns-name-resolution.yml
@@ -15,6 +15,6 @@
 
 - name: "Verify the correct DNS IP address to FQDN resolution (PTR record) for {{ sap_general_preconfigure_ip }}"
   ansible.builtin.shell: test "$(dig -x {{ sap_general_preconfigure_ip }} +short)" = "{{ sap_general_preconfigure_hostname }}.{{ sap_general_preconfigure_domain }}."
-  changed_when: false 
+  changed_when: false
   register: __sap_general_preconfigure_register_dns_test_ptr
   failed_when: __sap_general_preconfigure_register_dns_test_ptr.rc != 0
--- a/roles/sap_general_preconfigure/tasks/RedHat/generic/check-dns-name-resolution.yml
+++ b/roles/sap_general_preconfigure/tasks/RedHat/generic/check-dns-name-resolution.yml
@@ -4,20 +4,20 @@
   ansible.builtin.shell: test "$(dig {{ sap_general_preconfigure_hostname }}.{{ sap_general_preconfigure_domain }} +short)" = "{{ sap_general_preconfigure_ip }}"
   changed_when: false
  #ignore_errors: true
-  register: dug
-  failed_when: dug.rc != 0
+  register: __sap_general_preconfigure_register_dns_test_a
+  failed_when: __sap_general_preconfigure_register_dns_test_a.rc != 0
 
 ### BUG: dig does not use search path in resolv.con on PPCle
 - name: Check resolv.conf settings
   ansible.builtin.shell: test "$(dig {{ sap_general_preconfigure_hostname }} +search +short)" = "{{ sap_general_preconfigure_ip }}"
   changed_when: false
 #ignore_errors: true
-  register: dug
-  failed_when: dug.rc != 0
+  register: __sap_general_preconfigure_register_dns_test_resolvconf
+  failed_when: __sap_general_preconfigure_register_dns_test_resolvconf.rc != 0
 
 - name: Check reverse dns record (PTR)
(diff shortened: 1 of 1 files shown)


roles/sap_general_preconfigure/tasks/RedHat/generic/check-dns-name-resolution.yml
```diff
@@ -1,17 +1,23 @@
 ---
 
-- name: Check dns forwarding settings
+- name: Check forward dns record (A)
   ansible.builtin.shell: test "$(dig {{ sap_general_preconfigure_hostname }}.{{ sap_general_preconfigure_domain }} +short)" = "{{ sap_general_preconfigure_ip }}"
   changed_when: false
-  ignore_errors: true
+ #ignore_errors: true
+  register: dug
```
berndfinger, line 7: The default in the preconfigure roles is to prepend all variables defined by `register:` with `__<role>_register_`, so please modify the variable name to something similar to `__sap_general_preconfigure_register_dig_a` or `__sap_general_preconfigure_register_dns_resolve_a` .

roles/sap_general_preconfigure/tasks/RedHat/generic/check-dns-name-resolution.yml
```diff
@@ -1,17 +1,23 @@
...
+  failed_when: dug.rc != 0
 
 ### BUG: dig does not use search path in resolv.con on PPCle
 - name: Check resolv.conf settings
   ansible.builtin.shell: test "$(dig {{ sap_general_preconfigure_hostname }} +search +short)" = "{{ sap_general_preconfigure_ip }}"
   changed_when: false
-  ignore_errors: true
+#ignore_errors: true
+  register: dug
```
berndfinger, line 15: Likewise, modify the variable name to something similar to `__sap_general_preconfigure_register_dig_ptr` or `__sap_general_preconfigure_register_dns_resolve_ptr`.

sean-freeman:
Would suggest the Ansible Tasks are updated as suggested in #743 to correctly indicate to the end-user, that `dig` runs from the host to check the DNS Resolver contains the A/PTR records.

- `Check DNS Resolver in resolv.conf contains A Record for host`
- `Check DNS Resolver in resolv.conf contains reverse PTR Record for host`

Otherwise I'd say this PR is an easy one to merge 👍

berndfinger:
> Would suggest the Ansible Tasks are updated as suggested in #743 to correctly indicate to the end-user, that `dig` runs from the host to check the DNS Resolver contains the A/PTR records.
> 
> * `Check DNS Resolver in resolv.conf contains A Record for host`
> * `Check DNS Resolver in resolv.conf contains reverse PTR Record for host`

We should not include one specific DNS name resolution method because there are other methods which do not rely on an entry in /etc/resolv.conf. But I agree that the task names can be more detailed, as in:

* `Verify the correct DNS hostname to IP address resolution (A record) using the FQDN`
* `Verify the correct DNS hostname to IP address resolution (A record) using the hostname and the search list`
* `Verify the correct DNS IP address to FQDN resolution (PTR record)`

Just for the records: This check can be skipped by using the tag `sap_general_preconfigure_dns_name_resolution` in --`skip-tags`, so there should be no need to add another role parameter for this purpose.

roles/sap_general_preconfigure/tasks/RedHat/generic/check-dns-name-resolution.yml
```diff
@@ -1,17 +1,23 @@
...
-- name: Check dns forwarding settings
+- name: Check forward dns record (A)
   ansible.builtin.shell: test "$(dig {{ sap_general_preconfigure_hostname }}.{{ sap_general_preconfigure_domain }} +short)" = "{{ sap_general_preconfigure_ip }}"
   changed_when: false
-  ignore_errors: true
+ #ignore_errors: true
+  register: __sap_general_preconfigure_register_dns_test_a
+  failed_when: __sap_general_preconfigure_register_dns_test_a.rc != 0
 
 ### BUG: dig does not use search path in resolv.con on PPCle
```
berndfinger, line 10: typo: resolv.con -> resolv.conf

roles/sap_general_preconfigure/tasks/RedHat/generic/check-dns-name-resolution.yml
```diff
@@ -1,17 +1,23 @@
...
 - name: Check resolv.conf settings
   ansible.builtin.shell: test "$(dig {{ sap_general_preconfigure_hostname }} +search +short)" = "{{ sap_general_preconfigure_ip }}"
   changed_when: false
-  ignore_errors: true
+#ignore_errors: true
+  register: __sap_general_preconfigure_register_dns_test_resolvconf
```
berndfinger, line 15: We should maybe change the variable name to `__sap_general_preconfigure_register_dns_test_a_searchlist`

rob0d:
@berndfinger  @sean-freeman  I've updated it again to improve the text based on the latest recommendation from both of you. Is this acceptable now?

berndfinger (approved):
LGTM

PR #772, opened 2024-06-21
sap_install_media_detect: aws igw slow impact gpg key

sean-freeman:
sap_install_media_detect: aws igw slow impact gpg key

Consistent failures in 3 tasks where the AWS IGW performance is too slow and timeout occurs, causing Ansible to have a fatal error. Fixed with retries while waiting for the SNAT to work.


berndfinger (approved):
LGTM

PR #773, opened 2024-06-24
sap_install_media_detect: search known subdirs on re-run

sean-freeman:
sap_install_media_detect: search known subdirs on re-run

### DIFF
--- a/roles/sap_install_media_detect/tasks/prepare/move_files_to_main_directory.yml
+++ b/roles/sap_install_media_detect/tasks/prepare/move_files_to_main_directory.yml
@@ -1,12 +1,12 @@
 ---
 
-- name: SAP Install Media Detect - Prepare - Search known subdirectories - Initialize fact variables, phase 1b
+- name: SAP Install Media Detect - Prepare - Move files to parent for known subdirs - Initialize fact variables, phase 1b
   ansible.builtin.set_fact:
     __sap_install_media_detect_fact_find_result_phase_1b: []
 
 # If any files have been moved to non-extract subdirectories already, move them back to the top level, making the role idempotent
 # Reason for noqa: When using pipefail and there is no result from the grep -v, this tail will fail but it should never fail
-- name: SAP Install Media Detect - Prepare - Search known subdirectories - Find the relevant non-extract subdirectories # noqa risky-shell-pipe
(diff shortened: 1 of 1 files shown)


berndfinger:
Fixes issue https://github.com/sap-linuxlab/community.sap_install/issues/774.

roles/sap_install_media_detect/tasks/prepare/move_files_to_main_directory.yml
```diff
@@ -1,37 +1,40 @@
 ---
 
-- name: SAP Install Media Detect - Prepare - Initialize fact variables, phase 1b
+- name: SAP Install Media Detect - Prepare - Search known subdirectories - Initialize fact variables, phase 1b
```
berndfinger, line 3: When running the role, prepare/create_file_list_phase_2.yml is expecting all files on the top level directory. However, the role is moving all SAP archives into subdirectories as part of running organize_files.yml. In order to achieve idempotency, we therefore need to move any existing files in lower level directories back to the top level directory. That's the purpose of the task in this file.

Because of this, please either remove the string `Search known subdirectories` from the task names of your PR, or modify it to something similar to `Move files back to top level`.

Why are we doing these file move operations?
- If we left the files on the top level after the role has finished, there would be many files of different SAP file types in one directory, which we wanted to avoid. Note: By setting `sap_install_media_detect_move_or_copy_archives` to `false`, the role can be configured to **not** move SAP archive files to subdirectories.
- If we searched for all files in lower level directories (except the extraction directories), there might be duplicates on the top level and we would need to handle this case as well.

sean-freeman: How about the verbose `- Known subdirectories search and move files to parent directory -` ?

berndfinger: > How about the verbose `- Known subdirectories search and move files to parent directory -` ?

It would make the output lines very long and so the real task name could not be visible.

sean-freeman: That's why I went for something shorter, I'm not keen on just "Move files back to top level" because it has no context.

`- Move files to parent for known subdirs -` ?

berndfinger: > `- Move files to parent for known subdirs -` ?

OK for me.

berndfinger (approved):
LGTM

PR #779, opened 2024-06-25
docs(CONTRIBUTORS): update SUSE contributors file

Wabri:


sean-freeman:
@Wabri Very good spot, this file should be updated. It's right to keep a list of prior contributors from the origin code.

The file actually misses the last 5 names from the list in metadata > https://github.com/sap-linuxlab/community.sap_install/blob/main/galaxy.yml#L19

Wabri:
I think I'm done with it, there are some description that we want to change?

sean-freeman (approved):
@Wabri probably should update some of the text, but won't block the PR for that reason 👍 lgtm

marcelmamula (approved):
LGTM

Issue #780, opened 2024-06-26
sap_install_media_detect: reduce stdout with loop labels


sean-freeman:
sap_install_media_detect: reduce stdout with loop labels

berndfinger:
Solved in https://github.com/sap-linuxlab/community.sap_install/pull/781.

PR #781, opened 2024-06-26
sap_install_media_detect: append loop labels

sean-freeman:
sap_install_media_detect: append loop labels to address GH Issue #780


berndfinger (approved):
LGTM

Issue #784, opened 2024-06-27
sap_general_preconfigure: The strict DNS check enforcement of PR 765 causes failures in certain cloud environments


berndfinger:
So we better make the DNS check optional by default and introduce a new role parameter for failing the role if desired only.

berndfinger:
Note: The change was introduced by https://github.com/sap-linuxlab/community.sap_install/pull/765.

berndfinger:
Also the task names which contain role variables do not always show the correct content (depending on how the role was executed, e.g. when being called with `include_role`.

sean-freeman:
This issue can occur on different platforms, and may cause frustration for the end-user if it is outside of their control.

For Example, end user may only have access to DNS A Records in 1 DNS Zone provided via MS Azure Private DNS and `"Reverse DNS (PTR) records are not stored in a forward private DNS zone. Reverse DNS records are stored in a reverse DNS (in-addr.arpa) zone."`. Source: https://learn.microsoft.com/en-us/azure/virtual-network/virtual-networks-name-resolution-for-vms-and-role-instances

Making this optional will avoid a breaking change and ensure end users that want to be strict, can do so, and those who need more flexibility will see an error message but not stop the Ansible Role from continuing.

rob0d:
Currently I don't have access to any cloud environment where I could test it. I am not sure if GCP and AWS behave in the same way as Azure with regards to the PTR records. It seems that Azure behaviour is on the verge of non-compliance with RFC1912 section 2.1, but I guess it is understandable why they do that.
Are you able to confirm if Azure is returning two PTR records and one if them is correct? Also is there anyone who can check how GCP and AWS behave?

PR #785, opened 2024-06-27
sap_general_preconfigure: Revert to ignoring DNS failures

berndfinger:
Solves issue #784.

### DIFF
--- a/roles/sap_general_preconfigure/defaults/main.yml
+++ b/roles/sap_general_preconfigure/defaults/main.yml
@@ -146,6 +146,10 @@ sap_general_preconfigure_kernel_parameters: "{{ __sap_general_preconfigure_kerne
 sap_general_preconfigure_max_hostname_length: '13'
 # The maximum length of the hostname. See SAP note 611361.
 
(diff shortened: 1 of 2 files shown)


sean-freeman (approved):
lgtm 👍

berndfinger:
@rhmk Is not available this week, so I removed him from the list of reviewers.
@rob0d Please have a look at this PR and comment.

rob0d:
@berndfinger  yes, it would make it safer and more flexible to make the strict check optional and use something like `sap_general_preconfigure_dns_check_strict: false` as a default.
We can possibly also enhance the PTR check to search through all returned records in a hope that one will match as mentioned in #784 as it will possibly help with the Azure behaviour that @sean-freeman  mentioned. I'm not sure if it's worth the effort though.
I can have a look at this, but I am away until 15th July.

berndfinger:
I'm going to merge this PR now to get back the previous default behavior while also implementing an option to enforce strict DNS checking. We can continue discussing further options in the issue and see if we want or can improve the DNS checking later.

Issue #788, opened 2024-06-28
sap_hana_install: Do not fail if sap_hana_install_common_master_password is not set.


berndfinger:
See https://github.com/sap-linuxlab/community.sap_install/issues/787#issuecomment-2196565158.

berndfinger:
The other option is to set `sap_hana_install_master_password` to an empty string if `sap_hana_install_common_master_password` is not set.

berndfinger:
I have moved this issue to our backlog because it needs more investigation and does not appear to be urgent.

PR #789, opened 2024-06-28
sap_hana_install: Solve issue #788.

berndfinger:
We do not want to fail the role in task `Rename some variables used by hdblcm configfile` if `sap_hana_install_common_master_password` is not set.


sean-freeman (approved):
Makes sense. lgtm 👍

berndfinger:
Closing without merging because it needs more investigation related to possible side effects and because the issue does not appear to be urgent.

PR #795, opened 2024-07-07
sap_swpm: add error notes to dev doc

sean-freeman:
sap_swpm: add error notes to dev doc, include prettified versions of the (very long) error messages/strings

For obvious reason, these are some common error messages and do not encompass all error messages and the resulting fix.


marcelmamula (approved):
LGTM, I wish something like this was provided by SAP for all Basis people.

PR #806, opened 2024-07-16
sap_ha_pacemaker_cluster: Add override to use Classic SAPHanaSR agents

marcelmamula:
Description:

- Adds variable `sap_ha_pacemaker_cluster_saphanasr_angi_detection` that enables option to disable SAPHanaSR-angi detection to install Classic agents even on servers where Angi is available.


ja9fuchs (approved):
lgtm :+1:

Issue #752, opened 2024-06-12
Idempotency is missing when testing with molecule.


glavoix:
Dear Community,

We are in the process of implementing two roles from your collection:
_ sap_general_preconfigure
_ sap_hana_preconfigure

They are working as expected however we are running integration tests using molecule and the testing fails
during the idempotency phase as some role use OS command (like yum or sysctl), and therefore adding condition
 -> "changed_when: false" 
should be used for those specific task. Another way is to tag them with molecule-idempotence-notest as explained here:
https://ansible.readthedocs.io/projects/molecule/configuration/#ansible

Here are a few tasks where we found that case:
https://github.com/sap-linuxlab/community.sap_install/tree/main/roles/sap_hana_preconfigure/tasks/sapnote/2382421.yml#L70
https://github.com/sap-linuxlab/community.sap_install/tree/main/roles/sap_hana_preconfigure/tasks/sapnote/3024346.yml#L27

https://github.com/sap-linuxlab/community.sap_install/tree/main/roles/sap_general_preconfigure/tasks/RedHat/generic/configure-kernel-parameters.yml#L15

https://github.com/sap-linuxlab/community.sap_install/tree/main/roles/sap_general_preconfigure/tasks/RedHat/installation.yml#L118

Would it be possible to add such a tag or condition ?

Thanks,
Guillaume

kksat:
@glavoix 

You might have a look into roles general_preconfigure and hana_preconfigure in sap.sap_operations collection

https://galaxy.ansible.com/ui/repo/published/sap/sap_operations/content/?showing=role

Documentation is here: 
https://docs.galaxy.saponrhel.org/collections/sap/sap_operations/hana_preconfigure_role.html#ansible-collections-sap-sap-operations-hana-preconfigure-role

https://docs.galaxy.saponrhel.org/collections/sap/sap_operations/general_preconfigure_role.html#ansible-collections-sap-sap-operations-general-preconfigure-role

Roles are idempotent and tested with molecule (see documentation how reboot is handled)

cc @berndfinger @sean-freeman

berndfinger:
Hi @glavoix - please have a look at https://github.com/sap-linuxlab/community.sap_install/pull/807.

berndfinger:
Solved in https://github.com/sap-linuxlab/community.sap_install/pull/807.

PR #807, opened 2024-07-17
Fixes for testing with molecule

berndfinger:
These changes should ensure that the molecule idempotency tests no longer fail, see issue #752.

### DIFF
--- a/roles/sap_general_preconfigure/handlers/main.yml
+++ b/roles/sap_general_preconfigure/handlers/main.yml
@@ -43,6 +43,7 @@
 - name: Remount /dev/shm # noqa command-instead-of-module no-changed-when
   ansible.builtin.command: mount -o remount /dev/shm
   listen: __sap_general_preconfigure_mount_tmpfs_handler
+  tags: molecule-idempotence-notest
 
 - name: Check if /dev/shm is available
   ansible.builtin.command: df -h /dev/shm
--- a/roles/sap_general_preconfigure/tasks/RedHat/generic/configure-kernel-parameters.yml
+++ b/roles/sap_general_preconfigure/tasks/RedHat/generic/configure-kernel-parameters.yml
@@ -18,7 +18,8 @@
   register: __sap_general_preconfigure_register_sap_conf_sysctl_command
   changed_when: false
 
-- name: Get all currently active values of the parameters of file '{{ sap_general_preconfigure_etc_sysctl_sap_conf }}'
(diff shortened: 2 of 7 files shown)


ja9fuchs:
Why the overhead of executing "x entries + 1" commands per sysctl file and not just `sysctl -p <file>` without those checks first? The reload does not make changes as such, it only activates what has been defined. 

Since the checks do not validate the content of the config file, there is no benefit to checking if all of the settings are already active or not. If one deviates, it will also reload the whole file, not just the deviating entry. 

It's not 100% idempotent unless you add even more overhead to check and load each entry individually. 
But again, there is no benefit of doing this, unless there is an actual sanity verification of the settings to prevent a mis-configuration from being loaded.

berndfinger:
> Why the overhead of executing "x entries + 1" commands per sysctl file and not just `sysctl -p <file>` without those checks first? The reload does not make changes as such, it only activates what has been defined.

We have three different sysctl files to modify according to applicable SAP notes, and because we want to follow the SAP notes as closely as possible, we do the modification for each file separately. For each file, we perform the modification of the file and then reload the parameters from the file (using sysctl -p)  so that it becomes effective immediately.

> Since the checks do not validate the content of the config file, there is no benefit to checking if all of the settings are already active or not. If one deviates, it will also reload the whole file, not just the deviating entry.

I am not checking if all of the settings of a config file are active or not. I want to know if any, and which one(s) of the entries or which of the current settings have been modified after reloading all the settings.

So we need to compare the desired state with the current state. The solution I offered appears to be robust and easy to understand. Happy to implement a better solution if there is one!

> It's not 100% idempotent unless you add even more overhead to check and load each entry individually. But again, there is no benefit of doing this, unless there is an actual sanity verification of the settings to prevent a mis-configuration from being loaded.

We want to detect if there was a change or not. A misconfiguration should result in a task error in the tasks `Reload kernel parameters from file XXX`, so I don't think we need to do a sanity verification.

But maybe I misunderstood some of your points. In this case, can you please provide examples?

glavoix:
Dear @berndfinger 
After reviewing your commit, I believe the approach looks nice and we should now be able to enable "idem-potency" into our molecule testing.

Thanks a lot for working on it.

Guillaume

ja9fuchs (approved):
Thanks @berndfinger for the comments and explanation. Now I get why these tasks are there. :) 

:+1:

PR #809, opened 2024-07-17
sap_*preconfigure: edge case handling for sles pkgs

sean-freeman:
For SLES, `sapconf` should be stopped but will error if the package does not exist and further errors for `hdblcm` if the base OS Image does not include `insserv-compat`.


berndfinger (commented):
LGTM.

marcelmamula (approved):
LGTM, I just tested it and it works!

PR #820, opened 2024-07-22
fix(sap_hana_preconfigure): Update azure override readme

### LINKED ISSUE #819: sap_hana_preconfigure/SLES: Update azure override readme and conditional
Description:

- Update readme, defaults and arguments file https://github.com/sap-linuxlab/community.sap_install/blob/c1b479ef710f0f0803a5a081eafd621465c86d4b/roles/sap_hana_preconfigure/README.md?plain=1#L429 with extra details:
```readme
If the variable is set, an override file for saptune will be created (/etc/saptune/override/2382421)
to set net.ipv4.tcp_timestamps,
net.ipv4.tcp_tw_reuse and net.ipv4.tcp_tw_recycle to 0.
```

- Investigate possibility to add OS_MAJOR version switch as parameter `net.ipv4.tcp_tw_recycle` is no longer valid since SLES 15 GA. SLES 12 and older could use it, but they are no longer mentioned as supported versions.
```note
Remove net.ipv4.tcp_tw_recycle from readme if it is removed from configuration
```

- Remove old commented out code present in https://github.com/sap-linuxlab/community.sap_install/blob/c1b479ef710f0f0803a5a081eafd621465c86d4b/roles/sap_hana_preconfigure/tasks/SLES/assert-configuration.yml#L42 because it is already present in https://github.com/sap-linuxlab/community.sap_install/blob/c1b479ef710f0f0803a5a081eafd621465c86d4b/roles/sap_hana_preconfigure/tasks/SLES/configuration.yml#L94

Wabri:
Closes #819

### DIFF
--- a/roles/sap_hana_preconfigure/README.md
+++ b/roles/sap_hana_preconfigure/README.md
@@ -420,6 +420,7 @@ The saptune solution to apply (SLES for SAP Applications).<br>
 - _Default:_ `false`
 
 On Azure, TCP timestamps, reuse and recycle should be disabled (SLES for SAP Applications).<br>
+If the variable is set, an override file for saptune will be created (/etc/saptune/override/2382421) to set net.ipv4.tcp_timestamps and net.ipv4.tcp_tw_reuse to 0.<br>
 Set this parameter to `true` on Azure.<br>
 
 <!-- END: Role Input Parameters for sap_hana_preconfigure -->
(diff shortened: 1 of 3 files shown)


marcelmamula (approved):
LGTM

sean-freeman (approved):
lgtm 👍

Issue #823, opened 2024-07-23
collection: Synchronize all .yamllint.yml files


berndfinger:
These files just need to be synchronized with the collection level `.yamllint.yml`, so that `ansible-lint` on a role, when executed locally, does not report the `yamllint` violations any more.

Relates to https://github.com/sap-linuxlab/community.sap_install/issues/810 and https://github.com/sap-linuxlab/community.sap_install/pull/811.

berndfinger:
Fixed in #824.

PR #811, opened 2024-07-17
collection: Modify for latest yamllint requirements

berndfinger:
ansible-lint 24.7.0 produced:

```
WARNING  Found incompatible custom yamllint configuration (.yamllint.yml), please either remove the file or edit it to comply with:
  - comments.min-spaces-from-content must be 1
  - octal-values.forbid-implicit-octal must be true
  - octal-values.forbid-explicit-octal must be true.
```

This PR ensures that the collection is yamllint clean again.

Relates to issue #810.

Issue #810, opened 2024-07-17
collection: Fix 'yaml[octal-values]' 


berndfinger:
New yamllint rule demands quotes around octal values. See https://ansible.readthedocs.io/projects/lint/rules/yaml/#octals.

berndfinger:
Solved by #811 and #824.

PR #824, opened 2024-07-23
collection: Synchronize all .yamllint.yml files

berndfinger:
Solves issue #823.
Relates to https://github.com/sap-linuxlab/community.sap_install/issues/810 and https://github.com/sap-linuxlab/community.sap_install/pull/811.

### DIFF
--- a/roles/sap_anydb_install_oracle/.yamllint.yml
+++ b/roles/sap_anydb_install_oracle/.yamllint.yml
@@ -8,7 +8,9 @@ rules:
   brackets: {max-spaces-inside: 1, level: error}
 #  colons: {max-spaces-after: -1, level: error}
 #  commas: {max-spaces-after: -1, level: error}
-  comments: disable
+  comments:
+    require-starting-space: false
+    min-spaces-from-content: 1
   comments-indentation: disable
 #  document-start: disable
 #  empty-lines: {max: 3, level: error}
@@ -20,3 +22,6 @@ rules:
 #  new-lines: {type: unix}
 #  trailing-spaces: disable
   truthy: disable
+  octal-values:
+    forbid-implicit-octal: true
+    forbid-explicit-octal: true
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -1,3 +1,4 @@
+# SPDX-License-Identifier: Apache-2.0
 ---
 # Based on ansible-lint config
 extends: default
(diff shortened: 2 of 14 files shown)


berndfinger:
Hi @Wabri - Can you please have a look?

roles/sap_anydb_install_oracle/.yamllint.yml
```diff
@@ -8,7 +8,9 @@ rules:
   brackets: {max-spaces-inside: 1, level: error}
 #  colons: {max-spaces-after: -1, level: error}
 #  commas: {max-spaces-after: -1, level: error}
-  comments: disable
+  comments:
+    require-starting-space: false
```
Wabri, line 12: The min-spaces-from-content is good, but why the `require-starting-space`? Reading the documentation:

```
Use require-starting-space to require a space character right after the #. Set to true to enable, false to disable.
```

Even stick with the default value `true` is ok in our case, or not?

roles/sap_anydb_install_oracle/.yamllint.yml
```diff
@@ -20,3 +22,6 @@ rules:
 #  new-lines: {type: unix}
 #  trailing-spaces: disable
   truthy: disable
+  octal-values:
```
Wabri, line 25: This values are by default, do we want to keep just to be sure and to be more clear?

roles/sap_anydb_install_oracle/.yamllint.yml
berndfinger: As per the output I have shown in https://github.com/sap-linuxlab/community.sap_install/pull/811, ansible-lint 24.7.0 reported `incompatible custom yamllint configuration` and demanded `comments.min-spaces-from-content must be 1` if a custom yamllint configuration is desired (which we want/need).

When *not* also setting `require-starting-space: false` in the `comments:` section, ansible-lint will report many errors of type:
```
yaml[comments]: Too few spaces before comment
```

for each role.

roles/sap_anydb_install_oracle/.yamllint.yml
berndfinger: The only reason is that ansible-lint 24.7.0 demands that these values are specified, otherwise it will complain as I showed in the output in https://github.com/sap-linuxlab/community.sap_install/pull/811. So it seems that there is no alternative to my proposal. But if you can think of one or find one, please let me know so that we can change it. Maybe future ansible-lint versions behave differently - in this case we can remove these lines again.

Wabri: You are right, I forgot that warning. Sorry!

Wabri (approved):

PR #829, opened 2024-07-30
sap_ha_pacemaker_cluster: Stonith SBD enablement, stonith and repos dictionaries

marcelmamula:
**New features:**

- **Stonith SBD fence method added with all extra variables required for `ha_cluster`**

- Significant changes to Stonith task file, adding lot of extra validations to avoid execution if:
  - `ha_cluster` variables are provided, just skip them because you cannot override precedence anyway
  - variables are empty, not required list or string where list is expected (iterable check fails to differentiate between string and list)

- New repository dictionaries:
  - `__sap_ha_pacemaker_cluster_repos_dict` defined inside OS vars
  - Each platform vars loads `__sap_ha_pacemaker_cluster_repos` from dictionary
  - Dictionary is empty for SUSE as it comes with subscription.

- New fence agent dictionaries for each platform
  - `__sap_ha_pacemaker_cluster_stonith_default_dict` defined in each platform vars, then selected correct entry into `__sap_ha_pacemaker_cluster_stonith_default`
  - This serves as baseline for future if we have different fence agents like AWS, where RHEL and SLES have different agents.
  - There is potential for future to shift these dictionaries into new OS_stonith.yml var files if need arises for different OS parameters under same agents.

- Variable `sap_ha_pacemaker_cluster_aws_credentials_setup` that can be enabled to store AWS credentials. It is undefined by default to use attached IAM Role/Instance Profile done by AP4S.
  - Keeping AWS Credentials together with attached IAM, results in credentials overriding IAM permissions, requiring API user inside of AWS credentials to have fencing permissions.

- Default `ha_cluster` definition from AP4S moved inside of vars code, to create baseline nodes for `ha_cluster`
  - `pcs_address` is required only for `pcs`, but it does not harm `crmsh`
```yaml
    __sap_ha_pacemaker_cluster_ha_cluster:
      node_name: "{{ ansible_hostname }}"
      pcs_address: "{{ ansible_default_ipv4.address }}"
```

- Readme and argument_specs update with new variables and increased baseline version for required collections.

**Tested on:**

- AWS
  - SLES4SAP 15: SBD and default fencing
  - RHEL4SAP 8.6: Several tests executed, currently running SBD and default fence agent testing again.

### DIFF
--- a/roles/sap_ha_pacemaker_cluster/meta/argument_specs.yml
+++ b/roles/sap_ha_pacemaker_cluster/meta/argument_specs.yml
@@ -102,12 +102,6 @@ argument_specs:
               node_ip: 192.168.5.1
             - hana_site: DC02
 
-      sap_ha_pacemaker_cluster_fence_agent_minimal_packages:
-        type: list
-        default: ['fence-agents-all']
-        description:
-          - The minimal set of fence agent packages that will be installed.
-
       sap_ha_pacemaker_cluster_resource_defaults:
         type: dict
         default:
@@ -181,6 +175,8 @@ argument_specs:
               - The resource options listed in dictionary format, one option per line.
               - Requires the mandatory options for the particular stonith resource agent to be
                 defined, otherwise the setup will fail.
+              - "Example: stonith:fence_sbd agent requires devices option with list of SBD disks."
+              - "Example: stonith:external/sbd agent does not require devices option, but `sap_ha_pacemaker_cluster_sbd_devices`."
             required: true
 
         example:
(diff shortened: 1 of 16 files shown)


roles/sap_ha_pacemaker_cluster/meta/argument_specs.yml
```diff
@@ -194,6 +190,56 @@ argument_specs:
...
+
+        example:
+          sap_ha_pacemaker_cluster_sbd_devices:
+            - /dev/disk/by-id/scsi-3600
+
+      sap_ha_pacemaker_cluster_sbd_options:
+        type: list
+        description:
+          - Optional if `sap_ha_pacemaker_cluster_sbd_enabled` is enabled.
+          - Provide list SBD specific options that are added into SBD configuration file.
```
ja9fuchs, line 224: Typo: "list of..."

sean-freeman (approved):
Reviewed all commits and discussed in-depth, happy with changes to internal logic and will regression test once merged to `dev`

👍 Awesome work

ja9fuchs (approved):
:+1: 
Very nice work, thank you!

PR #839, opened 2024-08-21
sap_storage_setup: Add exact size disk check on top of approximate check

marcelmamula:
## Problem:
- Current disk assignment works on approximate validation with `(disk_size_gb-8) <= fs.disk_size <= (disk_size_gb+8)` which invalidates this role when filesystems with less than 8 GB differences are used.

- Current loop will pick first found, invalidating correct entries, causing cascading failure during LV creation because of `device: ""`

## Example:
Disks with sizes 15, 20 and 25 are used.

**/hana/shared will never be loaded, because `/hana/log` will find 25GB disk, use it and `/hana/shared` would be left without entry in following list of disks.`volume_map`**

```yaml
    sap_storage_setup_definition: [
      {
        "disk_size": 15,
        "filesystem_type": "xfs",
        "mountpoint": "/hana/data",
        "name": "hana_data"
      },
      {
        "disk_size": 20,
        "filesystem_type": "xfs",
        "mountpoint": "/hana/log",
        "name": "hana_log"
      },
      {
        "disk_size": 25,
        "filesystem_type": "xfs",
        "mountpoint": "/hana/shared",
        "name": "hana_shared"
      }
    ]
```

## Solution:
Add initial check for exact sizes, followed up by original approximate check which will run empty if all disk sizes were correctly found during first pass.

- First pass: `disk_size_gb == fs.disk_size`
- Second pass: `(disk_size_gb-8) <= fs.disk_size <= (disk_size_gb+8)`

### DIFF
--- a/roles/sap_storage_setup/tasks/generic_tasks/map_single_disks_to_filesystems.yml
+++ b/roles/sap_storage_setup/tasks/generic_tasks/map_single_disks_to_filesystems.yml
@@ -50,6 +50,8 @@
 # defined already that is to be enhanced with single disk definitions, if
 # applicable.
 
+# First pass assigns disks based on exact size.
+# Second pass assigns disks based on approximate size -8GB and +8GB
 - name: SAP Storage Setup - Set fact for target filesystem device mapping
   ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


ja9fuchs (approved):
lgtm, nice catch for tiny disk sizing
:+1:

PR #840, opened 2024-08-22
sap_swpm: New improved and simplified version

berndfinger:
I am providing an improved and simplified version of the role sap_swpm, with the following features:
- No longer use separate role modes while keeping the same functionality.
- Enable creating the sapinst input file `inifile.params` using both methods (inifile sections and direct parameter setttings), similar to combining the `default` and `advanced` modes of the previous version. In the first step (but only if the role parameter `sap_swpm_inifile_sections_list` is defined), the new sap_swpm version is creating the file `inifile.params` from the inifile sections in the template. Then, if the parameter `sap_swpm_inifile_parameters_dict` is defined, existing variable definitions can be modified, or additional variables and their settings can be added at the end of the file. This provides maximum flexibility for creating the file `inifile.params` while not increasing the complexity of the logic.
- No longer use the 3 or 4 level dictionary `sap_swpm_inifile_custom_values_dictionary` (for setting role variables) inside the role, as it was required for the previous modes `default_templates` and `advanced_templates`. The dictionary can still be used and maintained, by setting certain top level role parameters from dictionary members. Note: You cannot access low level dictionary members using a variable which has been defined in the same task. For defining mid-level variables of a dictionary, you have to use a separate task (e.g. `set_fact`) before.
- Enable creating a sapinst `inifile.params` parameter input file within a few seconds (using tag `sap_swpm_generate_inifile`), without running pre_install steps. This file can then be transferred to managed nodes, to be re-used later, even by the role `sap_swpm` for running unattended installations.
- Similar to the role `sap_hana_install` (parameter `sap_hana_install_configfile_directory`), the new version of sap_swpm is now checking for an existing installer input file on the managed node (parameter `sap_swpm_inifile_directory`), not on the control node.


berndfinger:
Note: I successfully tested with ap4s scenarios `sap_s4hana_sandbox` and `sap_s4hana_standard` on AWS. Testing with `sap_s4hana_distributed_ha` is still ongoing.

roles/sap_swpm/README.md
```diff
@@ -63,58 +63,66 @@ This role has been tested and working for the following SAP products
...
+
+#### sap_swpm_inifile_custom_values_dictionary -> sap_swpm_inifile_parameters_dict
+Previous name: sap_swpm_inifile_custom_values_dictionary
+New name: sap_swpm_inifile_parameters_dict
+Reason: This variable contains parameter names and values of the sapinst input file, `inifile.params`.
+The new variable name is reflecting this purpose.
+
+#### Migration from the `*_templates` modes of the previous version of sap_swpm
+The role `sap_swpm` does no longer use the dictionary `sap_swpm_templates_install_dictionary`.
+This dictionary was used in the previous role modes `default_templates` and `advanced_templates.
```
marcelmamula, line 97: Some formatting is needed:  
```
`advanced_templates. > `advanced_templates`.
```

roles/sap_swpm/README.md
```diff
@@ -63,58 +63,66 @@ This role has been tested and working for the following SAP products
...
+The following two role parameters have been renamed, without automatic conversion between old and new name:
+
+#### sap_swpm_inifile_list -> sap_swpm_inifile_sections_list
+Previous name: sap_swpm_inifile_list
+New name: sap_swpm_inifile_sections_list
+Reason: This variable contains sections of the sapinst input file, `inifile.params`.
+The new variable name is reflecting this purpose.
+
+#### sap_swpm_inifile_custom_values_dictionary -> sap_swpm_inifile_parameters_dict
+Previous name: sap_swpm_inifile_custom_values_dictionary
```
marcelmamula, line 90: </br> is missing in these lines resulting in confusing one line.

roles/sap_swpm/tasks/pre_install.yml
```diff
@@ -81,3 +74,32 @@
...
       - '   # tail -f $(cat /tmp/sapinst_instdir/.lastInstallationLocation)/sapinst.log'
+
+################
+# Set sapinst command
+################
+
+- name: SAP SWPM Pre Install - Set the sapinst command parameters
+  vars:
+    sap_swpm_swpm_command_guiserver: "{{ 'SAPINST_START_GUISERVER=false' if not sap_swpm_swpm_observer_mode else '' }}"
+    sap_swpm_swpm_command_observer: "{{ 'SAPINST_REMOTE_ACCESS_USER=' + sap_swpm_swpm_remote_access_user + ' SAPINST_REMOTE_ACCESS_USER_IS_TRUSTED=true' if sap_swpm_swpm_observer_mode and sap_swpm_swpm_remote_access_user is defined and sap_swpm_swpm_remote_access_user | length > 0 else '' }}"
```
marcelmamula, line 85: There will be merge conflict with https://github.com/sap-linuxlab/community.sap_install/pull/850/files

marcelmamula (changes requested):
Some small changes needed, while I continue testing.

roles/sap_swpm/README.md
berndfinger: What do you think is missing, and where?

marcelmamula: updated my comment as it was considered part of markdown

berndfinger: So you suggest adding an empty line after all `###` and `####` sections which do not have one?

marcelmamula: Yes, because it seems that intent was to have clean lines based on readme file.

```markdown
#### sap_swpm_inifile_list -> sap_swpm_inifile_sections_list
Previous name: sap_swpm_inifile_list
New name: sap_swpm_inifile_sections_list
Reason: This variable contains sections of the sapinst input file, `inifile.params`.
The new variable name is reflecting this purpose.
```

But actual Readme preview shows correctly that markdown does not do newline automatically
```markdown
sap_swpm_inifile_list -> sap_swpm_inifile_sections_list
Previous name: sap_swpm_inifile_list New name: sap_swpm_inifile_sections_list Reason: This variable contains sections of the sapinst input file, inifile.params. The new variable name is reflecting this purpose.
```

berndfinger: @marcelmamula I now see what you mean after previewing the README.md file. The lines below he caption are concatenated but they should remain separate.

marcelmamula (commented):
Additional changes that would be good to include:
1. remove /etc/hosts steps and replace them with /etc/hosts role instead.
2. Correct non-standard variable inputs like `sap_swpm_use_password_file == "y"`

I found them while updating Readme.

rhmk (approved):
LGTM

marcelmamula (changes requested):

roles/sap_swpm/tasks/pre_install/assert_hostname_resolution_for_ha.yml
```diff
@@ -0,0 +1,48 @@
...
+      failed_when: false
+
+    - name: SAP SWPM Pre Install - HA settings - Assign sap_swpm_virtual_ip from /etc/hosts
+      ansible.builtin.set_fact:
+        __sap_swpm_fact_virtual_ip: "{{ __sap_swpm_register_virtual_ip_etc_hosts.stdout_lines[-1] }}"
+      when: __sap_swpm_register_virtual_ip_etc_hosts.stdout_lines | length > 0
+
+- name: SAP SWPM Pre Install - HA settings - Assert that sap_swpm_virtual_hostname can be resolved
+  ansible.builtin.assert:
+    that: __sap_swpm_fact_virtual_ip | length > 0
```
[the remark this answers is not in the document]
berndfinger: Should no longer occur after implementing the changes requested for the task `HA settings - Assert that sap_swpm_virtual_hostname can be resolved` (see above).

roles/sap_swpm/tasks/pre_install/assert_hostname_resolution_for_ha.yml
```diff
@@ -0,0 +1,48 @@
...
+
+- name: SAP SWPM Pre Install - HA settings - Try to resolve sap_swpm_virtual_hostname from DNS
+  ansible.builtin.command: dig +short +tries=1 +time=1 "{{ sap_swpm_virtual_hostname }}"
+  register: __sap_swpm_register_virtual_ip_dns
+  changed_when: false
+  failed_when: false
+
+- name: SAP SWPM Pre Install - HA settings - Assign sap_swpm_virtual_ip from DNS
+  ansible.builtin.set_fact:
+    __sap_swpm_fact_virtual_ip: "{{ __sap_swpm_register_virtual_ip_dns.stdout_lines[-1] }}"
```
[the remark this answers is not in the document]
berndfinger: In my tests, `dig +short +tries=1 +time=1` resulted in two lines of outputs: one line displaying the FQDN and another line displaying the IPv4 address. Can you show me an example of another output which is also shows a valid DNS hostname resolution but with a different number and maybe order of records?

roles/sap_swpm/tasks/pre_install/assert_hostname_resolution_for_ha.yml
```diff
@@ -0,0 +1,48 @@
...
+  when: __sap_swpm_register_virtual_ip_dns.stdout_lines | length > 0
+
+- name: SAP SWPM Pre Install - HA settings - Try using /etc/hosts for name resolution
+  when: __sap_swpm_register_virtual_ip_dns.stdout_lines | length == 0
+  block:
+
+    - name: SAP SWPM Pre Install - HA settings - Try to resolve sap_swpm_virtual_hostname from /etc/hosts
+      ansible.builtin.shell: |
+        awk '/\s{{ sap_swpm_virtual_hostname }}\./||/\s{{ sap_swpm_virtual_hostname }}/{print $1}' /etc/hosts
```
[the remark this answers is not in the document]
berndfinger: - We don't need `set -o pipefail` because there is no pipe in the shell command.
- Searching for a string without the surrounding white spaces or a following dot (`.`) is not sufficient because we want to search for the exact hostname. I agree on excluding the comment character in column 1 so we do not catch lines which are part of a comment.
- We can accept an empty string as the result of the `awk` command because the task `Assert that sap_swpm_virtual_hostname can be resolved` will fail with the message `FAIL: HOST cannot be resolved` if `__sap_swpm_fact_virtual_ip` is empty after the previous task (`HOST` showing the hostname to be resolved).

roles/sap_swpm/tasks/pre_install/assert_hostname_resolution_for_ha.yml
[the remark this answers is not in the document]
marcelmamula: @berndfinger My comment was about /etc/hosts check where you can have invalid duplicates, but it seems I copied wrong link pointing to dig. Current awk command does not work properly.

```console
je1scs:~ # awk '/\sje1scsha\./||/\sje1scsha/{print $1}' /etc/hosts
192.168.2.20
#
8.8.8.8

je1scs:~ # grep je1scsha /etc/hosts
192.168.2.20    je1scsha.suseaws.com    je1scsha
# 123 je1scsha
8.8.8.8 je1scsha
je1scs:~ #
```

marcelmamula (approved):
LGTM, Thank you @berndfinger for patiently testing and fixing all issues with me to hopefully resolve all issues before merging.

ja9fuchs:
Successfully tested an ASCS/ERS installation. :+1:

PR #848, opened 2024-08-30
sap_hana_install: Run hdbuserstore via su

### LINKED ISSUE #847: sap_hana_install: Run the hdbuserstore command via su from superuser
When running the `sap_hana_install` role using a non-root user, the task `Create and Store Connection Info in hdbuserstore` fails with `Missing sudo password` if the credentials for connecting to the `<sid>adm` user have not been provided before, because the task is running with `<sid>adm` as the `become_user`.

This can be avoided by calling the `hdbuserstore` command via `su` from the `<sid>adm` user.

berndfinger:
Fixes #847.

### DIFF
--- a/roles/sap_hana_install/tasks/post_install/hdbuserstore.yml
+++ b/roles/sap_hana_install/tasks/post_install/hdbuserstore.yml
@@ -3,14 +3,13 @@
 
 - name: SAP HANA Post Install - Create and Store Connection Info in hdbuserstore
   ansible.builtin.shell: |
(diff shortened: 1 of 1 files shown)


marcelmamula (changes requested):
Doing su/sudo inside shell/command is not best, but did you try `omit` ?

```yaml
become: "{{ true if ansible_user_id == 'root' else omit }}"
become_user: "{{ (sap_hana_install_sid | lower ~ 'adm') if ansible_user_id == 'root' else omit }}"
```

berndfinger:
There is probably more to consider here, so closing for now.

PR #849, opened 2024-08-30
sap_hana_install: Implement an SAP HANA installation check only feature

### LINKED ISSUE #844: sap_hana_install: Allow the role to only perform an installation check
By adding the tag `sap_hana_install_check_installation` to the task `Rename some variables used by hdblcm configfile`, we can quickly enable the role `sap_hana_install` to only check an existing installation.

Users would have to call the role with role parameter `sap_hana_install_check_installation` and also include the tag with this name. Note: Without the tag, the role would perform the installation check in addition to the installation attempt.

The role could of course also be redesigned to only perform an installation check without using the tag - for example using a new role parameter `sap_hana_install_check_installation_only`

berndfinger:
Fixes #844.

With this change, the role sap_hana_install can be used to perform an SAP HANA installation check without performing an installation.

### DIFF
--- a/roles/sap_hana_install/tasks/main.yml
+++ b/roles/sap_hana_install/tasks/main.yml
@@ -10,6 +10,7 @@
     sap_hana_install_restrict_max_mem: "{{ sap_hana_install_mem_restrict | d(sap_hana_install_restrict_max_mem) }}"
   tags:
     - sap_hana_install_check_hana_exists
+    - sap_hana_install_check_installation
     - sap_hana_install_preinstall
     - sap_hana_install_set_log_mode
     - sap_hana_install_configure_firewall


marcelmamula:
@berndfinger Name of this PR and Issue gives me hint that this detects already installed HANA on target, but that is not case based on code that debugs variables, not actual target contents.

I would suggest maybe renaming it to something clearer.

berndfinger:
> @berndfinger Name of this PR and Issue gives me hint that this detects already installed HANA on target, but that is not case based on code that debugs variables, not actual target contents.

The current version of the role can already check an existing HANA installation using one of these two methods, and it is doing this in the task named `Perform an hdblcm installation check` DIRECTLY AFTER a SAP HANA installation has been finished.

This PR, however, is only about adding the related tag to one more task so that the role does not fail when doing ONLY an installation check, in those cases where variables are used which have to be converted first.

marcelmamula (approved):
LGTM

PR #856, opened 2024-09-04
sap_install_media_detect: Allow disabling RAR handling

### LINKED ISSUE #855: sap_install_media_detect: Add an option for skipping handling RAR files
We observed that when running the role on a RHEL system with the file `/etc/yum/vars/releasever` containing the string `9.4`, the installation of the `unar` package from EPEL fails:

```
TASK [community.sap_install.sap_install_media_detect : SAP Install Media Detect - Prepare - EPEL - Install the unar package] ***************
fatal: [hana-p]: FAILED! => changed=false 
  msg: 'Failed to download metadata for repo ''epel'': Cannot prepare internal mirrorlist: Status code: 404 for https://mirrors.fedoraproject.org/metalink?repo=epel-9.4&arch=x86_64&infra=$infra&content=$contentdir (IP: 1.2.3.4)'
  rc: 1
  results: []
```

This issue can be worked around by installing a package for handling rar files separately and setting the role parameter `sap_install_media_detect_rar_package` accordingly.

As the `unar` package is only required for listing the contents of, and extracting, self-extracting RAR files, we should however implement an option for ignoring the RAR handling.

berndfinger:
Fixes #855.

### DIFF
--- a/roles/sap_install_media_detect/defaults/main.yml
+++ b/roles/sap_install_media_detect/defaults/main.yml
@@ -1,6 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 ---
 
+# Set this parameter to `false` for skipping the handling of RAR files. In this case, also no unar or other RAR handling software
+# will be installed.
+sap_install_media_detect_rar_handling: true
+
 # Set this parameter to use either the unar package from EPEL or another software package for handling RAR files.
 # Based on this setting, the commands for listing and extracting RAR files are being set in tasks/prepare/enable_rar_handling.yml
 sap_install_media_detect_rar_package: 'EPEL'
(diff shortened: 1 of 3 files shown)


berndfinger:
Note: This additional option should only be used when it is known that there are no RAR or self-extracting RAR/EXE files in the `sap_install_media_detect_source_directory`. This is the case for all SAP HANA and SAP S/4HANA installations, to my knowledge. In the ap4s, we are installing such systems and in this case only downloading the required files.

berndfinger:
My tests with ap4s-scenarios `sap_hana` and `sap_s4hana_standard` finished successfully, using a RHEL 9.4 AMI.

ja9fuchs (approved):
lgtm and works well in a test on RHEL 9.4 with `sap_install_media_detect_rar_handling: false`. Before the fix and the parameter it failed due to the epel repo issue.

PR #866, opened 2024-09-16
  sap_ha_install_hana_hsr: fixes to work for multiple secondaries

ja9fuchs:
- improved some loops over the cluster nodes list using `selectattr`
- new task to validate that exactly 1 primary is defined and at least 1 secondary
- throttle 'secondary' tasks that execute on the primary at the same time otherwise and lead to conflicts
- some tasks only need to run once

This enables the base setup of Multi-Target System Replication.

Tested with 1 primary and 2 secondaries:
```
 sap_ha_install_hana_hsr_cluster_nodes:
      - hana_site: DC1
        node_name: su1hana1
        node_ip: <private node ip>
        node_role: primary
      - hana_site: DC2
        node_name: su1hana2
        node_ip: <private node ip>
        node_role: secondary
      - hana_site: DC3
        node_name: su1hana3
        node_ip: <private node ip>
        node_role: secondary
```

### DIFF
--- a/roles/sap_ha_install_hana_hsr/tasks/configure_hsr.yml
+++ b/roles/sap_ha_install_hana_hsr/tasks/configure_hsr.yml
@@ -32,6 +32,7 @@
 
 # looping through cluster definition to run on secondary node
 # and apply the respective 'site' value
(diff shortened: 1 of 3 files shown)


marcelmamula (approved):
LGTM, Tested on my cluster in AWS.

PR #872, opened 2024-09-25
fix: vip resources must be first in ASCS/ERS resource groups

ja9fuchs:
Fixing the order of the resources in ASCS and ERS resource groups.

The VIP resource has to be the first to be started and the last to be stopped.
This will align the automation with the documentation by all providers.

### Tested
RHEL 9.4 on AWS


marcelmamula (approved):
LGTM

Issue #868, opened 2024-09-19
sap_netweaver_preconfigure: libcpupower0 is not present for SLES 15 SP6 anymore


herrmie:
Hello,

in the role sap_netweaver_preconfigure there will be libcpupower0 installed. This package is not present in SLES 15 SP6, it is called libcpupower1 now.

Please adapt the code for SLES 15 SP6.

Thanks in advance and best regards,
Steffen Herrmann

marcelmamula:
Thank you @herrmie for reporting this issue.
Fix is now part of https://github.com/sap-linuxlab/community.sap_install/pull/876 and It will be merged to dev branch once it is reviewed.

PR #876, opened 2024-10-11
sap_netweaver_preconfigure/SUSE: Fix for issue 868 specific to SLES4SAP15 SP6

marcelmamula:
Several changes were added to both `sap_netweaver_preconfigure` and `sap_hana_preconfigure` roles to resolve issue reported in https://github.com/sap-linuxlab/community.sap_install/issues/868

- Add new var file for `SLES_SAP_15.6`
- Change order for var file loading to ensure that actual version is loaded first, instead of major distribution that was used before.</br> **Major + Minor combination var file would be ignored without this change!**
- Update saptune version flow in `sap_netweaver_preconfigure` to align with changes done in `sap_hana_preconfigure` before.
- Fix `yes/no` values in `sap_netweaver_preconfigure` defaults file causing linting errors.


Wabri (approved):
LGTM

Wabri (approved):

PR #877, opened 2024-10-14
sap_netweaver_preconfigure: remove specific version for saptune

remrozk:


remrozk:
Thank you @marcelmamula, #876 is better 👍

PR #878, opened 2024-10-18
sap_hana_preconfigure: add a condition to disable sapconf

remrozk:


marcelmamula:
@remrozk What OS version did you use that `sapconf` was not present?

remrozk:
Hi @marcelmamula, I am using SLES for SAP 15 SP5. Out of the box, the ISO does not include this package.

When attempting to install without a SUSE license, the task fails.  This task does not need to be executed if the package is not present.

However, I see that PR #809 has already made this improvement. Therefore, I am closing this pull request.

marcelmamula:
> Hi @marcelmamula, I am using SLES for SAP 15 SP5. Out of the box, the ISO does not include this package.
> 
> When attempting to install without a SUSE license, the task fails. This task does not need to be executed if the package is not present.
> 
> However, I see that PR #809 has already made this improvement. Therefore, I am closing this pull request.

@remrozk  That is good to hear. We have lot of changes waiting in dev branch so it would be worth checking it out first as we are preparing another release to main branch and galaxy.

Issue #879, opened 2024-10-21
sap_hana_preconfigure: Set THP to madvise from RHEL 9.2 onwards


berndfinger:
... as per SAP notes [2131662](https://me.sap.com/notes/2131662) (and [3108302](https://me.sap.com/notes/3108302)).

berndfinger:
Solved in https://github.com/sap-linuxlab/community.sap_install/issues/880.

See also related issue https://github.com/sap-linuxlab/community.sap_install/issues/885 and PR https://github.com/sap-linuxlab/community.sap_install/pull/886.

PR #880, opened 2024-10-21
sap_hana_preconfigure: Set THP to madvise from RHEL 9.2 onwards

berndfinger:
Solves issue #879 .


berndfinger:
We better also assert the current setting of THP and set it on the running system, as we also do for KSM.

rhmk (approved):
LGTM

PR #882, opened 2024-10-25
sap_ha_pacemaker_cluster: JAVA HA scenarios and complete refactor

marcelmamula:
## Purpose
1. Add SAP Netweaver SCS/ERS cluster scenario
2. Simplify variables to align with SAP naming (ASCS is ABAP SCS, there is no reason to specify ABAP ASCS as it is contained in its name).

## Changes

1. Added SCS/ERS ENSA1 and ENSA2 Classic and Simple Mount (same code as ASCS since there are no differences in config itself)

3. Refactor of all variables with `*nwas_abap_*` to `*nwas_*` to work with java and to make more sense in regards to SAP naming.
   - `*nwas_abap_*` is retained only as part of `sap_ha_pacemaker_cluster_host_type`
```yaml
nwas_abap_ascs_ers
nwas_abap_pas_aas
nwas_java_scs_ers
```

4. Refactor of private variables to remove all calculations from defaults and vars. All of them are moved to include_vars tasks separately so we can do
   - Use backwards compatible variables until they are removed
   - Use user variables if specified
   - Default back to empty string or calculated variable of user did not specify.

5. New asserts to cover mandatory variables, that were mandatory before, but never validated with assert.

6. Added variables for future proofing PAS/AAS, but not actual cluster setup code.

7. argument_spec and readme updated with variables that were not listed before (health checks, VIP, etc.)

8. Added `DEPRECATED_VARIABLES.md` with list of changed variables and their status.

### Operating system:
- SUSE Linux Enterprise Server for SAP applications 15 SP5
- SUSE Linux Enterprise Server for SAP applications 15 SP6

### Platforms:
- AWS
- Azure
- Google Cloud

### Tested scenarios:
- SAP HANA HA
- SAP ASCS/ERS
- SAP SCS/ERS

### DIFF
--- a/roles/sap_ha_pacemaker_cluster/DEPRECATED_VARIABLES.md
+++ b/roles/sap_ha_pacemaker_cluster/DEPRECATED_VARIABLES.md
@@ -1,0 +1,53 @@
+# sap_ha_pacemaker_cluster Deprecated input variables
+
+This deprecation of input variables is part of an ongoing effort to improve the codebase by removing unnecessary elements and streamlining the overall design.
+
+These variables fall into a few categories:
+- **Obsolete or unused**
+  - These variables are no longer used and are being removed to reduce technical debt and potential confusion.
+- **Renamed**
+  - These variables are being renamed to better reflect their current purpose and improve code readability.
+  - This is especially important when the variable's functionality has evolved over time.
+
+## Backwards compatibility
+All deprecated variables offer time limited backwards compatibility that will be removed in future.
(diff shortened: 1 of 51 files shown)


ja9fuchs (approved):
Tested installation and compared results:
- HANA cluster using deprecated vars
- ASCS/ERS cluster using deprecated vars
- ASCS/ERS cluster using new vars with custom values

No surprises, everything looked fine. :+1:

Issue #883, opened 2024-10-31
collection: ansible-lint does not like `latest` in requirements.yml


berndfinger:
I tested using `ansible-lint` versions `24.9.0` and `24.9.2`.

The version of `requirements.yml` producing this error contains:
```
  - name: community.general
    version: latest
    type: galaxy
```

The error message when running `ansible-lint` on the collection is:
```
ERROR! Unexpected Exception, this is probably a bug: Non integer values in LooseVersion ('latest')
```

The version of `requirements.yml` not producing this error contains:
```
  - name: community.general
    version: '>=9.5.0'
    type: galaxy
```

Note: The collection `community.general` version `9.5.0` is currently the latest one and has been published on Ansible Galaxy 23 days ago, see https://galaxy.ansible.com/ui/repo/published/community/general/.

marcelmamula:
@berndfinger This looks like very strange issue with state data type.

berndfinger:
I think it's better to specify a minimum version.

PR #884, opened 2024-10-31
collection: Specify minimum version for dependent collections

berndfinger:
Solves issue #883.


rhmk (approved):
LGTM

marcelmamula (approved):
LGTM, it might be safer to do `9.0.0` but this will go into release where `9.5.0` might be already n-1 or more.

Issue #689, opened 2024-03-21
sap_netweaver_preconfigure: verify compatibility with the applicable SAP notes for Adobe Document Services


berndfinger:
https://github.com/sap-linuxlab/community.sap_install/blob/d36cc155dc82fdb1848b1f36d6dbbedb9d06a3ab/roles/sap_netweaver_preconfigure/vars/RedHat_9.yml contains `libssh2`, which is not available on RHEL 9. So let's use this opportunity to verify all Adobe Document Services packages with the applicable SAP notes now.

berndfinger:
Solved in #888.

PR #888, opened 2024-11-18
sap_netweaver_preconfigure: Sync with applicable SAP notes for Adobe DS

berndfinger:
Solves issue #689.

Further changes in this commit:
- Add the SAP note numbers and versions in a comment before the list of packages in the RedHat*.yml files in vars.

- Add 'ansible_architecture' as a condition in the related installation task, so that there is no task failure when running on another architecture than 'x86_64'.

- Also revert back to the default of 'false' for the role variable 'sap_netweaver_preconfigure_use_adobe_doc_services'. See commit 384f4d82, PR #876.


ja9fuchs (approved):

marcelmamula (approved):
@berndfinger @ja9fuchs  LGTM, but it is good to start planning readme update to reflect `(RedHat specific)` variables so it is not confusing. I have updated `(SUSE specific)` already, but this one is good example of `(RedHat specific)` variable

PR #892, opened 2024-11-21
fix reboot in sap_generel_preconfigure

crysaki:
`changed_when: false` while notify the reboot handler will always cause to not call the handler, so I've changed to `changed_when: true`


marcelmamula (approved):
LGTM

berndfinger:
@crysaki Thanks for reporting this!

berndfinger:
... and also for suggesting the fix!

Issue #894, opened 2024-11-21
sap_hana_install: HANA 2.0 SPS08 needs the installation execution mode 'optimized'.


berndfinger:
When installing HANA 2.0 SPS08 on RHEL 9.2, the installation failed early with: `Installation of SAP HANA Local Secure Store is possible only in optimized execution mode`.

After adding the hdblcm command line parameter `--install_execution_mode=optimized` in [this task](https://github.com/sap-linuxlab/community.sap_install/blob/c1b479ef710f0f0803a5a081eafd621465c86d4b/roles/sap_hana_install/tasks/hana_install.yml#L28) and also defining `sap_hana_install_master_password` in addition to `sap_hana_install_common_master_password`, the installation of SPS08 on RHEL 9.2 finished successfully.

See also the chapter `SAP HANA Server Installation and Update (New and Changed)` in the `What's New in SAP HANA Platform 2.0` document on https://help.sap.com/)

berndfinger:
The problem can be also solved by using one of the following methods (both without changing hdblcm command line parameters):

- In your playbook or inventory, set: `sap_hana_install_install_execution_mode: 'optimized'`
- If you do not intend to install the LSS component, you can leave the default installation execution mode to `standard`. For not installing the LSS component, you can specify the HANA components to be installed by the role as in the following example: `sap_hana_install_components: server,client`

berndfinger:
Note: Specifying `sap_hana_install_components: all,!lss` for selecting all but one component of HANA 2 SPS08 does not work and will result in the error message:
`Value "!lss" is invalid. Use one of: all, server, client, lss`

berndfinger:
Solved in https://github.com/sap-linuxlab/community.sap_install/pull/896.

PR #896, opened 2024-11-22
sap_hana_install: Set the install execution mode to 'optimized'

berndfinger:
This is required when installing the HANA component 'LSS', which is the default in HANA 2.0 SPS08. The installation execution mode can be set to the previous value of 'standard' by using the role variable 'sap_hana_install_install_execution_mode'. See the file `defaults/main.yml` of this role.

Solves issue #894.

### DIFF
--- a/roles/sap_hana_install/README.md
+++ b/roles/sap_hana_install/README.md
@@ -374,7 +374,11 @@ With the following tags, the role can be called to perform certain activities on
 
 <!-- BEGIN Further Information -->
 ## Further Information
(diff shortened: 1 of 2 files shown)


marcelmamula (approved):
LGTM, Tested on SAP HANA HA SP07

ja9fuchs (approved):
:+1: 

Successfully tested installations for SPS07 and SPS08 with default settings.

```
| System     | Version         |         | 2.00.078.00.1715149848 (fa/hana2sp07)       |
| System     | Platform        |         | Red Hat Enterprise Linux 9.4 (Plow)         |
```

```
| System     | Version         |         | 2.00.080.00.1731056231 (fa/hana2sp08)       |
| System     | Platform        |         | Red Hat Enterprise Linux 9.4 (Plow)         |
```

PR #904, opened 2024-11-29
sap_ha_pacemaker_cluster: fixes for RHEL on MS Azure

ja9fuchs:
- Correction of release in repo name, otherwise this fails on anything later than RHEL 8.
- From RHEL 9 on the azure-lb RA is in a separate package. The same fix was already applied for AWS previously.


sean-freeman (approved):
lgtm 👍

sean-freeman (approved):
lgtm 👍

PR #910, opened 2024-12-09
sap_install_*_preconfigure, sap_install_ha*: Vars loading reworked, Linting fixes for yes/no

marcelmamula:
## Description:
1. Add vars file loading in line with `linux-system-roles` to load in order and remove duplicated variables.
2. Correct linting errors of `yes, no` across preconfigure roles
3. Removed `expect` dependency from `sap_ha_pacemaker_cluster` role for Suse post tasks.

```yaml
# Load variable files in order:
# Example for SUSE Linux Enterprise Server for SAP Applications 15 SP6:
# 1. Suse.yml - Specific to OS family.
# 2. SLES_15.yml - Specific to distribution (SLES and SLES_SAP) and major release.
# 3. SLES_15.6.yml - Specific to distribution (SLES and SLES_SAP) and minor release.
# 4. SLES_SAP_15.yml - Specific to distribution SLES_SAP and major release.
# 5. SLES_SAP_15.6.yml - Specific to distribution SLES_SAP and minor release.
```

## Next steps
- @berndfinger Please check RedHat var files to simplify them in separate PR as we discussed: Put shared parameters into `RedHat.yml` and version specific in their own files.


roles/sap_netweaver_preconfigure/defaults/main.yml
```diff
@@ -5,7 +5,7 @@
 # Perform an assertion run:
 sap_netweaver_preconfigure_assert: false
 
-# In case of an assertion run, if set to "yes", the role will abort for any assertion error:
+# In case of an assertion run, if set to "true", the role will abort for any assertion error:
```
berndfinger, line 8: Let's also change the double quotes for single ones:
`# In case of an assertion run, if set to 'true', the role will abort for any assertion error:`

berndfinger (approved):
LGTM for the RHEL related and non-HA parts.

ja9fuchs (approved):
All changes look fine.
Cluster setup tested and all good.
Vars loading tested and the output is looking good now too.
:+1:

PR #911, opened 2024-12-09
sap_ha_pacemaker_cluster: ANGI on RHEL and small improvements

ja9fuchs:
## Description
1. "ANGI" now also works on RHEL
   - RHEL pre-steps discover the package but will not resolve the package conflict if the classic agents are installed. A failure message explains the issue and the user has to solve it. This will prevent an unintended change of the entire setup if run against pre-configured nodes.
   - Adjustment of the VIP order constraint task to work with classic and 'angi'.

2. Adjustment of hook vars logic
   - Explicitly define hook preferences for `_tkover` and `_chksrv` vars in OS defaults.
   - The hook task now respects the hook control variable and the argument now works for user choices, overriding the OS preference.
   - The presence of the hook definition in the OS vars does not enforce the hook config anymore. This enables the definition of optional hooks.
   - In RHEL the `_chksrv` hook is also available and supported. The hook is optional, for customers to enable on demand.

3. A few cosmetical improvements of task names
4. Suspected typo fixed for `__sap_ha_pacemaker_cluster_hanacontroller_resource_clone_name`
   - I could not find a reference to `mst_` and the argument default info also refers to `cln_` as the prefix.

## Tested
- ... on RHEL 9.4 in AWS for SAP HANA HA with 'classic' and 'angi' resource agents.
- Verified the hook enable choice to configure the hook that is not enabled by default on RHEL.

### DIFF
--- a/roles/sap_ha_pacemaker_cluster/tasks/RedHat/pre_steps_hana.yml
+++ b/roles/sap_ha_pacemaker_cluster/tasks/RedHat/pre_steps_hana.yml
@@ -1,7 +1,51 @@
 # SPDX-License-Identifier: Apache-2.0
 ---
-# Identify if SAPHanaSR-angi package is available for installation.
-# SAPHanaSR-angi replaces SAPHanaSR and SAPHanaSR-ScaleOut.
+# Identify if 'sap-hana-ha' package is available for installation.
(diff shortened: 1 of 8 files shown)


marcelmamula (approved):
LGTM with `mst/cln` changes

PR #914, opened 2024-12-16
sap_general_preconfigure - Checking for english locales, issue 907

rhmk:
This pull request checks that English locales are installed and set as the default. Additionally, a new variable allows for the enforcement of a specific English locale. This pull request is compatible with RHEL and SLES.


berndfinger (approved):
LGTM!

roles/sap_general_preconfigure/tasks/sapnote/2369910.yml
```diff
@@ -0,0 +1,53 @@
...
+      register: __sap_general_preconfigure_locales_installed
+
+    - name: Assert that an English locale is installed
+      ansible.builtin.assert:
+        that: __sap_general_preconfigure_locales_installed.stdout_lines | select('match', '^en_') | list | length > 0
+        fail_msg: "FAIL: No English locale is installed. Please install an English locale!"
+        success_msg: "PASS: An English locale is installed."
+
+    - name: Get the current default locale
+      ansible.builtin.command: awk '/^LANG=/&&/en_/{print}' /etc/locale.conf
```
Klaas-, line 28: this does not take into account that C.UTF-8 is also english

rhmk: You are right. Although C is English it is not sufficient for HANA2 SPS08. The en_* locales have to be installed. 
A test run showed that having the system locale C while having en_* installed was sufficient.
I am not sure how we can deal with it. For now I suggest accepting C and en_* locales, as long as en_* is installed. IIRC SAP recommends en_US.UTF-8, so if some body has a proof for that and put here or to the issue would be great.

berndfinger: suggested new `awk` statement:
`awk '/^LANG=/&&(/C.UTF-8/||/en_/){print}' /etc/locale.conf`

roles/sap_general_preconfigure/tasks/sapnote/assert-2369910.yml
```diff
@@ -0,0 +1,36 @@
...
+      register: __sap_general_preconfigure_locales_installed
+
+    - name: Assert that an English locale is installed
+      ansible.builtin.assert:
+        that: __sap_general_preconfigure_locales_installed.stdout_lines | select('match', '^en_') | list | length > 0
+        fail_msg: "FAIL: No English locale is installed. Please install an English locale!"
+        success_msg: "PASS: An English locale is installed."
+
+    - name: Get the current default locale
+      ansible.builtin.command: awk '/^LANG=/&&/en_/{print}' /etc/locale.conf
```
berndfinger, line 28: suggested new `awk` statement:
`awk '/^LANG=/&&(/C.UTF-8/||/en_/){print}' /etc/locale.conf`

rhmk: Change made and also fixed an issue with the variable check to set the locale

roles/sap_general_preconfigure/meta/argument_specs.yml
```diff
@@ -341,3 +341,11 @@ argument_specs:
           sap_general_preconfigure_db_group_name: 'dba'
         required: false
         type: str
+
+      sap_general_preconfigure_default_locale:
+        description:
+          - Use this variable to specify the default system locale.
+        example:
+          sap_general_preconfigure_db_group_name: 'en_US.UTF-8'
```
Klaas-, line 349: I am guessing you mean:
sap_general_preconfigure_default_locale: 'en_US.UTF-8' -- or C.UTF-8 depending what you think should be the "default" recommendation :)

Klaas- (commented):
I did not run it (on holidays already) -- but from reading the code this now looks good to me (except for the small c&p problem in the example definition)

roles/sap_general_preconfigure/meta/argument_specs.yml
rhmk: Thanks @Klaas- fixed this one, in addition @berndfinger and I found a bug in the algorithm so that this was variable was never used in certain circumstances (non-English locale set as default would always stop although variable was set)
This variable above is just an example, per default nothing is changed (variable empty in defaults/main.yml)

rhmk: I also removed the code from being executed on SLES because the reviewers are not available for the next weeks -- We may need to open another PR to enable this on SLES

rhmk (commented):
Line 31 needs improvement UTF-8 and utf8 can be used synonymously but locale -a always lists utf8

roles/sap_general_preconfigure/tasks/sapnote/2369910.yml
```diff
@@ -0,0 +1,52 @@
...
+      changed_when: false
+      register: __sap_general_preconfigure_locales_installed
+
+    - name: Assert that an English locale is installed
+      ansible.builtin.assert:
+        that: __sap_general_preconfigure_locales_installed.stdout_lines | select('match', '^en_') | list | length > 0
+        fail_msg: "FAIL: No English locale is installed. Please install an English locale!"
+        success_msg: "PASS: An English locale is installed."
+
+    - name: Configure English locale
```
berndfinger, line 27: I tested and now suggest to use the following code (including the comments to explain the logic a bit):
```
# We are setting a new locale from sap_general_preconfigure_default_locale if:
# - the new locale starts with "en_" or if it is "C.UTF-8" AND
# - the new locale (replacing "UTF-8" by "utf8" if necessary) is part of the installed locales
    - name: Configure English locale
      when:
        - sap_general_preconfigure_default_locale is defined
        - sap_general_preconfigure_default_locale | length > 0
        - sap_general_preconfigure_default_locale.startswith('en_') or sap_general_preconfigure_default_locale == 'C.UTF-8'
        - __sap_general_preconfigure_locales_installed.stdout_lines | select('match', sap_general_preconfigure_default_locale | regex_replace('UTF-8', 'utf8')) | join('') | trim | length > 0
``` 
It allows for setting an English locale using either the `UTF-8` or `utf8` notation (the output of `locale -a` contains only `utf8` on RHEL 8 and RHEL 9) and also avoids setting an unsupported locale.

If there is an SAP supported OS which has `UTF-8` in the output of `locale -a`, we could of course add the `regex_replace` filter also after `__sap_general_preconfigure_locales_installed.stdout_lines`.

roles/sap_general_preconfigure/tasks/sapnote/2369910.yml
```diff
@@ -0,0 +1,56 @@
...
+    - sap_general_preconfigure_configure_locale
+  block:
+    - name: Get list of installed locales
+      ansible.builtin.command: locale -a
+      changed_when: false
+      register: __sap_general_preconfigure_locales_installed
+
+    - name: Assert that an English locale is installed
+      ansible.builtin.assert:
+        that: __sap_general_preconfigure_locales_installed.stdout_lines | select('match', '^en_') | list | length > 0
```
berndfinger, line 23: Maybe this one is better, as it looks for ' en_[A-Z]*' AND Unicode:
```
__sap_general_preconfigure_locales_installed.stdout_lines | regex_findall('en_[A-Z]*[.](utf8|UTF-8)') | length > 0
```

berndfinger: Update: As we only have to take into account operating systems on which the command `localectl` is available (RHEL 7+ and SLES15+), we do not have to check if an English locale is installed. Starting with RHEL 8 and SLES15, the `localectl set-locale` command is checking if a locale is also installed. For RHEL 7 managed nodes, we can cover this topic in the documentation.

So we can remove the tasks `Get list of installed locales` and `Assert that an English locale is installed`.

roles/sap_general_preconfigure/tasks/sapnote/2369910.yml
berndfinger: Update: With `localectl set-locale`, we can use the following code:
```
    - name: Configure an English locale
      when:
        - sap_general_preconfigure_default_locale is defined and sap_general_preconfigure_default_locale
        - sap_general_preconfigure_default_locale.startswith('en_') or 
          sap_general_preconfigure_default_locale == 'C.UTF-8' or
          sap_general_preconfigure_default_locale == 'C.utf8' 
      ansible.builtin.command: "localectl set-locale LANG={{ sap_general_preconfigure_default_locale }}"
```

roles/sap_general_preconfigure/tasks/sapnote/2369910.yml
```diff
@@ -0,0 +1,56 @@
...
+        path: /etc/locale.conf
+        regexp: ^LANG=
+        line: "LANG=\"{{ sap_general_preconfigure_default_locale }}\""
+        state: present
+        create: true
+        owner: root
+        group: root
+        mode: '0644'
+
+    - name: Get the current default locale
```
berndfinger, line 47: Update: We can perform a better validation with the following awk command:
```
      ansible.builtin.command: awk '{gsub("\"","")}/^LANG=/&&(/=C\./||/=en_/)&&(/utf8$/||/UTF-8$/){print}' /etc/locale.conf
```

roles/sap_general_preconfigure/tasks/sapnote/2369910.yml
```diff
@@ -0,0 +1,56 @@
...
+      ansible.builtin.command: awk '/^LANG=/&&(/C.UTF-8/||/en_/){print}' /etc/locale.conf
```
berndfinger, line 48: Update: We can perform a better validation with the following awk command:
```
      ansible.builtin.command: awk '{gsub("\"","")}/^LANG=/&&(/=C\./||/=en_/)&&(/utf8$/||/UTF-8$/){print}' /etc/locale.conf
```

Issue #915, opened 2024-12-17
sap_swpm: nw_java_import.buildJEEusingExtraMileTool templated as "True"


Galaxy8977:
When installing SAP NW750 JAVA or SOLMAN72SR2 JAVA instances,

In roles/sap_swpm/templates/inifile_params.j2, the nw_java_import.buildJEEusingExtraMileTool is templated from the follow line:

```
nw_java_import.buildJEEusingExtraMileTool = {{ true if sap_swpm_java_import_method == 'extramile' else false }}
```

This leads to the result of 

```
nw_java_import.buildJEEusingExtraMileTool = True
```

And SWPM actually think it is not `true` and then going to the `Deploy Controller Runner tool` logic, the sap_swpm role eventually errored out with the described error in `docs/developer_notes/sap_swmp_dev.md`.

```log
#### Error with the Deploy Controller Runner tool

When setting inifile.params `nw_java_import.buildJEEusingExtraMileTool = false` an error will occur with the Deploy Controller Runner tool:

[ LOG EXCEPTION ]  :: Exception while generating ID: [[ERROR CODE DPL.DC.3218] Deploy controller is not ready to serve yet. Probably a un/deploy operation after offline phase is in progress at the moment. DC state is Initialized]
com.sap.engine.services.dc.cm.DCNotAvailableException: [ERROR CODE DPL.DC.3218] Deploy controller is not ready to serve yet. Probably a un/deploy operation after offline phase is in progress at the moment. DC state is Initialized

[ Error ] Exception while generating ID: [[ERROR CODE DPL.DC.3218] Deploy controller is not ready to serve yet. Probably a un/deploy operation after offline phase is in progress at the moment. DC state is Initialized]

[ LOG EXCEPTION ]  :: com.sap.engine.services.dc.api.deploy.DeployException: [ERROR CODE DPL.DCAPI.1023] DCNotAvailableException.
Reason: [ERROR CODE DPL.DC.3218] Deploy controller is not ready to serve yet. Probably a un/deploy operation after offline phase is in progress at the moment. DC state is Initialized
Caused by: com.sap.engine.services.dc.cm.DCNotAvailableException: [ERROR CODE DPL.DC.3218] Deploy controller is not ready to serve yet. Probably a un/deploy operation after offline phase is in progress at the moment. DC state is Initialized

[ LOG EXCEPTION ]  :: java.lang.NullPointerException: while trying to get the length of a null array loaded from a local variable at slot 4

[...]

PR #916, opened 2024-12-17
sap_swpm: Use 'false' instead of 'False' in inifile.params

berndfinger:
Fixes issue #915.


ja9fuchs (approved):
looks fine to me

sean-freeman (approved):
lgtm, previous code was hardcoded to `true` string value without Jinja logic.  Oversight when appending additional logic.

Issue #909, opened 2024-12-06
sap_swpm: master password


sean-freeman:
Ansible Task `SAP SWPM Pre Install - Set other user passwords using master password` should have a conditional, otherwise it would override the input values provided by the end-user

berndfinger:
Solved in #920.

PR #920, opened 2025-01-02
sap_swpm: Use master password only when necessary

berndfinger:
Solves issue #909.

### DIFF
--- a/roles/sap_swpm/tasks/pre_install/generate_inifile.yml
+++ b/roles/sap_swpm/tasks/pre_install/generate_inifile.yml
@@ -105,12 +105,18 @@
         sap_swpm_db_schema_password: "{{ sap_swpm_db_schema_java_password }}"
       when: "'Java' in sap_swpm_product_catalog_id"
 
+# If the individual passwords are set to a non empty string, use those:
     - name: SAP SWPM Pre Install - Set other user passwords using master password
       ansible.builtin.set_fact:
-        sap_swpm_sapadm_password: "{{ sap_swpm_master_password }}"
-        sap_swpm_sap_sidadm_password: "{{ sap_swpm_master_password }}"
-        sap_swpm_diagnostics_agent_password: "{{ sap_swpm_master_password }}"
-      when: sap_swpm_master_password is defined and sap_swpm_master_password
+        sap_swpm_sapadm_password: "{{ sap_swpm_master_password
+          if sap_swpm_master_password | d('') and not sap_swpm_sapadm_password | d('')
+          else sap_swpm_sapadm_password | d('') }}"
+        sap_swpm_sap_sidadm_password: "{{ sap_swpm_master_password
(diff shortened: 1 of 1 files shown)


roles/sap_swpm/tasks/pre_install/generate_inifile.yml
```diff
@@ -110,6 +110,7 @@
         sap_swpm_sapadm_password: "{{ sap_swpm_master_password }}"
```
marcelmamula, line 110: It should rather be like this to avoid overwrite. If you enable master, you still might want to use granular passwords.
```yaml
    - name: SAP SWPM Pre Install - Set other user passwords using master password
      ansible.builtin.set_fact:
        sap_swpm_sapadm_password: "{{ sap_swpm_sapadm_password | d(sap_swpm_master_password) }}"
        sap_swpm_sap_sidadm_password: "{{ sap_swpm_sap_sidadm_password | d(sap_swpm_master_password) }}"
        sap_swpm_diagnostics_agent_password: "{{ sap_swpm_diagnostics_agent_password | d(sap_swpm_master_password) }}"
      when: sap_swpm_master_password is defined and sap_swpm_master_password
```

berndfinger: @marcelmamula Makes sense. However, as these passwords are set to an empty string in `defaults/main.yml`, the default behavior of this code would be to set the following entries in `inifile.params`:
```
nwUsers.sidadmPassword =
DiagnosticsAgent.dasidAdmPassword =
hostAgent.sapAdmPassword =
```
So we should better also comment out these three variables in `defaults/main.yml`.

@sean-freeman Can you please have a look? Do you see an issue if we comment out all the individual passwords in `defaults/main.yml` of this role?

BTW All the individual password variables in role `sap_hana_install` are commented out as well.

marcelmamula: You are right, so we can comment them out as you mentioned or switch to IF/ELSE j2.
This way it will ignore master if it is defined and not empty, while be replaced when it is empty or undefined.

Following snippet was tested locally.
- when is not needed as it is incorporated
- d() and truthy check will cleanup undefined

```yaml
    - name: SAP SWPM Pre Install - Set other user passwords using master password
      ansible.builtin.set_fact:
        sap_swpm_sapadm_password: "{{ sap_swpm_master_password
          if not sap_swpm_sapadm_password | d('') and sap_swpm_master_password | d('')
          else sap_swpm_sapadm_password | d('') }}"
```

berndfinger: @marcelmamula Thanks. I prefer having the master password condition first, and also added a small comment so that one can quickly understand the rationale behind this code.
@sean-freeman With this code, there is no need to have a look, and to discuss commenting out all the individual passwords in `defaults/main.yml` of this role, at this point in time.

PR #874, opened 2024-10-08
sap_swpm: add functionality to set specific DDIC and SAP* passwords during new installation

rob0d:
Fix for issue #764 and replacement of pull request #763

PR #921, opened 2025-01-03
sap_swpm: add functionality to set specific DDIC and SAP* passwords during new installation

rob0d:
This is a replacement for PR #874 as I had to rebase and redo the changes due to too many changes in dev branch since I did the original PR.

### DIFF
--- a/roles/sap_swpm/defaults/main.yml
+++ b/roles/sap_swpm/defaults/main.yml
@@ -236,8 +236,13 @@ sap_swpm_java_scs_instance_hostname: ""
 sap_swpm_master_password:
 
 # New Install - define DDIC 000 password
-# Restore - DDIC 000 password from backup
+#  If not set, DDIC will be set to the master password
+# Restore - DDIC 000 password from
 sap_swpm_ddic_000_password:
+
+# New Install - define SAP* 000 password
+#  If not set, SAP* will be set to the master password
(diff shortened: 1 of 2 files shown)


sean-freeman:
@rob0d Accidental removal of keyword "backup from line `Restore - DDIC 000 password from backup`. Otherwise this is lgtm 👍

berndfinger:
I have no idea how an (BTW unrelated) commit can cause a PR to be closed automatically, needs to be investigated. Any assistance welcome.

rob0d:
@sean-freeman  @berndfinger 
Hi Bernd,
Sorry, I thought this was already merged. Doesn't look like it was then.
Yes this is still required as it's a missing functionality. The current behaviour  has confusing implications in terms what DDIC and SAP* passwords are set to if they are not supposed to be set to the master password (which is the case for real production installations).
Can this be re-opened and merged? 
There doesn't seem to be any conflict.

marcelmamula:
@rob0d This PR cannot be merged because there were significant changes in inifile j2 template file, where we added a lot of validations inside of template. Even if merge was technically working, it would be like adding code without required validations inside of something that was enhanced already.

I am also about to create PR with significant changes to role itself so it would make sense if you recreated it again with new version. @berndfinger Is currently unavailable for week or two.

PR #922, opened 2025-01-03
sap_ha_pacemaker_cluster: Allow to append manual cluster config to the generated config

rob0d:
@marcelmamula  @ja9fuchs 
This is based on the discussion we had several months ago with @ja9fuchs.

This role generally respects pre-defined ha_cluster* variables and tries to merge the configuration it generates with the content of the pre-defined variables. It works really well, however, in some cases the order of resources is important. 

The new variables allow to append a specific cluster configuration AFTER the configuration generated by this role.

I've added information to the README file which explains the behaviour. It's a non-breaking change as it just enhances the functionality without changing anything else.


marcelmamula:
@rob0d 
I think you need to provide example of how current code affects output, because appending post fact results in *advanced* mode, because you 100% trust that whoever uses it knows everything: dict format of `ha_cluster` variable.

roles/sap_ha_pacemaker_cluster/README.md
```diff
@@ -1167,4 +1167,20 @@ Name of the NetWeaver SCS resource group.<br>
 
 Name of the Virtual IP resource for NetWeaver Central Services (SCS).<br>
 
+### sap_ha_pacemaker_cluster_constraints_colocation_append <br>sap_ha_cluster_constraints_location_append <br>sap_ha_pacemaker_cluster_constraints_order_append <br>sap_ha_cluster_resource_clones_append <br>sap_ha_cluster_resource_primitives_append
+
+- _Type:_ `list`
+
+This role generally respects pre-defined ha_cluster* variables and tries to merge the configuration it generates with the content of the pre-defined variables. <br>
```
marcelmamula, line 1174: @rob0d These sentences are bit on the nose, because it contains too many negatives as it tells me: "Hey, this role barely works"

It has to be reworked if PR is to be accepted.

ja9fuchs:
This specific role has certain limits and will not be able to cover all custom cases. 
The use case in question is of such specific nature, that it should rather be covered by making further advanced use of the `ha_cluster` role native parameters. If the order of resources is critical, then the source definition can be pre-defined including the resources which the role would add - in the desired order. The SAP HA role checks for duplicate resource IDs (names) and in the cases of presumably unique resources it also skips if the resource agent is already present in the resource primitives parameter. The user defined primitive content should always take precedence.

As @marcelmamula already mentioned, just appending the variables in this role is not good enough due to the vars having to follow the correct syntax. It is not safe to do this without quite some effort of syntax validation and such logic, to make sure that what will be fed into `ha_cluster` is sane input.

Please share more details of your particular use case, so we could assess if it would justify putting effort into this enhancement.

**Note**: The variables section in the README is generated from `meta/argument_specs.yml` and vars content manually added to the README file will be lost later if not added there accordingly. I will follow up on highlighting this. You can't know if we don't tell. :)

rob0d:
Hi @marcelmamula @ja9fuchs,

Sorry, I probably should have explained this a bit more/better.
The key thing here is that ha_cluster doesn't allow modification of cluster settings. It will always recreate the cluster from scratch. So for any non-standard behaviours or advanced functionality we need to add those to` ha_cluster*` or `sap_ha_cluster*` variables if we want to automate cluster deployment using Ansible.

The two typical non-default functions is to add are: Ethmonitor to monitor if the network interface has gone down [(see)](https://access.redhat.com/solutions/2049063) or checking the network connectivity using the ping agent [(see)](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html-single/configuring_and_managing_high_availability_clusters/index#moving_resources_due_to_failure).  This is obviously just our usage case, other people may need other functionality.

The Ethmonitor is a must if the cluster uses multiple NICs (e.g. a separate heartbeat interface as generally recommended).
Everything is fine when this config is put in `sap_ha_cluster*` variables and we call **sap_ha_pacemaker_cluster** role which then merges the predefined config with our SAP config. The problem is that for e.g. ethmonitor we need define location constraints which will move ASCS if the network has failed. (Note: This situation does not result in cluster detecting the network failure if the cluster has multiple interfaces! The only solution according to Redhat is to use Ethmonitor resource agent.)
Unfortunately the order of resources and primitives definitions matter, so if we defined Ethmonitor first (with the location constraints) everything is fine until ha_cluster tries to create the cluster. This fails as the location constraint can't be defined as the resource we are referring to doesn't exist. This is not something **ha_cluster** or **sap_ha_pacemaker_cluster**  can safely and correctly detect without some serious coding.
Therefore I am suggesting that we allow the *_append variables for the user to define additional config on top of what **sap_ha_pacemaker_cluster**  generates.
I think that whether we allow user to predefine some config in `sap_ha_cluster*` / `ha_cluster*`  variables or in the _append variables, it is advanced usage. In both cases the users can put stuff in the variables that will break the execution or create a non-functional cluster (regardless of checks that are performed). I would suggest that if user is using the _append variables, it's their responsibility to get it ultimately right.  **ha_cluster** will perform checks and it's not necessarily responsibility of **sap_ha_pacemaker_cluster**  to make sure the settings in these variables are correct.

An example of how I'm using it is below. To test the current behaviour you can just remove the _append  and it should run  (and then fail during cluster creation) with the current code.

```
        sap_ha_cluster_resource_primitives_append:
          - id: ethmonitor-ens192
            agent: ocf:heartbeat:ethmonitor
            instance_attrs:
              - attrs:
                  - name: interface
                    value: ens192
                  - name: link_status_only
                    value: 'true'
 
        sap_ha_cluster_resource_clones_append:
          - resource_id: ethmonitor-ens192
 
        sap_ha_cluster_constraints_location_append:
          - resource:
              id: "grp_{{ sap_system_sid }}_ASCS{{ sap_ha_pacemaker_cluster_nwas_abap_ascs_instance_nr }}"
            rule: "ethmonitor-ens192 ne 1"
            id: "location-grp_{{ sap_system_sid }}_ASCS{{ sap_ha_pacemaker_cluster_nwas_abap_ascs_instance_nr }}"
            options:
              - name: score
                value: -INFINITY
          - resource:
              id: "grp_{{ wdp_system_sid }}_WDP{{ sap_ha_pacemaker_cluster_wdp_instance_nr }}"
            rule: "ethmonitor-ens192 ne 1"
            id: "location-grp_{{ wdp_system_sid }}_W{{ sap_ha_pacemaker_cluster_wdp_instance_nr }}"
            options:
              - name: score
                value: -INFINITY
```

roles/sap_ha_pacemaker_cluster/tasks/construct_final_hacluster_vars.yml
```diff
@@ -69,17 +69,17 @@
 - name: "SAP HA Prepare Pacemaker - (ha_cluster) Define parameter 'ha_cluster_constraints_colocation'"
   when: __sap_ha_pacemaker_cluster_constraints_colocation is defined
   ansible.builtin.set_fact:
-    ha_cluster_constraints_colocation: "{{ __sap_ha_pacemaker_cluster_constraints_colocation }}"
+    ha_cluster_constraints_colocation: "{{ __sap_ha_pacemaker_cluster_constraints_colocation  + sap_ha_pacemaker_cluster_constraints_colocation_append | d([]) }}"
```
marcelmamula, line 72: We cannot just invent new naming convention.
New variables have to be named same as role itself to pass linting `sap_ha_pacemaker_cluster_constraints_colocation_append`

Same applies to rest of variables for appending.

roles/sap_ha_pacemaker_cluster/README.md
marcelmamula, line 1174: These sentences are bit on the nose, because it contains too many negatives as it tells me: "Hey, this role barely works"

roles/sap_ha_pacemaker_cluster/README.md
```diff
@@ -1167,4 +1167,20 @@ Name of the NetWeaver SCS resource group.<br>
...
+However, in some cases the order of resources is important. The above variables allow to append a specific cluster configuration AFTER the configuration generated by this role. <br>
+
+The order in which cluster configuration is generated before it's passed to `ha_cluster` role is:
```
marcelmamula, line 1177: README.md is loaded from argument_spec file so changes here do not make sense.

These instructions can be part of section https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_ha_pacemaker_cluster/README.md#further-information and they have to be marked as EXPERIMENTAL.

marcelmamula (changes requested):
@rob0d @ja9fuchs 
I have tested this code (with changes mentioned in my comments) and it worked.
We should be fine after they are implemented as long as this is marked as experimental functionality.

Readme instructions should be added in Further Information using `<details><summary>` so it is available. Example: https://github.com/sap-linuxlab/community.sap_install/blob/9ee20c70df0bd34a2444b47b555729c4e91129bd/roles/sap_general_preconfigure/README.md?plain=1#L75

ja9fuchs (changes requested):
Looks like the previously requested changes were accidentally added to #929 instead.

Please add them here - without the ^M LF. See my comment in #929 about how to possibly remove and prevent those in the future. :)

rob0d:
@marcelmamula  @ja9fuchs fixed the readme, updated argument_specs and renamed variables.

roles/sap_ha_pacemaker_cluster/README.md
```diff
@@ -146,6 +146,22 @@ It is recommended to execute this role together with other roles in this collect
 
 <!-- BEGIN Further Information -->
 ## Further Information
+
+### sap_ha_pacemaker_cluster_*_append variables
```
marcelmamula, line 150: This section should be done same as I did it here: https://github.com/sap-linuxlab/community.sap_install/blob/9ee20c70df0bd34a2444b47b555729c4e91129bd/roles/sap_general_preconfigure/README.md?plain=1#L75

```md
<details>
  <summary>Appending custom cluster resources</summary>
  ...
</details>
```

roles/sap_ha_pacemaker_cluster/meta/argument_specs.yml
```diff
@@ -361,6 +361,80 @@ argument_specs:
...
+      ##########################################################################
+
+      sap_ha_pacemaker_cluster_constraints_colocation_append:
+        type: dict
+        description:
+          - EXPERIMENTAL
+          - Additional colocation constraints to be appended to the generated constraints.
+          - Equivalent to `ha_cluster_constraints_colocation` in the `ha_cluster` Linux System Role.
+          - Supported options can be reviewed in the `ha_cluster` Linux System Role
+            [https://github.com/linux-system-roles/ha_cluster/blob/master/README.md].
```
marcelmamula, line 375:
```md
        description:
          - EXPERIMENTAL
          - Additional colocation constraints to be appended after the constraints generated by this role.
          - Dictionary structure is based on `ha_cluster_constraints_colocation` in the `ha_cluster` Linux System Role.
          - Supported options can be reviewed in the `ha_cluster` Linux System Role
            [https://github.com/linux-system-roles/ha_cluster/blob/master/README.md].
```

Repeat same for other variables

marcelmamula (changes requested):

roles/sap_ha_pacemaker_cluster/tasks/construct_final_hacluster_vars.yml
```diff
@@ -105,7 +105,7 @@
 - name: "SAP HA Prepare Pacemaker - (ha_cluster) Define parameter 'ha_cluster_resource_clones'"
   when: __sap_ha_pacemaker_cluster_resource_clones is defined
   ansible.builtin.set_fact:
-    ha_cluster_resource_clones: "{{ __sap_ha_pacemaker_cluster_resource_clones }}"
+    ha_cluster_resource_clones: "{{ __sap_ha_pacemaker_cluster_resource_clones + sap_ha_pacemaker_cluster_resource_clones_append | d([]) }}"
 
 - name: "SAP HA Prepare Pacemaker - (ha_cluster) Define parameter 'ha_cluster_resource_groups'"
```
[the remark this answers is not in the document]
rob0d: Possibly yes. None of my usage cases needed groups. However, added.

roles/sap_ha_pacemaker_cluster/README.md
```diff
@@ -146,6 +146,22 @@ It is recommended to execute this role together with other roles in this collect
...
+`*EXPERIMENTAL*`<br>
+This role inherits pre-defined `ha_cluster*` variables and merges them with role specific parameters. <br>
+However, in some cases the order of resources is important. The `*_append` variables allow to append a specific cluster configuration AFTER the configuration generated by this role. <br>
```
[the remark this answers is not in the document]
rob0d: This is the order in which resources are created by ha_cluster. I've updated the text to clarify.

marcelmamula:
Hello @rob0d , we have not looked at this PR for long time because requested changes were not completed.
- Readme is main sticking point as it has to be really well sanitized to ensure everyone understands this is Experimental with zero support.
- Since these are hidden experimental options, having it in argument_spec file is not good as they  will be copied to README.md during README.md generation.

I have originally tested it when PR came out an it worked fine for adding some extra resources.

We had quite few commits in repository and role itself so it would have to be redone.
Let us know how you wan to proceed, but with issues mentioned above we cannot proceed with this PR with @ja9fuchs

PR #925, opened 2025-01-06
sap_swpm, sap_general_preconfigure: feat: add variables for sap_install collection name for calling roles

marcelmamula:
## Description
- Add variables for setting collection name to remove hardcode dependency.
- Affected: `sap_swpm` and `sap_general_preconfigure` for calling `sap_maintain_etc_hosts`

## Validation
Change was tested on SLES4SAP 15 SP6 on AWS with SAP ASCS/ERS HA build.

### DIFF
--- a/roles/sap_general_preconfigure/defaults/main.yml
+++ b/roles/sap_general_preconfigure/defaults/main.yml
@@ -31,6 +31,9 @@ sap_general_preconfigure_system_roles_collection: 'fedora.linux_system_roles'
 # - fedora.linux_system_roles
 # - redhat.rhel_system_roles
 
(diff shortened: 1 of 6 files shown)


berndfinger (approved):
LGTM!

sean-freeman:
Updated the name slightly (for easier reference in future), and added to description the history (inc. specific commit where FQCN var was added previously).

Personally I only see the need for a variable that alters the `namespace` of the FQCN, I do not know how many people will want to fork (into new namespace) and then also deviate from the `collection name` "sap_install".

marcelmamula:
@berndfinger @sean-freeman This was exact reason why I added it this way - same functionality as with `*_system_roles_collection`.

PR #931, opened 2025-01-14
sap_ha_pacemaker_cluster: enable Simple Mount on RHEL

ja9fuchs:
### New for RHEL
- Simple-Mount is enabled by default
- Netweaver pre-steps check the resource agents package version and disable Simple-Mount if the defined minimum version is not available

Tested with different package versions available.
The pre-step check overrules custom vars.


marcelmamula (approved):
LGTM, good job on enabling simple mount on your end as well!

PR #937, opened 2025-01-23
sap_general_preconfigure: No longer install locale packages in RHEL 7

berndfinger:
I am not aware of any issues related to missing English locales on RHEL 7, so we can safely assume that a RHEL 7 system will have all necessary English locale packages installed as part of installing the mandatory software environment group.


rhmk:
Well, nonetheless the role will complain, if the packages would be missing in a particular customer environment. Because of that I can approve the #PR.

rhmk (approved):
LGTM

Issue #939, opened 2025-01-23
sap_netweaver_preconfigure: argument spec validation failed


berndfinger:
We need to have:
```__sap_netweaver_preconfigure_packages: []```
in `vars/main.yml` - otherwise the argument_specs validation fails.

Introduced by https://github.com/sap-linuxlab/community.sap_install/pull/930 in this commit: https://github.com/sap-linuxlab/community.sap_install/pull/930/commits/04e099904217d5f1139629e65fe947f650470bd5 .

PR #940, opened 2025-01-23
sap_netweaver_preconfigure: fix argument_specs validation error

berndfinger:
Solves #939.

Also moved sap_netweaver_preconfigure_packages to a similar place as in the other preconfigure roles and added this variable to README.md


marcelmamula (approved):
LGTM

PR #947, opened 2025-01-29
sap_ha_pacemaker_cluster: fix package detection on RHEL

ja9fuchs:
The task return code was incorrect due to disabling `failed_when` entirely.
The new conditions should fix that.


marcelmamula (approved):
LGTM

PR #948, opened 2025-01-29
sap_general_preconfigure: fix var role prefix

ja9fuchs:
ansible-lint complained about this:

```
var-naming[no-role-prefix]: Variables names from within roles should use sap_general_preconfigure_ as a prefix. (vars: __sap_hana_preconfigure_use_saptune)
roles/sap_general_preconfigure/vars/SLES_SAP_16.yml:32
```


marcelmamula (approved):
@ja9fuchs Thank you for catching this, seems like my linting did not catch it when I was changing that many files at same time.

PR #952, opened 2025-02-03
sap_*_preconfigure/Suse: Switch saptune from present to latest

marcelmamula:
## Description
Some older images contain older saptune version `3.1.2` which will not be upgraded to latest with `present` state.

This change is ensuring actual latest package version is installed, which then helps with subsequent verification steps (example: kernel.pid_max works different on `3.1.2` versus `3.1.4`)

## Tested
Validated on SLES4SAP 15 SP6 on AWS:
- PAYG with saptune `3.1.2`
- BYOS with saptune `3.1.4`


berndfinger (approved):
LGTM.

PR #954, opened 2025-02-04
sap_ha_pacemaker_cluster: stonith location constraints

ja9fuchs:
Moved default stonith setup into a block to share conditionals.

**New task**
Adding default location constraints when fence resources are defined per host.
This prevents per-host-fence resources from running on the same, targeted node.

Currently GCP has a default stonith definition that works this way.

Result on GCP:
```
Location Constraints:
  resource 'rsc_fence_gce_hana01' avoids node 'hana01' with score INFINITY
  resource 'rsc_fence_gce_hana02' avoids node 'hana02' with score INFINITY
```
```
  * rsc_fence_gce_hana01        (stonith:fence_gce):     Started hana02
  * rsc_fence_gce_hana02        (stonith:fence_gce):     Started hana01
```

**Tested (on RHEL)**
GCP: correct default resource and constraints configuration
AWS: runs the default fence resource task, then skips the constraints task as expected


marcelmamula (approved):
LGTM. Naming could be improved, but it works.
Tested on SLES4SAP 15 SP6 on GCP.

PR #959, opened 2025-02-05
sap_ha_pacemaker_cluster: fix ASCS constraint

ja9fuchs:
Start order when ASCS and ERS are both down does not matter. 
However, After an ASCS failover, ERS should stay until ASCS has been started. ASCS needs to read enqueue data from ERS at start.

The "start order" conditional was incorrect by mistake.

RHEL and SLES documentations contain the order constraint as "start ASCS, then stop ERS".

_Tested on RHEL with "Simple Mount" setup._


marcelmamula (approved):
LGTM, Tested on SLES4SAP 15 SP6 in Simple Mount

`order ord_ascs_first_AE1 Optional: grp_AE1_ASCS00:start grp_AE1_ERS10:stop symmetrical=false`

PR #965, opened 2025-02-12
sap_ha_pacemaker_cluster: several bug fixes

ja9fuchs:
A few bugs started showing during new tests and need fixing.

The combination of fixes was tested on AWS using
- RHEL 9.4
- HANA scale-up combined with NWAS ASCS/ERS on 2 nodes

### 1. Regression fix: Stonith resource duplicate definition
- This is a regression introduced in my previous stonith change. :/

During the previous change of the stonith code a conditional was removed that did not apply to the GCP default, but is required for other platform default stonith resources that use host maps.

### 2. Fix: Check-mode
Fixed a conditional that failed in check-mode. The .stout var is always defined, but can be empty and that fails the version comparison in the last conditional.

### 3. Fix: HANA VIP colocation score calculation
- This issue comes up due to the combination of HANA resources with NW groups.

This fixes the HANA VIP colocation constraint conditional to always apply the base score and recalculate only when HANA-VIP resource groups are present in the definition.
The previous code enhanced with an `else` in the loop for a default score ended up appending the loop results as strings.

Due to the combination of HANA and NW resource definitions the group loop was processed put the if-conditional does not match, so the issue started showing up in the form of an empty score returned due a missing fallback value inside the for-loop. This causes the ha_cluster role to fail when trying to apply the constraint.
The change to variable assignment was done to overcome the issue with the loop appending results instead of overriding one value.
    
Assigning new values inside blocks/loops does not work (anymore) and a namespace object has to be used instead.
    
Reference: https://jinja.palletsprojects.com/en/latest/templates/#assignments

### 4. Mini enhancement: tag ha_cluster role import task
Added a tag to allow running without executing the ha_cluster role by skipping the tag `run_ha_cluster`.
This speeds up the construction of the input config file for debugging purpose and could also be used to run everything else without recreating the cluster config again.

Example usage that saves ~7 minutes of execution time just for producing the input parameters for review:
```
ansible-playbook ... -e 'sap_ha_pacemaker_cluster_create_config_varfile: true' -C --skip-tags run_ha_cluster
```

### 5. Enhancement: flexibility for combined setups

- fixing the hook dictionary lookup in combined setups
- adds a check to the host type definition to prevent invalid combinations

### DIFF
--- a/roles/sap_ha_pacemaker_cluster/tasks/RedHat/pre_steps_nwas_ascs_ers.yml
+++ b/roles/sap_ha_pacemaker_cluster/tasks/RedHat/pre_steps_nwas_ascs_ers.yml
@@ -20,4 +20,5 @@
       when:
         - sap_ha_pacemaker_cluster_nwas_cs_ers_simple_mount | bool
         - __sap_ha_pacemaker_cluster_sapstartsrv_check.stdout is defined
+        - __sap_ha_pacemaker_cluster_sapstartsrv_check.stdout | length > 0
         - "(__sap_ha_pacemaker_cluster_sapstartsrv_check.stdout) is version(__sap_ha_pacemaker_cluster_nwas_simple_mount_version, '<')"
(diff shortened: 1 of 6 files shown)


marcelmamula (approved):
LGTM, tested ASCS/ERS and HANA HA clusters on GCP.

ja9fuchs:
@marcelmamula Sorry, I added one more small fix/enhancement for the topic. :)

marcelmamula (approved):
LGTM

PR #966, opened 2025-02-13
sap_ha_pacemaker_cluster: fix internal-error

ja9fuchs:
For some reason the LSR collection variable is in the way of the ansible syntax-check of some tools, when the task that calls the ha_cluster LSR uses `import_role`.
The same syntax works without errors when the task uses `include_role` instead.

Fixes galaxy-importer error "internal-error: Unexpected error code 1".


marcelmamula (approved):
LGTM, Very refreshing to see all green checks

sean-freeman:
@ja9fuchs I'm 99.9999% sure you verified the difference in the logic between these two built-in modules, but for sanity purposes please find my own notes about the difference at the Playbook level.....

```
# Use include_role / include_tasks inside Ansible Task block, instead of import_roles.
# This ensures Ansible Roles, and the tasks within, will be parsed in sequence instead of parsing at Playbook initialisation.
```

PR #972, opened 2025-02-14
sap_ha_pacemaker_cluster: fix NWAS

ja9fuchs:
## Fix: SAPStartSrv monitor
The SAPStartSrv monitor operation should not run on an interval, only for a probe once at start. The added operation parameters will make sure that the monitor operation is disabled.
To make it very clear that the interval=0 is not a mistake, the `enabled` parameter is set to `false` as well.

Before:
```
monitor: rsc_SAPStartSrv_ELF_ERS33-monitor-interval-120s
  interval=120s timeout=20s
```

After:
```
monitor: rsc_SAPStartSrv_ELF_ERS33-monitor-interval-0s
  interval=0s timeout=20s enabled=0
```

This resulting config remains active on RHEL.
Currently the SUSE post-inst deletes all operations afterwards, so this change should not make any difference there.

_Reference_:
> If false, ignore this operation definition.
https://clusterlabs.org/projects/pacemaker/doc/2.1/Pacemaker_Explained/singlehtml/#operation-properties

## Fix: NWAS pre/post workflow

New file for NWAS (A)SCS/ERS pre-steps for changes to be run before the cluster is installed.
This contains...

- Instance systemd registration is now only run when the systemd service
does not exist yet. This prevents unnecessary re-registration.
    
- Some tasks were moved from post-installation to pre-installation, to
safely run them before the cluster is installed. This prevents some
conflicts that occur after the cluster already has control over the
instances.


ja9fuchs:
Pushed a fix and included a bunch of additional tags to be able to run only the preparations and ha_cluster.

marcelmamula (approved):
LGTM, Tested on SLES4SAP 15 SP6 on AWS

PR #973, opened 2025-02-14
collection and sap_hostagent: ansible-lint fixes

ja9fuchs:
- **collection**: added `.ansible/` to excludes, this is auto-created by pre-commit and contains the collection, on which the role excludes do not work

- **sap_hostagent**: removed `ansible_become` from defaults, it does not belong in the role but in the playbook, and it triggers the no-role-prefix rule

- **sap_hostagent**: changed some truthy values for consistency

- **sap_hostagent** is now free of linter errors and included in checks

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -2,6 +2,7 @@
 # Collection wide lint-file
 # DO NOT CHANGE
 exclude_paths:
+  - .ansible/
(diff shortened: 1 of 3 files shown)


berndfinger (approved):
LGTM.

Issue #132, opened 2022-05-16
sap_general_preconfigure: commented lines in /etc/hosts should not be checked for duplicates


ja9fuchs:
roles/sap_general_preconfigure/tasks/RedHat/generic/configure-etc-hosts.yml

The tasks that are checking for duplicates in /etc/hosts also parse comments and fail if matches are found in commented lines.

All /etc/hosts checks should skip comments.

ja9fuchs:
Reviewed and tested. I thought this must have been fixed, but it seems to still match comments:

```
[root@costop1 ~]# grep costop1.jfuchs /etc/hosts
10.99.30.14 costop1.jfuchs costop1
#10.99.30.14 costop1.jfuchs costop1
```

Result:
```
fatal: [costop1]: FAILED! => 
    assertion: __sap_general_preconfigure_register_fqdn_once_assert.stdout == '1'
    changed: false
    evaluated_to: false
    msg: 'FAIL: There is no line, or more than one line, containing ''costop1.jfuchs''
        in /etc/hosts!'
```

berndfinger:
> Reviewed and tested. I thought this must have been fixed

I think it was meant to be fixed as part of https://github.com/sap-linuxlab/community.sap_install/issues/697 but it appeared again after https://github.com/sap-linuxlab/community.sap_install/pull/737.

berndfinger:
Solved by https://github.com/sap-linuxlab/community.sap_install/pull/981.

Issue #881, opened 2024-10-24
sap_maintain_etc_hosts: Change space to tab delimiter


marcelmamula:
### Problem
`sap_install_media_detect` currently updates `/etc/hosts` file with spaces, creating usable but unsightly output.
Example of first line overwritten by role, while second line is kept unchanged.

```console
10.10.10.170 je1scs.aws.com je1scs
10.10.10.138    je1ers.aws.com      je1ers
```

### Solution
Change from space to Tab `\t` as already done in https://github.com/sap-linuxlab/community.sap_infrastructure/blob/0e67afc14738c8731192ef9f5040496c4a96e9b1/roles/sap_vm_provision/tasks/common/set_etc_hosts.yml#L24

@berndfinger  @rhmk

marcelmamula:
Resolved in https://github.com/sap-linuxlab/community.sap_install/pull/1212

PR #981, opened 2025-02-19
sap_general_preconfigure, sap_maintain_etc_hosts: Ignore comments

berndfinger:
... in column 1 of /etc/hosts lines.

Solves issue #132.


ja9fuchs (approved):
:+1: 

Tested on RHEL.

Verified:
- duplicate but commented entries are ignored by both roles with the PR code, while both fail independent from each other when using the current `dev` code
- the preconfigure role still fails correctly when there is a commented entry but no active one

marcelmamula (approved):
LGTM, it works but it still looks disheveled because of https://github.com/sap-linuxlab/community.sap_install/issues/881
```console
# 10.1.1.1 testnode.suse.com testnode
# 10.1.1.1 testnode.suse.com testnode
10.1.1.1 testnode.suse.com testnode
```

PR #930, opened 2025-01-14
sap_*_preconfigure/Suse: Rework of preconfigure roles for Suse, add missing notes.

marcelmamula:
## Description
Rework of all preconfigure roles to improve preconfiguration on Suse systems.

### Common changes across all 3 roles:
1. Split var files and update with correct packages for each product. Remove unnecessary obsolete packages and SAP patterns from base SLES.
2. Add support for `SLES_SAP_16` - WIP version.
3. Rework all `assert-` code to reflected updated tasks.
4. Add generic tasks for: saptune install and takeover, GRUB updates
5. Add handling for `packagekit.service` which blocks zypper on systems with enabled `SLE-Module-Desktop-Applications`
6. Fix default empty variables with correct `[] or {}`
7. Remove dependency on `python3-rpm` by replacing `ansible_facts.packages` with rpm query.
8. Move `ansible.builtin.package_facts` away from main to appropriate tasks, to enable point 7.

### `sap_general_preconfigure` specific
1. Add `2578899 - SUSE Linux Enterprise Server 15: Installation Note`

### `sap_hana_preconfigure` specific:
1. Remove `1275776 - Linux: Preparing SLES for SAP environments` - it is part of saptune generic tasks.
2. Rework `1944799 - SAP HANA Guidelines for SLES Operating System Installation` to remove hardcoded obsolete packages.
3. Remove `2578899 - SUSE Linux Enterprise Server 15: Installation Note` - its part of `sap_general_preconfigure`
4. Rework `2684254 - SAP HANA DB: Recommended OS settings for SLES 15 / SLES for SAP Applications 15`

### `sap_netweaver_preconfigure`
1. Add some variables that are in `sap_general_preconfigure` and `sap_hana_preconfigure`
2. Add handlers file, same as `sap_general_preconfigure` and `sap_hana_preconfigure`
3. Change input package variable to correct `sap_netweaver_preconfigure_packages`
4. Remove `1275776 - Linux: Preparing SLES for SAP environments` - it is part of saptune generic tasks.

## Test results
All 3 roles were tested on following systems and platforms:
- SLES 15 SP 5 on SUSE Rancher VM
- SLES 15 SP 6 on SUSE Rancher VM
- SLES_SAP 15 SP 5 on AWS and SUSE Rancher VM
- SLES_SAP 15 SP 6 on AWS and SUSE Rancher VM

PR #983, opened 2025-02-20
sap_*_preconfigure/Suse: Enhance saptune revert logic

marcelmamula:
### Description
PR https://github.com/sap-linuxlab/community.sap_install/pull/930 has introduced new way of handling `saptune` workflow, but it has shown new edge case where:
1. `sap_general_preconfigure` saptune applies sap note `2578899`
2. `sap_hana_preconfigure`: saptune applies solution HANA which already contains `2578899`
3. `sap_hana_preconfigure`: saptune verification of solution HANA fails, because: Application of solution skipped note `2578899`, which was older version, resulting in return code 1 in verification.

It is recommended to always revert saptune before applying solutions and this PR is handling that. 

### Test Results
This issue appeared on Cloud PAYG instances and this PR was tested on AWS and AZ with SLES4SAP 15 SP6, which had issue before.

### DIFF
--- a/roles/sap_general_preconfigure/tasks/sapnote/2578899/configuration.yml
+++ b/roles/sap_general_preconfigure/tasks/sapnote/2578899/configuration.yml
@@ -17,29 +17,44 @@
   when: __sap_general_preconfigure_grub_cmdline | length > 0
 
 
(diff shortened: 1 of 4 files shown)


berndfinger (approved):
LGTM.

PR #986, opened 2025-02-25
sap_ha_pacemaker_cluster: fix(check-mode)

ja9fuchs:
### Issue
The check-mode was inconsistent and some of it only surfaced on fresh nodes with no previous cluster config yet.

### Fix
Check-mode handling has been added to various tasks in different ways:
- `shell/command` tasks with no changes are executed, e.g. for discovering facts
- tasks that should never run in check-mode are skipped explicitly
- tasks that fail in check-mode ignore the failure - it is still useful to see what would be done

The last point especially concerns the `ha_cluster` role include. The entire include will be executed, but since the role is not check-mode safe, it will display the ignored errors. This way it is possible to see the workflow and review it as much as possible before the actual execution.

### Tested
- RHEL 9.4 on GCP
- RHEL 9.4 on AWS
- HANA scale-up
- NWAS ASCS/ERS
- fresh systems with no cluster installed before (no paths, no packages, etc.), but the SAP landscape is installed and configured as required
- re-run on the existing cluster, i.e. after the cluster configuration

@marcelmamula 
I was not sure about the desired check-mode behavior in the affected Suse tasks, that's why I did not make changes here:
- `tasks/Suse/post_steps_nwas_abap_ascs_ers.yml`
- `tasks/Suse/post_steps_nwas_java_scs_ers.yml`

However, they do need some improvements as well in my opinion, since there are a lot of commands executed in check-mode, which would either not work on fresh systems, or make actual changes in existing systems explicitly. Which is incorrect behavior in check-mode. Not sure if this is intended, so I kept my hands off for now. :)

### DIFF
--- a/roles/sap_ha_pacemaker_cluster/tasks/RedHat/pre_steps_hana.yml
+++ b/roles/sap_ha_pacemaker_cluster/tasks/RedHat/pre_steps_hana.yml
@@ -15,6 +15,7 @@
       ansible.builtin.command:
         cmd: dnf provides sap-hana-ha
       changed_when: false
+      check_mode: false
       register: __sap_ha_pacemaker_cluster_saphanasr_angi_check
       failed_when:
         - __sap_ha_pacemaker_cluster_saphanasr_angi_check.rc != 0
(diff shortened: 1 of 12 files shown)


marcelmamula (approved):
LGTM

Only problem I can see is this part failing on empty server
```yaml
- name: "SAP HA Pacemaker srHook - Get contents of global.ini"
  ansible.builtin.command:
    cmd: cat "{{ __sap_ha_pacemaker_cluster_hana_global_ini_path }}"
  register: __sap_ha_pacemaker_cluster_global_ini_contents
  changed_when: false
  check_mode: false
```

PR #989, opened 2025-03-03
sap_hana_install: Update fapolicyd conditionals

marcelmamula:
### Description
- Add missing conditionals for `fapolicyd` related tasks that can run only on system where `fapolicyd` is supported.

### Reasoning
- `fapolicyd` is not available or supported on SLES.
- Limiting `fapolicyd` tasks removes need to install `python3-rpm` on SLES in order to allow execution of `ansible.builtin.package_facts` which requires higher python version.


berndfinger (approved):
Tested and LGTM.

PR #993, opened 2025-03-04
sap_hana_preconfigure/SLES: Add package libltdl7 to vars

marcelmamula:
### Description
- Package `libltdl7` is required for SAP HANA installation on SLES, but it is only preinstalled on Cloud images. This change adds it to list of required packages to be installed for `sap_hana_preconfigure` role.

### Testing
- Tested on SLES4SAP 15 SP6 on AWS and local VM.

### Sources
[2998886 - SAP HANA DB installation fails with "hdbnsutil: error while loading shared libraries"](https://me.sap.com/notes/3029056)
[3029056 - Error while loading shared libraries: libltdl.so.7: cannot open shared object file: No such file or directory](https://me.sap.com/notes/2998886)


berndfinger (approved):
LGTM.

PR #994, opened 2025-03-06
sap_*_preconfigure/SLES: Enhance saptune handling and detection

marcelmamula:
### Description:
# New features
1. Introduced new variable with zypper patterns, which will be used with `command` instead of `package` module.
   - Reason: `package` module does not run zypper in default mode and it skips `Recommends` packages, resulting in discrepancines.
   - Example: Pattern `sap-hana` recommends packages that install `libltdl7`, but `package` will not install it when using rpm of pattern:  `patterns-sap-hana`
2. Introduced new variables `sap_hana_preconfigure_saptune_solution_force` and `sap_netweaver_preconfigure_saptune_solution_force` that allow forcing application of saptune solution regardless if it is already enabled and verified.
   - This follows recommendation that fresh re-enablement is always correct way to go, but variable is `false` by default.

# Fixes
3. Reverted commit https://github.com/sap-linuxlab/community.sap_install/commit/0dc982c2704b5a5c657c9432d9fb6a779fc73134
4. Fixed typo in saptune variable validation, resulting in skipping block
5. Re-enabled validation using `saptune solution enabled` after finding edge case where NETWEAVER can be successfully verified on host with HANA applied (overlap of parameters)
6. Updated RedHat specific variable comments with `(RedHat specific)` where appropriate: `sap_general_preconfigure_packagegroups`, `sap_general_preconfigure_envgroups`
7. Removed saptune solution enabled check from saptune_takeover as it is not necessary, since it is done in configuration step as well.

### Testing
All preconfigure roles were tested on SLES 15 SP6 and SLES4SAP 15 SP6

### DIFF
--- a/roles/sap_general_preconfigure/README.md
+++ b/roles/sap_general_preconfigure/README.md
@@ -258,7 +258,7 @@ If you set the RHEL minor release, then you must also use the `eus` or `e4s` rep
 - _Type:_ `str`
 - _Default:_ (set by platform/environment specific variables)
 
-The name of the software package group to install.<br>
+(RedHat specific) The name of the software package group to install.<br>
 The default for this parameter is set in the vars file which corresponds to the detected OS version.<br>
 
 Example:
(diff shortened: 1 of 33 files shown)


berndfinger (approved):
LGTM.

PR #996, opened 2025-03-12
sap*preconfigure: sysctl checks fail when config file has comments

rob0d:
Hi @berndfinger,

PR#807 introduced checks of kernel parameters to support indepotency.
However, under very rare circumstances when the sysctl conf files contain comments (#) this check fails as the awk command includes the line with the comments as a kernel parameter to check.
It generates something like:
```sysctl #This is comment net.ipv4.tcp_rmem```

This PR makes a minor change to make awk exclude all lines starting with # (i.e. ^#).

Kudos @James-Allen_nbs.

### DIFF
--- a/roles/sap_general_preconfigure/tasks/RedHat/generic/configure-kernel-parameters.yml
+++ b/roles/sap_general_preconfigure/tasks/RedHat/generic/configure-kernel-parameters.yml
@@ -17,7 +17,7 @@
 # in sap_general_preconfigure_etc_sysctl_sap_conf so that the "Reload kernel parameters from file ..." task
 # can correctly report its 'changed' state. See also https://github.com/sap-linuxlab/community.sap_install/issues/752 .
 - name: Construct the command for getting all current parameters of file '{{ sap_general_preconfigure_etc_sysctl_sap_conf }}'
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by berndfinger** [comment]
@rob0d Thanks for this. I had a quick look at `man sysctl.d` (RHEL 9) and it mentions that `empty lines and those whose first non-whitespace character is "#" or ";" are ignored`.

The following example should contain any allowed patterns I can think of (note that the second empty line contains three spaces and that the third empty line contained a `tab` character, and that the lines with a comment sign at column 8 were starting with a `tab` character):

```
net.ipv4.tcp_max_syn_backlog=8192
net.ipv4.tcp_timestamps=1

   
        
net.ipv4.tcp_slow_start_after_idle=0
net.core.wmem_max=4194304
#net.core.wmem_max=4194304
# net.core.wmem_max=4194304
 # net.core.wmem_max=4194304
        # net.core.wmem_max=4194304
;net.core.wmem_max=4194304
; net.core.wmem_max=4194304
 ; net.core.wmem_max=4194304
        ; net.core.wmem_max=4194304
net.core.rmem_max=6291456
```

It seems that the following awk statement will display only valid parameters:

`awk 'BEGIN{FS="="; printf ("sysctl ")}NF>1&&!/^\s*[#;]/{printf ("%s ", $1)}' FILE`

Note: NF is 0 for a line without a white space and 1 for a line with one or more white space characters.

rob0d:
@berndfinger  oh yes. We rushed a fix out as we needed to continue, but this is definitely better. Updated. Thanks Bernd.

Issue #730, opened 2024-04-26
sap_swpm: add (pseudo)idempotency 


rob0d:
It seems that SWPM in an unattended mode happily overwrites already installed systems. We've just experienced this when we executed the full workflow on a partially installed environment and it reinstalled one of the instances (and remove all the configuration that was done to that instance).

I would like to suggest to make `sap_swpm`  idempotent (dependant on settings). 

The safest and simplest way to do this would be probably to check the existence of an instance profile. This can be done either generating the file name from the parameters that are set or  by using code from [sap_profile_update](https://github.com/sap-linuxlab/community.sap_operations/blob/e247f47e7172c7f9e48c2bc87ea466b7a4ce1e47/roles/sap_profile_update/tasks/main.yml#L3C1-L7C8).

There would be three possible behaviours:
 - default: as is now and just execute SWPM and overwrite any existing instances
 - if instance already exists:
   - throw error
   - skip installation but continue
  
This can be  achieved by having something like  `sap_swpm_check_if_instance_exists ` with options:
- `false` to keep the current behaviour (default)
- `error` to throw error
- `ignore` to ignore it and continue

I guess similar outcome can be achieved by splitting playbooks into smaller chunks and manually re-executing the failed bits if something goes wrong with one of the workflow steps, but that feels error prone and user dependent.

marcelmamula:
@rob0d We will be implementing validations in sap_install roles some time in future to achieve this. @sean-freeman 
There are already some ways we handle it during our tests and developments, which we can ten transfer to actual roles.

I am for example using this idempotency to detect and skip:
- hana install, hsr replication
- ascs, ers, pas, aap installed

Issue is of course caused by SWPM, which will blast home folder of sidadm user, then fail on subsequent steps after it finds kernel files already present.

rob0d:
@marcelmamula that would be good. I can probably test some of it if it's available in the next month or so.
I think check in /usr/sap/sapservice and/or instance profile is probably the safest way of determining if something is (supposedly) installed without changing the state of anything.
And yes, the only way to prevent SWPM from updating the user profiles is not to start it at all.

marcelmamula:
`sap_hana_install` role already contains check for HANA existence and I think it can be partially reused to enhance `sap_swpm` role to not destroy system during re-executions.
- check for existence of SAP processes / executables
- if exists, then use them for SID + InstNR check
[...]

PR #1003, opened 2025-03-17
sap_swpm: Add idempotency

berndfinger:
Should solve #730. I tested new and existing installations with:
- `NW_ABAP_OneHost:S4HANA2023.FNDN.HDB.ABAP` and
- `NW_DI:S4HANA2023.FNDN.HDB.PD`.

### DIFF
--- a/roles/sap_swpm/tasks/instance_exists.yml
+++ b/roles/sap_swpm/tasks/instance_exists.yml
@@ -1,0 +1,121 @@
+# SPDX-License-Identifier: Apache-2.0
+---
+
+## Try to use saphostctrl to list installed SAP instances
+- name: SAP Instance Checks - Check if saphostctrl is installed
+  ansible.builtin.stat:
+    path: /usr/sap/hostctrl/exe/saphostctrl
+  check_mode: false
+  register: __sap_swpm_register_stat_saphostctrl
+  failed_when: false
+
+- name: SAP Instance Checks - Check with saphostctrl if SAP instances are installed
+  when: __sap_swpm_register_stat_saphostctrl.stat.exists
+  block:
+
+    - name: SAP Instance Checks - Get list of installed SAP instances
+      ansible.builtin.shell: set -o pipefail && /usr/sap/hostctrl/exe/saphostctrl -function ListInstances | cut -d":" -f2-
+      register: __sap_swpm_register_instancelist
+      changed_when: false
+
+    - name: SAP Instance Checks - Display instances
+      ansible.builtin.debug:
+        var: __sap_swpm_register_instancelist.stdout_lines
+        verbosity: 1
+
+    - name: SAP Instance Checks - Desired SAP instance is installed and running
+      ansible.builtin.set_fact:
+        __sap_swpm_fact_is_installed: true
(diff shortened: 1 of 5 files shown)


roles/sap_swpm/tasks/pre_install/swpm_prepare.yml
```diff
@@ -24,52 +24,85 @@
...
-  when: sap_swpm_run_sapinst
+- name: SAP SWPM - SAP instance is already installed
+  when: __sap_swpm_fact_is_installed
+  block:
+
+    - name: SAP SWPM - Report that the desired SAP instance is already installed
+      ansible.builtin.debug:
+        msg: "The instance with SID {{ sap_swpm_sid }} and one of the requested instance numbers is already installed."
+
+    - name: SAP SWPM - Stop here if the instance is already installed
```
marcelmamula, line 54: This should be removed and replaced with already existing when conditional (same as originally from hana install).
https://github.com/sap-linuxlab/community.sap_install/blob/9ee20c70df0bd34a2444b47b555729c4e91129bd/roles/sap_hana_install/tasks/main.yml#L62

`meta: end_play` is very inconsistent and loves to introduce uncontrolled behavior.

berndfinger: I tried to avoid `end_play` but so far without success because it appeared to me that this would require a complete overhaul of the role. Simply re-using the logic of `sap_hana_install` does not work here because there are many more tasks after the SAP instance existence check, most of which would have to be skipped.

We would have a lot of unnecessary outputs and could of course at the end display that the instance is already installed. But I'd prefer to exit quickly. Any further ideas welcome!

roles/sap_swpm/tasks/instance_exists.yml
```diff
@@ -0,0 +1,121 @@
...
+      when:
+        - __sap_swpm_register_stat_usr_sap_sid_assert.stat.exists
+        - __sap_swpm_register_files_in_usr_sap_sid_assert.matched | int != 0
+
+- name: SAP Instance Checks - SAP admin user check
+  when:
+    - not __sap_swpm_fact_is_installed
+  block:
+
+    - name: SAP Instance Checks - Get info about '{{ sap_swpm_sid | lower }}adm' user
```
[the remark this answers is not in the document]
berndfinger: I'd rather avoid reading the complete passwd database/file but only check for the one entry we are looking for - see commit e358c2dc.

berndfinger:
Note: If there is a communication error (e.g. visible in the output of `systemctl status SAP.slice`), causing the installed and running instances not being displayed in the output of `/usr/sap/hostctrl/exe/saphostctrl -function ListInstances`, then the user check is the only place where the role avoids an additional installation attempt.

Maybe we also need to detect if there is a malfunction when running `/usr/sap/hostctrl/exe/saphostctrl -function ListInstances`, and/or perform an additional check for the presence of an instance on the file system level.

berndfinger:
Closing this PR for now, see also https://github.com/sap-linuxlab/community.sap_install/discussions/1005 .

Issue #1011, opened 2025-04-03
sap_swpm: Remove unused dependency from Readme


marcelmamula:
### Details
Dependency mentioned in `sap_swpm` role is not used anywhere in code as selinux is done in preconfigure roles.

```yaml
<!-- BEGIN Dependencies -->
## Dependencies
- `fedora.linux_system_roles`
    - Roles:
        - `selinux`

Install required collections by `ansible-galaxy collection install -vv -r meta/collection-requirements.yml`.
<!-- END Dependencies -->
```
@berndfinger

PR #1012, opened 2025-04-03
sap_swpm: Remove selinux role dependency from README.md

berndfinger:
The selinux role is not called from sap_swpm, so the Dependencies section in README.md needs to be empty.


berndfinger:
Solves issue https://github.com/sap-linuxlab/community.sap_install/issues/1011.

marcelmamula (approved):
LGTM

PR #1014, opened 2025-04-16
Adding additional when to precent user settings to be changed

sjoerd-lubbers:


marcelmamula:
Please re-recreate PR correctly for us to review it.

Branch from `dev` and PR goes to `dev`

PR #1015, opened 2025-04-16
Add extra when for change user expire setting

sjoerd-lubbers:


marcelmamula:
@sjoerd-lubbers Please create branch again, but include `dev`. This current PR is `main` branch > `dev` which is not correct.

@berndfinger

PR #1016, opened 2025-04-16
sap_hana_install: Add opt-out for setting sidadm to noexpire 

sjoerd-lubbers:

### DIFF
--- a/roles/sap_hana_install/defaults/main.yml
+++ b/roles/sap_hana_install/defaults/main.yml
@@ -113,10 +113,6 @@ sap_hana_install_check_sidadm_user: true
 # an existing HANA system.
 sap_hana_install_new_system: true
 
-# Set to `true` to ensure the SAP system user '{{ sid | lower }}adm' is non-expiring.
-# Set to `false` to skip this step — typically used when the user is domain-managed.
-sap_hana_install_set_sidadm_noexpire: true
-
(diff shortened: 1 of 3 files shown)


marcelmamula:
@sjoerd-lubbers It looks much better now, with only 1 commit, no mergeback and no direct `main` merge.

You are trying to add new variable `sap_hana_install_set_sidadm_noexpire` which makes sense, but we are not using hidden variables in our roles, only when they are runtime set `__variables` (not role inputs).

Adding this variable would require to document it in `defaults/main.yml` and `README.md`. @berndfinger

roles/sap_hana_install/defaults/main.yml
```diff
@@ -113,6 +113,10 @@ sap_hana_install_check_sidadm_user: true
 # an existing HANA system.
 sap_hana_install_new_system: true
 
+# Set to `true` to ensure the SAP system user '{{ sid | lower }}adm' is non-expiring.
```
berndfinger, line 116: Please move this section to right above the section for `sap_hana_install_create_initial_tenant` so that it is in the same order as in the tasks file.

berndfinger:
@sjoerd-lubbers Many thanks for your suggestion, for providing the code, and for adding the missing pieces. I am now running a final test with your code.

Issue #347, opened 2023-03-22
sap_swpm: Using different TEMP dir fails playbook


slubb:
I'm installing SAP using a different `TEMP` directory:

```
- name: Play install SAP Web Dispatcher
  hosts: hosts
  gather_facts: True
  become: true
  environment:
    TEMP: /usr/sap/tmp
```
Installation works as expected however there are 2 issues:

1. The summary pre start: It now shows the wrong path.
```
TASK [community.sap_install.sap_swpm : SAP SWPM - Installation Process] ******************************************************
ok: [xxhost] => {
    "msg": [
        "   Ready to run SWPM for SAP Software Installation  ",
        " ",
        "   SAP Product        -       NW_Webdispatcher:NW750.IND.PD  ",
        "   SID                -       WAD ",
        "   Host               -       xxhost",
        " ",
        "   The installation can take up to 3 hours. Run the following command as root",
        "   on xxhost to display the installation logs:",
        "   # tail -f $(cat /tmp/sapinst_instdir/.lastInstallationLocation)/sapinst.log"
    ]
}

``` 
2. This step now fails because it doesn't look at the correct tmp dir.
```
TASK [community.sap_install.sap_swpm : SAP SWPM - Find last installation location] *******************************************
fatal: [xxhost]: FAILED! => 
{
    "changed": false,
    "cmd": [
        "cat",
        "/tmp/sapinst_instdir/.lastInstallationLocation"
    ],
    "delta": "0:00:00.008100",
    "end": "2023-03-17 12:14:35.335597",
    "msg": "non-zero return code",
    "rc": 1,
    "start": "2023-03-17 12:14:35.327497",
    "stderr": "cat: /tmp/sapinst_instdir/.lastInstallationLocation: No such file or directory",
    "stderr_lines": [
        "cat: /tmp/sapinst_instdir/.lastInstallationLocation: No such file or directory"
    ],
    "stdout": "",
    "stdout_lines": []
}
```

sean-freeman:
Hi @slubb , the error occurs because these Ansible Tasks statically use `/tmp`.

Can you provide justification why you want to change the Temporary Directory for the installation, and why `/usr/sap/tmp` (which should really be reserved `<sid>adm` processes).

I am unsure whether adding a variable to switch the Temporary Directory is a good idea. I would need convincing.

sean-freeman:
[...]

PR #1019, opened 2025-04-18
sap_swpm: Fix TEMP handling

berndfinger:
Also refactor existing variable sap_swpm_tmpdir and other registered variables.

Solves issue #347.


marcelmamula:
@berndfinger Please ensure this is tested with different product installations. SWPM and SUM tend to fail if you nest its location too much as they have character limit for paths.

Using Ansible temporary folder adds fairly long folder name from `/tmp/`

sjoerd-lubbers:
@berndfinger When is this planned to be merged to dev, I want to double check our rollout and I now have stuff in dev and this branch. Thanks.

Issue #1023, opened 2025-04-24
sap_swpm: For triggering the re-use of inifile.params, this file needs to be present on both the managed node and the control node


berndfinger:
### Ansible Role

sap_swpm

### OS Family

N/A

### Ansible Controller - Python version

Python 3.13.3

### Ansible-core version

ansible [core 2.18.3]

### Bug Description

In [this task](https://github.com/sap-linuxlab/community.sap_install/blob/8a89687bc58f865f6d77a38d960323345ff2a60f/roles/sap_swpm/tasks/pre_install/generate_inifile.yml#L13), the role `sap_swpm` is checking if there is a file `'{{ sap_swpm_inifile_directory }}/inifile.params'` on the **managed node**. If it is, it will display an INFO message that this file will be used for running sapinst.

However, in the [task which follows](https://github.com/sap-linuxlab/community.sap_install/blob/8a89687bc58f865f6d77a38d960323345ff2a60f/roles/sap_swpm/tasks/pre_install/generate_inifile.yml#L28), the role attempts to copy this file from the **control node** (using the same path and file name) to the desired location on the managed node.

The role should rather only require the file to be present on the managed node - so the latter task should get a `remote_src: true`, similar to [this task](https://github.com/sap-linuxlab/community.sap_install/blob/8a89687bc58f865f6d77a38d960323345ff2a60f/roles/sap_swpm/tasks/pre_install/swpm_prepare.yml#L14) in the same role.

### Bug reproduction

1. Place a file named `inifile.params` in a directory specified by the role variable `sap_swpm_inifile_directory`.
2. Run the role sap_swpm.
3. Observe that the role will fail in the task https://github.com/sap-linuxlab/community.sap_install/blob/8a89687bc58f865f6d77a38d960323345ff2a60f/roles/sap_swpm/tasks/pre_install/generate_inifile.yml#L28 .

### Community participation

Happy to implement this bug fix

berndfinger:
Fixed by https://github.com/sap-linuxlab/community.sap_install/pull/1024.

PR #1024, opened 2025-04-25
sap_swpm: Require inifile.params only on the managed node when re-using an existing inifile.params.

berndfinger:
Solves issue #1023.

Also display the full path name of the inifile in a separate task, similar to the role sap_hana_install.


marcelmamula (approved):
LGTM

ja9fuchs (approved):
looks good :+1:

PR #1028, opened 2025-05-06
sap_ha_pacemaker_cluster/SUSE: Remove python3-rpm dependency in pre_steps_hana

marcelmamula:
## Description
Remove dependency on `python3-rpm` that is needed for `ansible.builtin.package_facts`.

This was already done for other preconfigure roles as well as `ha_cluster` in LSR.

### TODO
@berndfinger It would be good to remove it also from  `sap_swpm` firewall tasks, but that can be done separately.


ja9fuchs (approved):
looks good to me

PR #1032, opened 2025-05-16
sap_*_preconfigure/SUSE: Add retry attempts to zypper pattern installation 

marcelmamula:
## Description

Add 2 more attempts to `zypper -t pattern` installation to resolve issues with some pattern installations, in particular `sap-nw` for `sap_netweaver_preconfigure` role.


berndfinger (approved):
LGTM

PR #1033, opened 2025-05-16
sap_anydb_install_oracle: Add handling of OS specific vars

marcelmamula:
## Description
- Added loading of OS specific variables based on ansible distribution and versions, similar to existing one used in other roles.
- Corrected packages for SUSE as existing list was not valid.

### DIFF
--- a/roles/sap_anydb_install_oracle/tasks/main.yml
+++ b/roles/sap_anydb_install_oracle/tasks/main.yml
@@ -1,6 +1,35 @@
 # SPDX-License-Identifier: Apache-2.0
 ---
 
+# Example of files loading order:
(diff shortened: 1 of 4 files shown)


sean-freeman:
@marcelmamula Would suggest moving the `# See further` commented code into the actual var files. Otherwise completely approve, big oversight from the original coder (👋 ) verifying only with OL (a RHELative).

marcelmamula:
> @marcelmamula Would suggest moving the `# See further` commented code into the actual var files. Otherwise completely approve, big oversight from the original coder (👋 ) verifying only with OL (a RHELative).

Thank you @sean-freeman , changes were pushed and comments were moved to var file.

sean-freeman (approved):
lgtm 👍

Issue #1034, opened 2025-05-22
sap_general_preconfigure/RHEL:  Modify the kernel command line for SELinux on all RHEL > 8


berndfinger:
### Ansible Role

sap_general_preconfigure

### OS Family

RHEL

### Ansible Controller - Python version

Python 3.13.3

### Ansible-core version

ansible [core 2.18.3]

### Bug Description

The kernel command line needs to be checked and modified for SELinux also on RHEL 10 systems. See also https://github.com/linux-system-roles/selinux/issues/274#issuecomment-2901382530.

### Bug reproduction

Run the role `sap_general_preconfigure` against a RHEL 10 system and try to disable or enable SELinux using the corresponding role variables.

### Community participation

Happy to implement this bug fix

berndfinger:
Solved in https://github.com/sap-linuxlab/community.sap_install/pull/1036.

PR #1036, opened 2025-06-02
sap_general_preconfigure: Modify the kernel command line for SELinux also for RHEL 10

berndfinger:
This PR modifies the kernel command line for SELinux also for RHEL 10.

Note: Replacing all current SELinux related code by the LSR `selinux` role was not yet possible because I found no way to postpone the `selinux` role failure in all cases. In other words, there were cases where the transition from one SELinux configuration to another caused the `selinux` role to fail immediately, leading to an abort of the role `sap_general_preconfigure`.


ja9fuchs (approved):
looks correct :+1:

berndfinger:
Solves issue #1034.

PR #1061, opened 2025-06-26
sap_ha_pacemaker_cluster/SUSE: SAP HANA scaleup post steps updated

marcelmamula:
## Changes
- SAP HANA post steps restructured to ensure that each step is executed when SAP HANA is idle and it does not fail.
- Steps were split for Angi and non-Angi

## Tested
Cluster was tested on SLES 15 SP6 and SLES 16 RC1.
First command `cs_wait_for_idle -s 5` takes around 5 minutes while SAP HANA Is starting, then subsequent commands are quick.


ja9fuchs (approved):
:+1:

Issue #1060, opened 2025-06-25
sap_swpm: Add input validation based on sap_swpm_inifile_sections_list


marcelmamula:
## Problem

Current implementation does not validate input variables and proceeds to use them regardless.
Example:
1. `sap_swpm_wd_instance_nr` is defined as `None` and it is used in jinja2 template when `if 'nw_config_webdisp_generic' in sap_swpm_inifile_sections_list`
2. `sap_swpm_db_instance_nr` is used in firewall steps as well as jinja2 template, but it is never validated that it is string and not empty.

## Solution
Add pre_install task file to validate all variables in relation to selected `sap_swpm_inifile_sections_list` item.

@berndfinger

marcelmamula:
This was solved in our extensive validation PR https://github.com/sap-linuxlab/community.sap_install/pull/1108

PR #1067, opened 2025-07-09
sap swpm: Perform input validation for vars in inifile sections

berndfinger:
Solves issue #1060.
Also add empty strings for all password vars in `defaults/main.yml`.
Also improve the existing assert loops in `roles/sap_swpm/tasks/pre_install/generate_inifile.yml` and `roles/sap_swpm/tasks/pre_install/assert_hostname_length.yml`.


marcelmamula (changes requested):

roles/sap_swpm/tasks/pre_install/assert_inifile_vars_loop_block.yml
```diff
@@ -0,0 +1,31 @@
...
+---
+
+# loop block for pre_install/assert_inifile_sections_vars.yml
+
+- name: SAP SWPM Pre Install - Assert variable '{{ __sap_swpm_vars_to_be_asserted_line_item.name }} for 'length_greater_zero''
+  ansible.builtin.assert:
+    that:
+      - lookup('ansible.builtin.vars', __sap_swpm_vars_to_be_asserted_line_item.name, default='') is defined
+      - lookup('ansible.builtin.vars', __sap_swpm_vars_to_be_asserted_line_item.name, default='') is string
+      - lookup('ansible.builtin.vars', __sap_swpm_vars_to_be_asserted_line_item.name, default='') | length > 0
```
marcelmamula, line 11: Do we need to add `| trim` ? Current check will allow `" "` strings.
```yaml
    that:
      - lookup('ansible.builtin.vars', __sap_swpm_vars_to_be_asserted_line_item.name, default='') is defined
      - lookup('ansible.builtin.vars', __sap_swpm_vars_to_be_asserted_line_item.name, default='') is string
      - lookup('ansible.builtin.vars', __sap_swpm_vars_to_be_asserted_line_item.name, default='') | trim | length > 0
```

There is risk of course, if there is some variable that can contain spaces.

roles/sap_swpm/tasks/pre_install/assert_inifile_vars.yml
```diff
@@ -0,0 +1,58 @@
...
+      }}
+  loop: "{{ __sap_swpm_inifile_vars_to_be_asserted | subelements('sections') }}"
+  loop_control:
+    loop_var: __sap_swpm_vars_to_be_asserted_initial_line_item
+    label: |
+      {{ __sap_swpm_vars_to_be_asserted_initial_line_item.0.name,
+         __sap_swpm_vars_to_be_asserted_initial_line_item.0.check,
+         __sap_swpm_vars_to_be_asserted_initial_line_item.1 }}
+
+- name: SAP SWPM Pre Install - Create a new list of vars only used in defined sections
```
marcelmamula, line 47: This extra variable can be removed by adding when conditional to task above and defaulting in vars/main to `[]`

`when: __sap_swpm_vars_to_be_asserted_initial_line_item.1 in sap_swpm_inifile_sections_list | d([])`

marcelmamula (changes requested):

marcelmamula (changes requested):
Items like `sap_swpm_sapcar_path` and `sap_swpm_swpm_path` also have to be defaulted to `''` because they are used in many places without validating.

Example: `"{{ sap_swpm_sapcar_path | d(sap_swpm_software_path) }}"` will never work because it is defined as `None`.

roles/sap_swpm/tasks/pre_install/assert_inifile_vars_loop_block.yml
berndfinger: Yes, I think we can safely use `| trim` to also catch vars defined as one or more spaces.

roles/sap_swpm/tasks/pre_install/assert_hostname_length.yml
```diff
@@ -21,6 +23,6 @@
   loop_control:
     loop_var: __sap_swpm_hostnames_line_item
   when:
-    - lookup('ansible.builtin.vars', __sap_swpm_hostnames_line_item) is defined
-    - lookup('ansible.builtin.vars', __sap_swpm_hostnames_line_item) is not none
-    - lookup('ansible.builtin.vars', __sap_swpm_hostnames_line_item)
+    - lookup('ansible.builtin.vars', __sap_swpm_hostnames_line_item, default='') is defined
+    - lookup('ansible.builtin.vars', __sap_swpm_hostnames_line_item, default='') is not none
+    - lookup('ansible.builtin.vars', __sap_swpm_hostnames_line_item, default='') | string | length > 0
```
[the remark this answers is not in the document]
berndfinger: @marcelmamula I don't understand, can you please provide more details on your suggestion?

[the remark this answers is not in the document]
marcelmamula: @berndfinger I mentioned that some roles are using dedicated host type variables that are used for this specific purpose, to determine what vars are to be asserted and validated.

Is there a way to do same in SWPM role so you dont have to go and validate every variable, regardless if it is needed?
- ASCS install? no need to check for ERS, PAS, etc, but at same time dont skip them if they are undefined, because it does not mean they are not required.

```yaml
sap_storage_setup_host_type: "hana_primary"
sap_ha_pacemaker_cluster_host_type | select('search', 'hana_scaleup') | length > 0
```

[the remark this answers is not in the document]
berndfinger: We are validating 6 different hostname vars early, in the first task of `pre_install.yml`, and we are validating 4 of these hostnames again later, in the task `SAP SWPM Pre Install - Perform input validation for vars in sap_swpm_inifile_sections_list` - dependent on which sections of the inifile have been defined.

The know-how about which hostname vars are needed for certain installation scenarios is already present in file `templates/inifile_params.j2`.

I can think of the following:

We extend the list of dicts `__sap_swpm_inifile_vars_to_be_asserted` as defined in `vars/main.yml` to also contain the two missing hostname vars, `sap_swpm_java_scs_instance_hostname` and `sap_swpm_virtual_hostname`.

We then eiter:
1. Do not validate the hostname vars early but only later, in the task `SAP SWPM Pre Install - Perform input validation for vars in sap_swpm_inifile_sections_list`
2. We continue validating the hostnames early. I will then extract the 6 hostname vars from the list `__sap_swpm_inifile_vars_to_be_asserted` (maybe move them into a separate list) so we can validate them early.

roles/sap_swpm/tasks/pre_install/assert_inifile_vars_number_loop_block.yml
marcelmamula, line 1: Another reworked and improved loop that works for new ansible.
- it now correctly checks for double digit number, allowing 2 digit integers (fails on 01 integers) as well as String.

```yaml
- name: SAP SWPM Pre Install - Assert variable {{ __sap_swpm_var_name }} for 'two_digit_number'
  ansible.builtin.assert:
    that:
      - vars[__sap_swpm_var_name] is defined
      - (lookup('ansible.builtin.vars', __sap_swpm_var_name) | trim) is ansible.builtin.regex('^\\d{2}$')
    success_msg: |
      PASS: The variable '{{ __sap_swpm_var_name }}' used in section '{{ __sap_swpm_var_section }}' is valid.
    fail_msg: |
      {% if vars[__sap_swpm_var_name] is not defined %}
        FAIL: The variable '{{ __sap_swpm_var_name }}' used in section '{{ __sap_swpm_var_section }}' is undefined.
      {% elif lookup('ansible.builtin.vars', __sap_swpm_var_name) | string | trim | length == 0 %}
        FAIL: The variable '{{ __sap_swpm_var_name }}' used in section '{{ __sap_swpm_var_section }}' is empty.
      {% else %}
        FAIL: The variable '{{ __sap_swpm_var_name }}' used in section '{{ __sap_swpm_var_section }}' is not a double digit number.
        Value can be defined as String or Integer in range from 00 to 99.
      {% endif %}
  vars:
    __sap_swpm_var_name: "{{ __sap_swpm_vars_to_be_asserted_for_number_line_item.name }}"
    __sap_swpm_var_section: "{{ __sap_swpm_vars_to_be_asserted_for_number_line_item.section }}"
  when:
    - __sap_swpm_var_section in sap_swpm_inifile_sections_list
```

roles/sap_swpm/tasks/pre_install/generate_inifile.yml
```diff
@@ -56,9 +56,11 @@
 # sap_swpm_product_catalog_id, sap_swpm_software_path, sap_swpm_sid, sap_swpm_fqdn
 - name: SAP SWPM Pre Install - Assert that certain variables are set
   ansible.builtin.assert:
-    that: "{{ __sap_swpm_vars_line_item }} is defined and {{ __sap_swpm_vars_line_item }}"
-    fail_msg: "FAIL: '{{ __sap_swpm_vars_line_item }}' is either undefined or empty!"
-    success_msg: "PASS: '{{ __sap_swpm_vars_line_item }}' is set."
+    that:
+      - lookup('ansible.builtin.vars', __sap_swpm_vars_line_item, default='') is defined
```
marcelmamula, line 60: This line is not 2.19 ready but we can fix it in separate PR with rest of fixes for sap_swpm.

marcelmamula (changes requested):
Review after last commits

roles/sap_swpm/tasks/pre_install/assert_inifile_vars_number_loop_block.yml
berndfinger: @marcelmamula With your suggestion in chat for using `lookup('ansible.builtin.vars', __sap_swpm_var_name, default='SAP_SWPM_UNDEFINED_VARIABLE_DETECTED') != 'SAP_SWPM_UNDEFINED_VARIABLE_DETECTED'`  instead of `vars[__sap_swpm_var_name] is defined` , I tested successfully on Ansible up to and including 2.19.0b4." See also my next commit, which has this code.

berndfinger:
@marcelmamula Thanks for your input. In the meantime, I believe I found a better solution by using the template for the validation instead of defining a list of dicts, created from the template. After incorporating your suggestion of  looping with Jinja instead of Ansible loops, it will also provide a very nice and compact output in case of validation failures. I will soon create a PR. In the meantime, https://github.com/berndfinger/community.sap_install/tree/issue-1060-improved-swpm-input-validation can be used as a preview.

PR #1076, opened 2025-07-16
sap_*_preconfigure: Define needs-restarting also for RHEL 10, assert mode

berndfinger:
Fixes issue #1075.


ja9fuchs (approved):
lgtm :+1:

Issue #971, opened 2025-02-14
sap_ha_install_hana_hsr: possibility to define the backup path


domses:
Hello,

in role sap_ha_install_hana_hsr, task file run_backup.yml and step "SAP HSR - Verify backup files" you have hard coded the default Backup Path of /usr/sap/<SID>/HDBXX/backup -  this should be more flexible, in many cases there are customer defined Backup Paths on NFS Shares and so on.

So from my POV, it would be an improvement to either use a var to define the backup basepath, or retrieve the configured backup basepaths from the parameters in the system: basepath_databackup,  basepath_logbackup and basepath_catalogbackup.

marcelmamula:
Thank you @domses , I have created PR with this change in mind and few other improvements.
https://github.com/sap-linuxlab/community.sap_install/pull/1090

PR #1090, opened 2025-07-30
sap_ha_install_hana_hsr: Add backup location detection and user variable

marcelmamula:
## Changes
This PR was built with idea from https://github.com/sap-linuxlab/community.sap_install/issues/971

- Add user variable `sap_ha_install_hana_hsr_backup_path` that can be used to define custom backup location.
- Add detection of HANA parameter instead of hardcoded path, but leave fall back default path as variable.
- Add handling of SYSTEMDB and TENANTDB folders when `sap_ha_install_hana_hsr_backup_path` is provided.

## Tests
This PR was tested on SLES 16 HANA HA nodes, with all combinations of variables.

## NOTE
This role will require some refactoring in future in order to catch it up with other roles in repository.

### DIFF
--- a/roles/sap_ha_install_hana_hsr/defaults/main.yml
+++ b/roles/sap_ha_install_hana_hsr/defaults/main.yml
@@ -1,13 +1,43 @@
 # SPDX-License-Identifier: Apache-2.0
 ---
-sap_ha_install_hana_hsr_sid: "{{ sap_hana_sid }}"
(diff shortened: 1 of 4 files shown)


ja9fuchs (approved):
:+1: 

Tested successfully in check_mode and real run using both, the defaults and with `sap_ha_install_hana_hsr_backup_path` defined.

Issue #1093, opened 2025-07-31
feat: sap_hana_preconfigure allow disabling modification of command line boot args


GutiPerez:
### Ansible Role

sap_hana_preconfigure

### OS Family

All

### Feature/Enhancement request

I believe after 1.7.0 release changes, specifically, the boot command line arguments checks added on #1083, there should a way to disable the modification of the arguments. I was expecting `sap_hana_preconfigure_modify_grub_cmdline_linux` variable to do so, but it is not.

Actually, I think it is contraprodudent that all arguments are always set if not present. For instance, `processor.max_cstate` and `intel_idle.max_cstat` should not be set at all if we use sap-hana tuned profile (which the role does by default with `sap_hana_preconfigure_use_tuned` parameter). In our case, we have HPE hardware, and a mandatory HPE package for RHEL handles tsx and transparent_hugepages arguments.

Finally, a minor bug/unexpected situation. For [the following task](https://github.com/sap-linuxlab/community.sap_install/blob/1.7.0/roles/sap_hana_preconfigure/tasks/RedHat/include_vars.yml#L8) , included in v1.7.0, given that we need to run the playbook with `become: true` for the role to work, now I can not run the playbook from an user that can not become root, as this task is delegated to localhost. This could be easily fixed with a `become: false` on that specific task

### Community participation

Unfortunately I am not in a position to help with the feature

berndfinger:
@GutiPerez Release 1.7.1 now has [the fix](https://github.com/sap-linuxlab/community.sap_install/pull/1098) for #1093, as well as [the fix](https://github.com/sap-linuxlab/community.sap_install/pull/1097) for #1095 (which is the second issue you reported in this issue). I omitted the ability to set your own list of boot command line arguments for this release because we first need to discuss how to validate such a list, and I wanted to quickly fix this issue.

With this, there are now the following options, just like before release 1.7.0:
- **Activate the tuned profile** sap-hana (or another one) and do **not modify** the boot command line as stated in the applicable SAP notes for RHEL.
- **Activate the tuned profile** sap-hana (or another one) and **also modify** the boot command line as stated in the applicable SAP notes for RHEL.
- **Do not activate the tuned profile** sap-hana (or another one) but **modify** the boot command line as stated in the applicable SAP notes for RHEL.

GutiPerez:
Thanks a lot @berndfinger

Agreed, v1.7.1 solves my initial issues.

[...]

PR #1098, opened 2025-08-04
sap_hana_preconfigure/RHEL: Be more flexible with boot command line args

berndfinger:
Solves issue #1093.

### DIFF
--- a/roles/sap_hana_preconfigure/defaults/main.yml
+++ b/roles/sap_hana_preconfigure/defaults/main.yml
@@ -167,14 +167,18 @@ sap_hana_preconfigure_run_grub2_mkconfig: true
 # By default, the role will run `grub2-mkconfig` to update the Grub configuration if necessary.
 # Set this parameter to `false` if this is not desired.
 
-# sap_hana_preconfigure_db_group_name: (not defined by default)
-# Use this parameter to specify the name of the RHEL group which is used for the database processes.
-# It will be used to configure process limits as per step "Configuring Process Resource Limits" of SAP note 2772999.
-# Example: See README.md
+sap_hana_preconfigure_boot_command_line_args: "{{ __sap_hana_preconfigure_boot_command_line_args }}"
+# Use this parameter to set your own combination of boot command line arguments. If unchanged,
+# the arguments as defined in the vars files are used.
 
 # sap_hana_preconfigure_thp: (not defined by default)
 # Override the default setting for THP, which is determined automatically by the role, depending on the RHEL version.
 # Can be one of 'always', 'madvise', or 'never'.
+
+# sap_hana_preconfigure_db_group_name: (not defined by default)
+# Use this parameter to specify the name of the RHEL group which is used for the database processes.
+# It will be used to configure process limits as per step "Configuring Process Resource Limits" of SAP note 2772999.
+# Example: See README.md
 
 # (SUSE specific) Version of saptune to install.
 # It is recommended to install latest version by keeping this variable empty.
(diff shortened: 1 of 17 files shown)


roles/sap_hana_preconfigure/defaults/main.yml
```diff
@@ -167,15 +167,19 @@ sap_hana_preconfigure_run_grub2_mkconfig: true
...
+sap_hana_preconfigure_boot_command_line_args: "{{ __sap_hana_preconfigure_boot_command_line_args }}"
+# Use this parameter to set your own combination of boot command line arguments. If unchanged,
+# the arguments as defined in the vars files are used.
 
 # sap_hana_preconfigure_thp: (not defined by default)
 # Override the default setting for THP, which is determined automatically by the role, depending on the RHEL version.
 # Can be one of 'always', 'madvise', or 'never'.
 
+# sap_hana_preconfigure_db_group_name: (not defined by default)
+# Use this parameter to specify the name of the RHEL group which is used for the database processes.
```
marcelmamula, line 179: Please add `{Red Hat specific)` as we established it on project level. There are still other RH vars that are lacking this comment.

roles/sap_hana_preconfigure/tasks/RedHat/generic/configure-c-states-for-lower-latency.yml
```diff
@@ -6,9 +6,9 @@
   ansible.builtin.set_fact:
     __sap_hana_preconfigure_boot_command_line_args_final: >-
       {{ __sap_hana_preconfigure_boot_command_line_args_final + [__sap_hana_preconfigure_kernel_arg_line_item] }}
-  loop: "{{ __sap_hana_preconfigure_boot_command_line_args['c-states']['args'] }}"
+  loop: "{{ sap_hana_preconfigure_boot_command_line_args['c-states']['args'] }}"
```
marcelmamula, line 9: You have opted to using `sap_hana_preconfigure_boot_command_line_args` without checking its validity.
It assumes that user provided is correct, without telling them what format they need to provide.

It should be always:
1. Check if user var is provided and valid, only then reassign it as `__var` for further processing
2. Use default `__var` if user var was not provided and or invalid.

roles/sap_hana_preconfigure/vars/RedHat_7.yml
```diff
@@ -249,4 +249,10 @@ __sap_hana_preconfigure_packages_and_services:
   numad:     { pkg: 'numad',           svc: 'numad',     systemd_enabled: 'no', systemd_state: 'stopped', svc_status: 'disabled', svc_state: 'inactive' }
   kdump:     { pkg: 'kexec-tools',     svc: 'kdump',     systemd_enabled: 'no', systemd_state: 'stopped', svc_status: 'disabled', svc_state: 'inactive' }
   firewalld: { pkg: 'firewalld',       svc: 'firewalld', systemd_enabled: 'no', systemd_state: 'stopped', svc_status: 'disabled', svc_state: 'inactive' }
+
+__sap_hana_preconfigure_boot_command_line_args:
```
marcelmamula, line 253: This is set only for RHEL 7, which means all other calls from `default/main.yml` for `sap_hana_preconfigure_boot_command_line_args` will fail because you did not define it as `{}`.

marcelmamula (changes requested):

roles/sap_hana_preconfigure/defaults/main.yml
berndfinger: > Please add `(Red Hat specific)` as we established it on project level. There are still other RH vars that are lacking this comment.
This variable is not used in this role, so I will remove it.

Issue #1101, opened 2025-08-05
Validate compatibility with ansible-core 2.19


marcelmamula:
## Details
Source: https://github.com/ansible-community/ansible-build-data/issues/538

ansible-core has implemented an extensive rewrite of templating and error handling as part of the https://github.com/ansible/ansible/pull/84621 coming in core-2.19. 
You can find information about this change in that PR description, as well as in the [core 2.19 porting guide](https://docs.ansible.com/ansible-core/devel/porting_guides/porting_guide_core_2.19.html).

Please read [background and testing requests,](https://forum.ansible.com/t/data-tagging-preview-and-testing/40759) which details where you can file potential bugs.

You can find advice from fellow maintainers at [Making a collection compatible with data tagging](https://forum.ansible.com/t/making-a-collection-compatible-with-data-tagging/41717), and we welcome your additions to that forum topic.

## Required changes
All roles need to be tested using new `ansible-core 2.19` and fixed before we can establish support matrix for it.
```bash
# Setup python venv and install pre-release ansible-core
pip install --upgrade --pre 'ansible-core~=2.19.0b1'
```

PR #1102, opened 2025-08-07
collection: Adjust roles for ansible-core 2.19 part 1

marcelmamula:
## Reason
Details about upcoming ansible-core 2.19 are listed in https://github.com/sap-linuxlab/community.sap_install/issues/1101

## Changes
- Add conditionals where truthy was attempted against non-boolean variables
- Add checks for `none` or `type_debug` when required.
- Add condition to all `regex_search` in when checks, because they return `NoneType` which is not truthy (This is not an issue inside of `set_fact`).

**NOTE 1: Dictionary assignments have to be carefully tested because `2.19` now marks them as `string`, invalidating lot of code.**

**NOTE 2: Changes to role `sap_swpm` are prepared and waiting for https://github.com/sap-linuxlab/community.sap_install/pull/1067 to be completed first.**

## Tests
Latest commit was tested on `ansible-core 2.16 2.18 2.19` targeting SLES 15 SP6 on AWS and SLES 16 on premise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ __pycache__/
 
 # vi swap file
 *.swp
+
+# Ignore ansible workspace
+.ansible
--- a/roles/sap_general_preconfigure/handlers/main.yml
+++ b/roles/sap_general_preconfigure/handlers/main.yml
@@ -31,8 +31,9 @@
 
 - name: "Debug grub-mkconfig BIOS mode"
   ansible.builtin.debug:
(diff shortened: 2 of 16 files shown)


ja9fuchs (approved):
:+1: 
looks good

berndfinger:
A good alternative to adding a condition to all `regex_search` in `when` checks appears to be using the `is search` or `is regex` test instead of `| regex_search`.

Issue #1104, opened 2025-08-13
feat: sap_swpm: Use a hostname test and filter for validating SAP hostnames


berndfinger:
### Ansible Role

sap_swpm

### OS Family

All

### Feature/Enhancement request

We should implement a test and a filter for validating hostnames against the criteria mentioned in SAP note [611361](https://me.sap.com/notes/611361).

### Community participation

Happy to implement this feature

berndfinger:
Fixed in #1105.

PR #1105, opened 2025-08-13
sap_swpm: Implement a test and a filter for validating hostnames

berndfinger:
The related SAP note is 611361 and it defines the criteria for valid hostnames for SAP ABAP Platform systems.

Solves issue #1104.

### DIFF
--- a/plugins/filter/validate_sap_abap_platform_hostname.py
+++ b/plugins/filter/validate_sap_abap_platform_hostname.py
@@ -7,9 +7,11 @@ DOCUMENTATION = """
     name: validate_sap_abap_platform_hostname
     short_description: Validate hostname and return detailed results
     description:
-        - Returns detailed validation results for SAP ABAP Platform hostname criteria
+        - Returns detailed validation results for SAP ABAP Platform hostname criteria.
+        - Applies to SAP ABAP Platform (SAP note 611361) and SAP JAVA (SAP note 3216549).
         - Checks if it is a string, has 13 characters or less, does not start with a number,
-          only contains alpha characters, digits and the hyphen character, and does not contain a dot character
+          does not contain any characters other than alpha characters, digits and the hyphen character,
+          and does not contain a dot character.
     options:
         _input:
             description: The value to validate
--- a/plugins/test/valid_sap_abap_platform_hostname.py
+++ b/plugins/test/valid_sap_abap_platform_hostname.py
@@ -7,10 +7,11 @@ DOCUMENTATION = """
     name: valid_sap_abap_platform_hostname
     short_description: Test if a string is a valid SAP hostname
     description:
-        - Tests if a string meets hostname validation criteria
+        - Tests if a string meets hostname validation criteria.
+        - Applies to SAP ABAP Platform (SAP note 611361) and SAP JAVA (SAP note 3216549).
         - Checks if it is a string, has 13 characters or less, does not start with a number,
           does not contain any characters other than alpha characters, digits and the hyphen character,
-          and does not contain a dot character
+          and does not contain a dot character.
     options:
         _input:
             description: The string to test


plugins/filter/validate_sap_abap_platform_hostname.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # Convert to regular string to handle AnsibleUnsafeText
+    str_value = str(value)
+
+    # String too long
+    if len(str_value) > 13:
+        result['valid'] = False
+        result['failed_conditions'].append(f"Length must be 13 characters or less, got {len(str_value)}!")
+
+    # String stars with a number
```
ja9fuchs, line 85: typo: stars -> starts

plugins/filter/validate_sap_abap_platform_hostname.py
```diff
@@ -0,0 +1,107 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: validate_sap_abap_platform_hostname
```
marcelmamula, line 5: `sap_swpm` is used to install both ABAP and Java instances, and these new filters are called from task that does not differentiate between them.

This should be named as `validate_sap_hostname` since it is used by all types of products.

plugins/test/valid_sap_abap_platform_hostname.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def valid_sap_abap_platform_hostname(value):
+    """
+    Test if value is a valid hostname according to specified criteria:
+    1. Must be a string
+    2. Must be 13 characters or less
+    3. Must not start with a number
+    4. Must not contain a dot
+    5. Only contains alpha characters, digits and hyphen characters
```
marcelmamula, line 45: `5. Only contains alpha characters, digits, and the hyphen character`

Documentation block contains `the hyphen`.

marcelmamula (changes requested):
Code is working well and I raised few changes only for typos and naming convention.
```bash
failed: [ae1ascs] (item=sap_swpm_ers_instance_hostname) =>
    __sap_swpm_hostnames_line_item: sap_swpm_ers_instance_hostname
    ansible_loop_var: __sap_swpm_hostnames_line_item
    assertion: __sap_swpm_hostnames_var is community.sap_install.valid_sap_abap_platform_hostname
    changed: false
    evaluated_to: false
    msg: |-
        FAIL: The variable 'sap_swpm_ers_instance_hostname' is defined as:
        '2.value#is@too.long', which fails the following validation criteria:
        - Length must be 13 characters or less, got 19!
        - Must not start with a number, starts with '2'!
        - Must not contain any other character than alpha characters, digits, or hyphens!
        - Must not contain a dot character, contains at least one, at position 2!
failed: [ae1ascs] (item=sap_swpm_aas_instance_hostname) =>
    __sap_swpm_hostnames_line_item: sap_swpm_aas_instance_hostname
    ansible_loop_var: __sap_swpm_hostnames_line_item
    assertion: __sap_swpm_hostnames_var is community.sap_install.valid_sap_abap_platform_hostname
    changed: false
    evaluated_to: false
    msg: |-
        FAIL: The variable 'sap_swpm_aas_instance_hostname' is defined as:
        '['test']', which fails the following validation criteria:
        - Value must be a string, got list!
```

plugins/filter/validate_sap_abap_platform_hostname.py
berndfinger: The problem is that SAP note [611361](https://me.sap.com/notes/611361) is only about `Hostnames of SAP ABAP Platform servers`, so although we perform (and have performed) the hostname validation also for JAVA servers, the note text does not cover those.

I did some more research and found SAP note [3216549](https://me.sap.com/notes/3216549), which contains (in version 3 dated 29-Mar-2025): `The mentioned SAP Note `(611361) `can be used for AS Java servers as well, the restrictions are more likely applicable to different Operating Systems.`

But some of these restrictions, especially the hostname length, do not apply to SAP HANA servers. So the names `valid_sap_hostname` and `validate_sap_hostname` does not seem to be appropriate here, especially if we use it for plugins on the collection level.

Assuming that there is no other, specific SAP note with restrictions for SAP JAVA servers, and knowing that SAP note [3216549](https://me.sap.com/notes/3216549) only points to the SAP note for SAP ABAP Platform servers, I think there are good reasons for keeping the names `valid_sap_abap_platform_hostname` and `validate_sap_abap_platform_hostname`. We can later add another test and filter for SAP HANA servers, which could be named `valid_sap_hana_hostname` and `validate_sap_hana_hostname`, for example.

plugins/test/valid_sap_abap_platform_hostname.py
berndfinger: Thanks. I used the opportunity to rephrase all related wording in filter and test plugin after checking English grammar and style recommendations.

PR #1124, opened 2025-10-02
sap_ha_pacemaker_cluster: prepare scaleout HA setup

ja9fuchs:
## === DRAFT ===

Preparations for scale-out HA setup...

### What works
- classic scale-out HA setup of the HANA nodes on RHEL
- scale-out HA setup with 'angi' of the HANA nodes on RHEL
- classic scale-up setup tested, still works (RHEL) ;)
- basic validation of # of play hosts and failure if the majority-maker host (per variable def) is a play host

### Missing / Known issues
Sharing this as an early preview. I ran out of time to complete it yet, due to my upcoming vacation.
I will fill the gaps when back.

- Suse specific preparations
- SR state attributes are partially missing since there is no event happening that triggers the attribute update (need to test if that happens in scale-up and we overlooked it, could possibly be ignored since there is a fallback)
- no majority-maker setup covered yet, only HANA nodes - I have ideas for that


ja9fuchs:
Reworking...

PR #1127, opened 2025-11-11
sap_*_preconfigure/SLES: Improved pattern check, sanitize log output and add loop_vars

marcelmamula:
## Changes
- Add zypper pattern check using regex to ensure that we check for exact name and keep existing less accurate method as fallback.
- Replace `ignore_errors: true` with `failed_when: false` in tasks where non-zero return code is expected, hiding "failure" from log output to improve readability and to remove confusion with long 🟥 blocks of text.
- Add `loop_var` for all loops to remove deprecation warning.
- Update assert `that` sections to remove deprecation warning.
- Change debug tasks from `var` to `msg` to remove Ansible formatting and showing internal variable names.

## Tested
- SLES4SAP 15 SP7 on AWS
- SLES4SAP 16 on premise

### DIFF
--- a/roles/sap_general_preconfigure/tasks/SLES/assert-configuration.yml
+++ b/roles/sap_general_preconfigure/tasks/SLES/assert-configuration.yml
@@ -3,7 +3,7 @@
 
 - name: Assert - List required SAP Notes
   ansible.builtin.debug:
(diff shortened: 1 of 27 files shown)


berndfinger (approved):
LGTM

PR #1131, opened 2025-11-24
sap_maintain_etc_hosts: Add TAB to etc hosts and fix wrong role name in README

marcelmamula:
## Changes
- Add TAB between IP and hostnames to conform with good practices
- Fix wrong role name in README file.

## Tests
Tested on SLES 16.

## Solves
https://github.com/sap-linuxlab/community.sap_install/issues/881


marcelmamula:
Closing this PR and updating original issue with new scope.
@berndfinger This role is in need of rework as well as other roles that were reworked this year, so this new variable for separator would make sense there.

PR #1144, opened 2026-01-12
sap_swpm: Update example playbooks with supported product versions

SunnyCrockett:
I have updated `sap_swpm_product_catalog_id` to 2021 since 2020 is not supported anymore, fixed the `sap_swpm_inifile_list` for ERS and renamed `sap_swpm_inifile_list` to `sap_swpm_inifile_sections_list`.

### DIFF
--- a/roles/sap_ha_pacemaker_cluster/README.md
+++ b/roles/sap_ha_pacemaker_cluster/README.md
@@ -1149,7 +1149,7 @@ Mandatory for NetWeaver PAS cluster setup.<br>
 
 Name of the Virtual IP resource for NetWeaver PAS.<br>
 
-### sap_ha_pacemaker_cluster_vip_nwas_abap_ascs_ip_address
+### sap_ha_pacemaker_cluster_vip_nwas_ascs_ip_address
 - _Type:_ `string`<br>
 
 Virtual IP of the NetWeaver ABAP Central Services (ASCS) instance.<br>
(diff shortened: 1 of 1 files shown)


sean-freeman:
It's probably a good idea to add a README.md in the `/playbooks` subdir, explicitly describing these are just samples that are not intended to "just work" and that for click-and-go "just work" Playbooks, the Ansible Playbooks for SAP repository is where they need to visit? https://github.com/sap-linuxlab/ansible.playbooks_for_sap

As for the updates to the YAML in this PR, it's a good catch because we don't really update the samples and the var names have slightly changed since. Which were updated in AP4S > https://github.com/sap-linuxlab/ansible.playbooks_for_sap/blob/main/deploy_scenarios/sap_s4hana_distributed_ha/ansible_extravars.yml#L277-L295

SunnyCrockett:
@berndfinger Going directly to 2023 is fine, I guess. In my tests I used 2021, but 2023 will sure work for a longer time.

marcelmamula:
@berndfinger I have never used these playbooks, but just looking at them, they need more work than just simple product version change.
> **Passwords must be removed as soon as possible.**

@SunnyCrockett This PR has 12 commits for few lines of changes so it would be good if you decided on scope for this PR, before leaving it without reviewers and adding more commits in future.
I would personally cancel this PR and recreate new PR with one commit to have it clean, but that is me not liking reverts.

PR #1147, opened 2026-01-16
sap_hana_install: Revert wrong conditional from PR#1138

marcelmamula:
## Changes 
https://github.com/sap-linuxlab/community.sap_install/pull/1138 was reworking tags, but it also touched conditional by mistake.
Post tasks should always execute, not only on existing installations.

**NOTE: This is breaking bug which will require 1.8.1 Release with other new bugs identified**

## Ref
https://github.com/sap-linuxlab/community.sap_install/blame/dev/roles/sap_hana_install/tasks/main.yml#L73


berndfinger (approved):
LGTM.

PR #1149, opened 2026-01-23
sap_swpm: Allow user defined environment variables when executing swpm

rob0d:
This PR adds relatively simple functionality to SWPM which allows custom defined environment shell variables to be set before executing SWPM. This is something we have briefly discussed @berndfinger more than a year ago.

Rationale:
- Many production systems have noexec and nosuid flag set /tmp filesystem which is preventing SWPM from starting. As per SAP Notes [3651688](http://me.sap.com/notes/3651688) and [2063079](http://me.sap.com/notes/2063079) solution is to set TMP environment variable to a more suitable location.
- Some database libraries e.g. for DB/2 zOS require environment variables to be set during installation (e.g. sdb_ssfs_connect).

This change makes a generic provision to define any helpful environment variables.

 @marcelmamula @berndfinger @sean-freeman the variable is currently called sap_swpm_swpm_env_vars which is in-line with the naming convention used in this role (e.g. `sap_swpm_swpm_installation_type: , sap_swpm_swpm_installation_header:     sap_swpm_swpm_command_virtual_hostname:, sap_swpm_swpm_command_mp_stack:` etc). If you want to start moving towards a different naming convention, happy to change it, but I am mentioning it because Marcel had concerns about the name of this variable.

### DIFF
--- a/roles/sap_swpm/README.md
+++ b/roles/sap_swpm/README.md
@@ -355,6 +355,7 @@ Example:
 ```yaml
 sap_swpm_env_vars:
   TMP: /var/tmp
+```
 
 ### Variables specific to SAP Netweaver
 


rob0d:
> @rob0d Good catch on those SNotes about `/tmp` path, this is relative standard challenge when using hardened OS Images and truthfully I didn't come across those notes before. This would be the main use case for this PR - so I'd cite the SNotes in the commented code?

I will comment to the code.

> 
> The IBM Db2 for zOS use case is not something we would test given it only has a UNIX-compliant addendum environment thereby way outside the Linux scope we aim for (same reason we do not scope WinS). The only person that would've tested it on zOS or AIX would've been myself - and I definitely didn't test it 🙂 If you tested it and it worked, that would be very interesting/cool.

This is on zLinux and Linuxx86_64 app servers running against DB2 zOS DB. It works fine as it doesn't really touch the DB apart from talking to it, so no need for us to test it further. If people need to have different env variables, they can use this functionality.

marcelmamula:
@rob0d yes empty dict is good when you sanitize it with `| length > 0` when used.
@berndfinger raised issue upon my request https://github.com/sap-linuxlab/community.sap_install/issues/1150 and already did some changes there.

In your case it should be (I am up for first simple name):
- either `sap_swpm_env_vars`
- or `sap_swpm_env_vars_dictionary`
- or `sap_swpm_env_vars_dict`

PR #1148, opened 2026-01-21
collection: Update readme formatting to align with galaxy format

marcelmamula:
## Changes
- All README.md documents were updated to align with formatting in Ansible Galaxy to avoid broken tables and lists.
- Markdown `  ` newline was replaced with `<br>`. Following command can be used to find them `grep -rnE "  $" --include="*.md" .`
- Update collection requirements table for better formatting.

**NOTE:** `sap_ha_pacemaker_cluster` role will be updated in upcoming PR along with STONITH improvements.

PR #1151, opened 2026-01-27
sap_ha_pacemaker_cluster: Add fence_aws for SLES 16 and remove Tech Debt

marcelmamula:
## Changes
- Redesign AWS fence agent dictionary to allow for different definition for SLES 15 and 16
- Define `fence_aws` for SLES 16 on AWS
- **Fix incorrect variable loading that conflicted with precedence and remove appends.**
  - We will never append to user defined variables as there can be intent for testing or debugging. Example: No longer add fencing properties if `sap_ha_pacemaker_cluster_cluster_properties` is defined already.
- Remove Tech Debt for deprecated variables
- Update arg spec file to align with new readme update tool.
- Update README.md to align with https://github.com/sap-linuxlab/community.sap_install/pull/1148

## Tests
Tested on:
- SLES_SAP 15 SP7 on AWS, GCP and Azure.
- SLES_SAP 16.0 on GCP and Azure.

**NOTE: AWS cannot be currently tested because of issues with Resource Agents! **


ja9fuchs (approved):
👍

PR #1158, opened 2026-02-04
sap_ha_pacemaker_cluster: Reorganize folder structure for simplicity

marcelmamula:
## Changes
- File and folder structure of role was redesigned for simpler management of all different files.
- Each operation has separate folder with prefixes for visibility in IDE tabs.

@ja9fuchs Please keep note of these comments I made, because this has to be removed by splitting it into non-changing code for variables section and then actual code that does changes.
There was very specific section in main that designated at which point changes happen, but this task does them before that.

## Tests
Tested on SLES_SAP 15 SP7 on AWS with ASCS/ERS and HANA HA clusters.


ja9fuchs (approved):
Looks good and I successfully tested a HANA cluster setup on RHEL 9 on AWS.

Issue #1146, opened 2026-01-16
sap_swpm: Conditions not sanitized for 2.19


marcelmamula:
## Issue
The following task is not sanitized for 2.19 and it fails.

https://github.com/sap-linuxlab/community.sap_install/blob/76cc9677c9f8c7a646a7728742c2ec69ea8eb8dd/roles/sap_swpm/tasks/pre_install/install_type.yml#L87

```bash
TASK [community.sap_install.sap_swpm : SAP SWPM Pre Install - Assert that necessary variable for HA is defined] **************************
 [started TASK: community.sap_install.sap_swpm : SAP SWPM Pre Install - Assert that necessary variable for HA is defined on ae1ascs]
[ERROR]: Task failed: Conditional result (True) was derived from value of type 'str' at 'extravars.yml:90:34'. Conditionals must have a boolean result.
```

One more issue with not sanitized conditional:
https://github.com/sap-linuxlab/community.sap_install/blob/6384d652e82a1fe95f037d774c868276bac05f1a/roles/sap_swpm/tasks/post_install/firewall.yml#L26

3rd item: Same file shows missing defaults, because if WD is not used it should never loop over undefined var.
https://github.com/sap-linuxlab/community.sap_install/blob/6384d652e82a1fe95f037d774c868276bac05f1a/roles/sap_swpm/tasks/post_install/firewall.yml#L42

## Fix
Add length validation for `str`.

## Note
Please check if there are any other leftovers that were not sanitized.

marcelmamula:
@berndfinger Added one more conditional that fails on 2.19
https://github.com/sap-linuxlab/community.sap_install/blob/6384d652e82a1fe95f037d774c868276bac05f1a/roles/sap_swpm/tasks/post_install/firewall.yml#L26

PR #1159, opened 2026-02-06
sap_swpm: Sanitize conditions for 2.19

berndfinger:
Solves issue #1146 .


marcelmamula (approved):
LGTM. Ignoring firewall steps as they will be replaced with same workflow as hana and HSR roles.

PR #1167, opened 2026-02-18
sap_hana_install: Fix hana versions mismatch and replace injected vars

marcelmamula:
## Issue
SAP mentions '0' in 'hdblcm --list systems', while having '00' in manifest file for patch level 0, which would lead to version mismatch.

Example:
```
FAIL: Existing installation version mismatch!
Existing version: 2.00.077.0
Expected version: 2.00.077.00
```

## Changes
- Replace literal string check with integer segmented comparison, that will solve issue with zeroes.
- Replace injected ansible vars like `ansible_hostname` with ansible facts like `ansible_facts['hostname'] which is new warning in 2.20.

## Linting Note
Linting warning is very useful for catching injected vars, but there is new warning when even ansible_facts are before filter.
`[WARNING]: Falling back to Ansible unique filter as Jinja2 one failed: 'ansible_facts' is undefined`

This was triggered for `pre_tasks/identify_hosts.yml` and resolution is defaulting before filter: `[ansible_facts['hostname'] | d(''), inventory_hostname | d(''), inventory_hostname_short | d('')] | unique`

### DIFF
--- a/roles/sap_hana_install/tasks/pre_tasks/check_version.yml
+++ b/roles/sap_hana_install/tasks/pre_tasks/check_version.yml
@@ -38,11 +38,11 @@
           - (__sap_hana_install_fact_manifest_version.split('.') | map('int') | list) ==
             (__existing_version.split('.') | map('int') | list)
         success_msg: |
-          PASS: Existing installation has expected version.
-          Expected: {{ __sap_hana_install_fact_manifest_version }}
+          PASS: The existing installation has the expected version.
           Existing: {{ __existing_version }}
+          Expected: {{ __sap_hana_install_fact_manifest_version }}
         fail_msg: |
-          FAIL: Existing installation version mismatch!
+          FAIL: The existing installation version does not match the expected version.
           Existing: {{ __existing_version }}
           Expected: {{ __sap_hana_install_fact_manifest_version }}
       vars:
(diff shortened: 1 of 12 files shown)


roles/sap_hana_install/tasks/pre_tasks/check_version.yml
```diff
@@ -26,19 +26,29 @@
...
+          # Compare as lists of integers: [2, 0, 77, 0] == [2, 0, 77, 0]
+          - (__sap_hana_install_fact_manifest_version.split('.') | map('int') | list) ==
+            (__existing_version.split('.') | map('int') | list)
         success_msg: |
           PASS: Existing installation has expected version.
-          Expected version: {{ __sap_hana_install_fact_manifest_version }}.
+          Expected: {{ __sap_hana_install_fact_manifest_version }}
+          Existing: {{ __existing_version }}
         fail_msg: |
-          FAIL: Existing installation does not have expected version!
```
berndfinger: Please add "the" before "Exiting" and "expected", for correct grammar and better readability.
New: "The existing installation does not have the expected version."

berndfinger (changes requested):

berndfinger (approved):
LGTM.

PR #1195, opened 2026-03-23
sap_maintain_etc_hosts Rework

berndfinger:
This PR contains the following changes:
- Preserve existing comments in valid /etc/hosts lines
- Replace `shell` and `command` modules by native Ansible modules
- Replace `omit` by `[]` when defining the variable `sap_maintain_etc_hosts_list`, for consistency with existing roles and easier handling of the default
- Use `ansible.builtin.slurp` for reading the contents of `/etc/hosts` instead of reading the file in multiple tasks
- Change variable names
- New `ipv6` regex pattern to also support compressed notation
- Some minor changes


marcelmamula (changes requested):
I have found issues that range from new changes back to original code that is not sufficient.
It is important to decide what is intent of this PR, if proper refactoring/rework or just smaller improvements.

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
```diff
@@ -1,205 +1,193 @@
...
 
-- name: Ensure that the entry in hosts file is correct
+- name: Ensure that the entry for node_ip in {{ __sap_maintain_etc_hosts_file }} is correct
   ansible.builtin.lineinfile:
     path: "{{ __sap_maintain_etc_hosts_file }}"
     regexp: '^{{ thishost.node_ip }}\s'
-    line: "{{ thishost.node_ip }} {{ thishost.node_name }}.{{ __sap_maintain_etc_hosts_domain }} {{ thishost.node_name }} {{ __sap_maintain_etc_hosts_aliases }} {{ __sap_maintain_etc_hosts_comment }}"
+    line: "{{ thishost.node_ip }} {{ thishost.node_name }}.{{ __sap_maintain_etc_hosts_fact_domain }} {{ thishost.node_name }} {{ __sap_maintain_etc_hosts_fact_aliases }} {{ __sap_maintain_etc_hosts_fact_comment }}"
     backup: true
   when:
```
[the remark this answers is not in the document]
berndfinger: This is due to the default of `merge` for `alias_mode`. If set to `overwrite`, `alias2` will be removed from `/etc/hosts` in your example.

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
```diff
@@ -1,205 +1,193 @@
...
   become: true
 
 # After all nodes are added or deleted, run the consistency check against the hosts file
+
+- name: Get the content /etc/hosts after possible modification
+  ansible.builtin.slurp:
+    src: "{{ __sap_maintain_etc_hosts_file }}"
+  register: __sap_maintain_etc_hosts_register_file_post_modification
+
 - name: Check for duplicate or missing entries of hostname and fqdn in {{ __sap_maintain_etc_hosts_file }}
```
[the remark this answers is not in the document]
berndfinger: We allow for the IP address to be more than once, in which case we do not modify /etc/hosts. But we do not allow hostname or FQDN to appear more than once in /etc/hosts.
I will catch this case in the next commit.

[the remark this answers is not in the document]
marcelmamula: Sure, just make sure to ensure that no change occurs if it would lead to duplicate. I think that checking after changes is bad idea.

berndfinger:
@marcelmamula I think with the commits after your review, all the open issues should have been addressed.

roles/sap_maintain_etc_hosts/tasks/main.yml
```diff
@@ -57,24 +16,203 @@
...
+             | selectattr('state', 'eq', 'present') | list)
+           }}"
+
+    - name: Hosts File - Display all hosts to be present
+      ansible.builtin.debug:
+        msg:
+          - "{{ __sap_maintain_etc_hosts_fact_present }}"
+        verbosity: 2
+
+    - name: Hosts File - Ensure 'node_ip' and 'node_name' are defined for hosts with state!='absent'
```
marcelmamula: This task does not execute on `state!='absent'` while claiming it does. It executes on all host facts.

Task name is not very good with `state!='absent'`. I would rather say:
`with 'state=present'` or `with state not 'absent'`

Var even says `__present_item` so it is `present`, not `!=absent`.

roles/sap_maintain_etc_hosts/tasks/main.yml
```diff
@@ -57,24 +16,203 @@
...
+                    'defined_ip': __node.node_ip,
+                    'actual_ip': __host.ip
+                  }) -%}
+                {%- endif -%}
+              {%- endfor -%}
+            {%- endif -%}
+          {%- endfor -%}
+          {{ mismatches }}
+
+    - name: Hosts File - Assert that there are no IP address mismatches
```
marcelmamula: Why are we checking for attemptto set different IP for existing hostname, but accepting setting different hostname for existing IP?

PASS:
```
10.10.10.12 host1 > 10.10.10.12 host2
```

FAIL:
```
10.10.10.12 host1 > 10.10.10.13 host1
```

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
```diff
@@ -1,205 +1,199 @@
...
-# 2. loop over all hostname entries in the selected line (2 bis NF=last element in line)
-# 3. stop looping when a comment sign is found (because these are comments)
-# 4. print an element if it is not the hostname or FQDN we want to add
-#
-# => __sap_maintain_etc_hosts_register_aliases.stdout contains a list of aliases of thishost.node_ip
-#
-# trunk-ignore(checkov/CKV2_ANSIBLE_3)
-- name: Merge existing aliases with new alias list
-  when: __sap_maintain_etc_hosts_alias_mode != "overwrite"
+    __sap_maintain_etc_hosts_fact_alias_mode: "{{ __etc_hosts_item.alias_mode | default('merge') }}"
```
marcelmamula, line 63: Merge should never be allowed if IP + Host has changed.

This is result of test described above:
```
# Action
10.10.10.12 host1 > 10.10.10.12 host2

# Before
10.10.10.12 host1.domain host1 alias1 alias2

# After
10.10.10.12 host2.domain host2 host1.domain host1 alias1 alias2
```

roles/sap_maintain_etc_hosts/tasks/main.yml
```diff
@@ -57,24 +16,203 @@
...
-  loop: "{{ sap_maintain_etc_hosts_list }}"
-  loop_control:
-    label: "{{ thishost.node_name | d(thishost.node_ip) }}"
-    loop_var: thishost
+    ( sap_maintain_etc_hosts_list | trim == '') or
+    ( sap_maintain_etc_hosts_list | length == 0)
+  block:
+    - name: Hosts File - Preserve existing comment if no new comment is provided
+      ansible.builtin.set_fact:
+        __sap_maintain_etc_hosts_fact_comment:
```
marcelmamula, line 305: Stopped working with current commits. All comments are being removed.

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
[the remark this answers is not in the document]
marcelmamula: Merge is still not working predictably. It works only sometimes.

```bash
TASK [sap_maintain_etc_hosts : Hosts File - Presence - Display host, DNS domainname and IP address before the modification] ********************************
 [started TASK: sap_maintain_etc_hosts : Hosts File - Presence - Display host, DNS domainname and IP address before the modification on sles16demo-gmc]
ok: [sles16demo-gmc] =>
    msg:
    - hostname   = 'host111'
    - domain     = 'domain'
    - ip         = '10.10.10.10'
    - comment    = ''
    - aliases    = ''
    - alias mode = 'merge'
```

What it actually did:
```
# Before
10.10.10.10 host11.domain host11

# After
10.10.10.10 host111.domain host111
```

marcelmamula (changes requested):
Some problems were resolved, but new bugs appeared.

roles/sap_maintain_etc_hosts/tasks/main.yml
berndfinger: The variable `__sap_maintain_etc_hosts_fact_present`, over which we are looping here, is filled in two of the previews few tasks as follows:

In https://github.com/berndfinger/community.sap_install/blob/f04b01db2931117fce5dac4b3bf2dc214d1b6a1a/roles/sap_maintain_etc_hosts/tasks/main.yml#L149, we are adding all dicts from `sap_maintain_etc_hosts_list` which do not have a member named `state`.
Then, in the task which follows, we are adding all dicts from `sap_maintain_etc_hosts_list` which are defined with `state` `present`.

For each of those, we demand `node_ip` and `node_name` to be defined.

Maybe:
```
    - name: Hosts File - Ensure 'node_ip' and 'node_name' are defined for hosts to be added
```

is a good task name here.

marcelmamula: Yes that is clearer name. Adding literal conditions into task names is not good practice, it should be readable instead.

roles/sap_maintain_etc_hosts/tasks/main.yml
berndfinger: Although `/etc/hosts` may contain two lines with the same hostname, according to my research, the resolvers typically appear to use only the first matching entry. No such restriction appears to be present for `/etc/hosts` lines which contain two lines for the same IP address. It even appears to be best practices.

The comment in line https://github.com/berndfinger/community.sap_install/blob/b22db19999f94cfe6aebc1809d75fd40b69f89b9/roles/sap_maintain_etc_hosts/tasks/update_host_present.yml#L107 (current main branch) seems to confirm this.

roles/sap_maintain_etc_hosts/tasks/main.yml
berndfinger: In which case?

- If `sap_maintain_etc_hosts_list` is defined with content (block starting at https://github.com/berndfinger/community.sap_install/blob/f04b01db2931117fce5dac4b3bf2dc214d1b6a1a/roles/sap_maintain_etc_hosts/tasks/main.yml#L24), comments will be removed. -> **OK for me**
- If `sap_maintain_etc_hosts_list` is not undefined or has no content (block starting at https://github.com/berndfinger/community.sap_install/blob/f04b01db2931117fce5dac4b3bf2dc214d1b6a1a/roles/sap_maintain_etc_hosts/tasks/main.yml#L182), comments for the current host will be preserved -> **This is what I want**.

marcelmamula: I am testing with `sap_maintain_etc_hosts_list` and comments should not be removed.

You cannot have 1 role resulting in 2 different outcomes for something that user cannot define. Comments should be preserved in every case.

roles/sap_maintain_etc_hosts/tasks/main.yml
marcelmamula: I would still keep it unique on both ends since hosts/aliases should be on one line.

roles/sap_maintain_etc_hosts/tasks/main.yml
berndfinger: Sorry, there was a typo in my reply.
Here is the correct version:
--->
- If `sap_maintain_etc_hosts_list` is defined with content (block starting at https://github.com/berndfinger/community.sap_install/blob/f04b01db2931117fce5dac4b3bf2dc214d1b6a1a/roles/sap_maintain_etc_hosts/tasks/main.yml#L24), comments will be removed. -> OK for me
- If `sap_maintain_etc_hosts_list` is undefined or has no content (block starting at https://github.com/berndfinger/community.sap_install/blob/f04b01db2931117fce5dac4b3bf2dc214d1b6a1a/roles/sap_maintain_etc_hosts/tasks/main.yml#L182), comments for the current host will be preserved -> This is what I want.

<---

From your last answer, I assume you are fine with preserving existing commends if `sap_maintain_etc_hosts_list` is undefined, has no content, has no member named `node_comment`, or has a member named `node_comment` but its content is empty. So let's focus on the other case:

Do you suggest to implement one of the following behavior if `sap_maintain_etc_hosts_list` is defined with a comment?
- the existing comment should be preserved and the new comment from `sap_maintain_etc_hosts_list` should be appended after a space character, or:
- the new comment from `sap_maintain_etc_hosts_list` should be put to the beginning of comment section and a space and the existing comment should be added at the end?

marcelmamula: Comments could be potentially removed only in 1 case (when `node_comment` is defined and non-empty), but that one is also not valid because https://github.com/berndfinger/community.sap_install/blob/f04b01db2931117fce5dac4b3bf2dc214d1b6a1a/roles/sap_maintain_etc_hosts/defaults/main.yml#L28 mentions appending, not replacing.

So your new preservation for comments, that was added, is good but it cannot be selective, it has to happen for all cases. Be it `sap_maintain_etc_hosts_list` invalid (undefined or empty) or valid.

roles/sap_maintain_etc_hosts/tasks/main.yml
berndfinger: Before we change the behavior to no longer allow more than one line containing the same IP address, we need to provide proper justification, not just because of that existing comment.

In the meantime, I found sources which appear to suggest that there should be only one line per Iv4P address. For example, the first paragraph of `man -s5 hosts` (RHEL 10) mentions that there should be only one line per host, and this statement could be read in the sense that there should also be exactly one line per IPv4 address. In the following chapter, it says, "If required, a host may have two separate entries in this file; one for each version of the Internet Protocol (IPv4 and IPv6)."

@marcelmamula What is your suggestion for handling cases where there is more than one line with the same IP address in /etc/hosts: fail the role or skip modifying this record?

marcelmamula: @berndfinger I am 100% for forcing 1 line per IP, because having multiple lines breaks reverse lookup.
- during pre_tasks: Check if more than one exists, fail if so.
- during present task: Update existing IP line or add new if not present.

```bash
sles16demo-gmc:~ # cat /etc/hosts | grep 10.10.10.10
10.10.10.10 host1.domain host1
10.10.10.10 host2.domain host2 alias1 alias2

sles16demo-gmc:~ # getent hosts 10.10.10.10
10.10.10.10     host1.domain host1
sles16demo-gmc:~ #
```

roles/sap_maintain_etc_hosts/tasks/main.yml
berndfinger: That description in `defaults/main.yml` reads:

`String which is appended to line in hosts after comment string`, and I understood it as follows:
 `String which is appended to line in hosts after the comment character`

If we allow the existing comment to be preserved and a new comment (as defined in `sap_maintain_etc_hosts_list` to be added), we lose idempotency: The comment gets longer and longer in each playbook run.

So I suggest the following logic:
- `sap_maintain_etc_hosts_list` is defined AND `node_comment` is defined with length > 0: Do **not** preserve the existing comment but use the content of `node_comment`.
- `sap_maintain_etc_hosts_list` is defined AND `node_comment` is not defined or has a length of 0, **OR**:
- `sap_maintain_etc_hosts_list` is not defined or is none or is an empty string or has a length of 0: Either use an existing comment or use the default of `managed by Ansible role sap_maintain_etc_hosts`.

marcelmamula: You will not lose idempotency if you check for `comment + end of line character` in that line, before appending.
If someone added anything to comment in between runs, it is still correct to append from idempotency run, but if nothing was changed then it would recognize that comment already exists.

`String which is appended to line in hosts after comment string` = String that will be added after existing `# comment# string, not after `#` character. That would be different wording.

If you will not change that explanation, then append must always be done, never replace of comment.

berndfinger: Should be solved by commit 43233154.

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
berndfinger: Should be solved by commits 923b993f, fc20ed15 and ef23c23d.

roles/sap_maintain_etc_hosts/tasks/main.yml
berndfinger: We finally agreed on only allowing one line per IP address. The code change is in 5776ff20.

roles/sap_maintain_etc_hosts/tasks/main.yml
berndfinger: Fixed in 590aca20.

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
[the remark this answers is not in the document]
berndfinger: Solved, see https://github.com/sap-linuxlab/community.sap_install/pull/1195#issuecomment-4171515396

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
```diff
@@ -1,205 +1,193 @@
...
+            | map('split') | flatten
+            | reject('equalto', thishost.node_ip)
+            | reject('equalto', thishost.node_name)
+            | reject('equalto', thishost.node_name ~ '.' ~ __sap_maintain_etc_hosts_fact_domain) }}"
 
     - name: Add defined aliases
       ansible.builtin.set_fact:
-        __sap_maintain_etc_hosts_aliases: "{{ (__sap_maintain_etc_hosts_register_aliases.stdout.split(' ')
-           + thishost.aliases | d([])) | unique | join(' ') | trim }}"
+        __sap_maintain_etc_hosts_fact_aliases:
```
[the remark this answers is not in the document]
berndfinger: Solved, see https://github.com/sap-linuxlab/community.sap_install/pull/1195#issuecomment-4171515396.

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
```diff
@@ -1,205 +1,193 @@
...
   loop_control:
     loop_var: line_item
 
-- name: Ensure that the entry in hosts file is correct
+- name: Ensure that the entry for node_ip in {{ __sap_maintain_etc_hosts_file }} is correct
   ansible.builtin.lineinfile:
     path: "{{ __sap_maintain_etc_hosts_file }}"
     regexp: '^{{ thishost.node_ip }}\s'
-    line: "{{ thishost.node_ip }} {{ thishost.node_name }}.{{ __sap_maintain_etc_hosts_domain }} {{ thishost.node_name }} {{ __sap_maintain_etc_hosts_aliases }} {{ __sap_maintain_etc_hosts_comment }}"
+    line: "{{ thishost.node_ip }} {{ thishost.node_name }}.{{ __sap_maintain_etc_hosts_fact_domain }} {{ thishost.node_name }} {{ __sap_maintain_etc_hosts_fact_aliases }} {{ __sap_maintain_etc_hosts_fact_comment }}"
```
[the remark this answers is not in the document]
berndfinger: Solved, see https://github.com/sap-linuxlab/community.sap_install/pull/1195#issuecomment-4171515396.

roles/sap_maintain_etc_hosts/tasks/main.yml
```diff
@@ -57,24 +16,316 @@
...
+             | selectattr('state', 'eq', 'present') | list)
+           }}"
+
+    - name: Hosts File - Display all hosts to be present
+      ansible.builtin.debug:
+        msg:
+          - "{{ __sap_maintain_etc_hosts_fact_present }}"
+        verbosity: 2
+
+    - name: Hosts File - Ensure 'node_ip' and 'node_name' are defined for hosts to be added
```
marcelmamula, line 215: You are checking if defined so there is no reason to run it in default mode because values do not matter. Please add `quiet: true` and it will remove unnecessary output.
https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/assert_module.html#parameter-quiet

```bash
TASK [sap_maintain_etc_hosts : Hosts File - Ensure 'node_ip' and 'node_name' are defined for hosts to be added] ********************************************
 [started TASK: sap_maintain_etc_hosts : Hosts File - Ensure 'node_ip' and 'node_name' are defined for hosts to be added on sles16demo-gmc]
ok: [sles16demo-gmc] => (item={'node_ip': '10.10.10.12', 'node_name': 'host22', 'aliases': ['alias1', 'alias3'], 'alias_mode': 'overwrite'}) =>
    __present_item:
        alias_mode: overwrite
        aliases:
        - alias1
        - alias3
        node_ip: 10.10.10.12
        node_name: host22
    ansible_loop_var: __present_item
    changed: false
    msg: All assertions passed
ok: [sles16demo-gmc] => (item={'node_ip': '10.10.10.10', 'node_name': 'host11', 'state': 'present'}) =>
    __present_item:
        node_ip: 10.10.10.10
        node_name: host11
        state: present
    ansible_loop_var: __present_item
    changed: false
    msg: All assertions passed
```

marcelmamula, line 215: Make sure to add short loop label to avoid these long dictionary outputs. 
`label: "{{ __present_item.node_ip | d('N/A') }}"` should be enough 

Current output shows full dictionary:
```
TASK [sap_maintain_etc_hosts : Hosts File - Ensure 'node_ip' and 'node_name' are defined for hosts to be added] ********************************************
 [started TASK: sap_maintain_etc_hosts : Hosts File - Ensure 'node_ip' and 'node_name' are defined for hosts to be added on sles16demo-gmc]
ok: [sles16demo-gmc] => (item={'node_ip': '10.10.10.12', 'node_name': 'host22', 'aliases': ['alias1', 'alias3'], 'alias_mode': 'overwrite'})
...
ok: [sles16demo-gmc] => (item={'node_ip': '10.10.10.10', 'node_name': 'host11', 'state': 'present'}) =>
...
```

roles/sap_maintain_etc_hosts/tasks/main.yml
```diff
@@ -57,24 +16,316 @@
...
+                  {%- if entry.ip != __host.node_ip and __hostname not in duplicates -%}
+                    {%- set _ = duplicates.append(__hostname) -%}
+                  {%- endif -%}
+                {%- endfor -%}
+              {%- endif -%}
+            {%- endfor -%}
+          {%- endfor -%}
+          {{ duplicates }}
+
+    - name: Hosts File - Assert that no IP address conflicts for hostnames and aliases
```
marcelmamula, line 247: Something is wrong with previous task, because resulting list is alias? That message is also not clear what is actual problem, because it just says `alias1`, not which IP is in conflict and what is a problem.

```bash
fatal: [sles16demo-gmc]: FAILED! =>
    assertion: __sap_maintain_etc_hosts_fact_duplicate_hostnames | length == 0
    changed: false
    evaluated_to: false
    msg: 'FAIL: IP address conflicts found for: alias1'
```

This is /etc/hosts:
```
10.10.10.10 host1.domain host1 # comment
10.10.10.10 host2.domain host2 alias1 alias2
```

This is role vars:
```
        sap_maintain_etc_hosts_list:
          - node_ip: 10.10.10.10
            node_name: host11
            state: present
          - node_ip: 10.10.10.12
            node_name: host22
            aliases:
              - alias1
              - alias3
            alias_mode: overwrite
```

roles/sap_maintain_etc_hosts/tasks/main.yml
marcelmamula: This applies to all other asserts, not just this one.

roles/sap_maintain_etc_hosts/tasks/main.yml
```diff
@@ -57,24 +16,316 @@
...
+                    'defined_ip': __node.node_ip,
+                    'actual_ip': __host.ip
+                  }) -%}
+                {%- endif -%}
+              {%- endfor -%}
+            {%- endif -%}
+          {%- endfor -%}
+          {{ mismatches }}
+
+    - name: Hosts File - Assert that there are no IP address mismatches for hostnames
```
marcelmamula, line 151: Output is direct list of dictionaries which is not readable. Can you change it to `j2 for loop` and show them properly?
Current:
```
    msg: 'FAIL: IP address mismatches found: [{''hostname'': ''host2'', ''defined_ip'':
        ''10.10.10.12'', ''actual_ip'': ''10.10.10.11''}]'
```

Example of better output:
```
AIL: IP address mismatches found:
Hostname 'host2' found with IP '10.10.10.12', but '10.10.10.11' was defined.
Hostname 'host1' found with IP '10.10.10.13', but '10.10.10.14' was defined.
```

roles/sap_maintain_etc_hosts/tasks/main.yml
```diff
@@ -57,24 +16,316 @@
...
+        fail_msg: "FAIL: member with state 'absent' is defined without 'node_ip' or 'node_name': {{ __absent_item }}."
+      loop: "{{ sap_maintain_etc_hosts_list | selectattr('state', 'defined') | selectattr('state', 'eq', 'absent') }}"
+      loop_control:
+        loop_var: __absent_item
+
+    - name: Hosts File - Get list of hosts to be present with no state definition
+      ansible.builtin.set_fact:
+        __sap_maintain_etc_hosts_fact_present: "{{ sap_maintain_etc_hosts_list | selectattr('state', 'undefined') | list }}"
+
+    - name: Hosts File - Add list of hosts with state=present
```
marcelmamula, line 200: This task changes order of items in input list, most likely because of difference filter. We should retain what users defined.

Input
```
        sap_maintain_etc_hosts_list:
          - node_ip: 10.10.10.10
            node_name: host1
            state: present

          - node_ip: 10.10.10.11
            node_name: host2
            aliases:
              - alias1
              - alias2
```

Actual execution is in reversed order:
```
TASK [sap_maintain_etc_hosts : Hosts File - Maintain entries in "/etc/hosts"] ******************************************************************************
 [started TASK: sap_maintain_etc_hosts : Hosts File - Maintain entries in "/etc/hosts" on sles16demo-gmc]
included: /home/mmamula/workspace/bernd/roles/sap_maintain_etc_hosts/tasks/update_host_present.yml for sles16demo-gmc => (item=host2)
included: /home/mmamula/workspace/bernd/roles/sap_maintain_etc_hosts/tasks/update_host_present.yml for sles16demo-gmc => (item=host1)
```

Visible in all tasks
```
TASK [sap_maintain_etc_hosts : Hosts File - Presence - Check for duplicate or missing entries of hostname and fqdn in /etc/hosts] **************************
 [started TASK: sap_maintain_etc_hosts : Hosts File - Presence - Check for duplicate or missing entries of hostname and fqdn in /etc/hosts on sles16demo-gmc]
ok: [sles16demo-gmc] => (item=host2.domain) =>
    __sap_maintain_etc_hosts_line_item: host2.domain
    ansible_loop_var: __sap_maintain_etc_hosts_line_item
    changed: false
    msg: 'PASS: Hostname or FQDN ''host2.domain'' exists exactly once.'
ok: [sles16demo-gmc] => (item=host2) =>
    __sap_maintain_etc_hosts_line_item: host2
    ansible_loop_var: __sap_maintain_etc_hosts_line_item
    changed: false
    msg: 'PASS: Hostname or FQDN ''host2'' exists exactly once.'
```

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
```diff
@@ -1,205 +1,205 @@
...
-    - "{{ thishost.node_name }}"
+    - "{{ __etc_hosts_item.node_name }}.{{ __sap_maintain_etc_hosts_fact_domain }}"
+    - "{{ __etc_hosts_item.node_name }}"
   loop_control:
     loop_var: line_item
+  when:
+    - not ansible_check_mode
 
-- name: Ensure that the entry in hosts file is correct
+- name: Hosts File - Presence - Ensure that the entry for node_ip in {{ __sap_maintain_etc_hosts_file }} is correct
```
marcelmamula, line 146: # Major problem
This task should never execute in this situation. We are not arbiters of proper /etc/hosts formatting (more so that I like TAB more than spaces), so if users have valid identical line, we should not overwrite it with our formatting. Same happens when `sap_maintain_etc_hosts_list` is defined or not.

Example for role without `sap_maintain_etc_hosts_list` defined. Take note of 2 spaces between comment hash.
```
10.144.74.25    sles16demo-gmc.domain sles16demo-gmc  # my comment
```

Log
```
TASK [sap_maintain_etc_hosts : Hosts File - Presence - Ensure that the entry for node_ip in /etc/hosts is correct] *****************************************
 [started TASK: sap_maintain_etc_hosts : Hosts File - Presence - Ensure that the entry for node_ip in /etc/hosts is correct on sles16demo-gmc]
changed: [sles16demo-gmc]
```

Output where TAB is replaced with space and 2 comment hash spaces are replaced with 1 space.
```
10.144.74.25 sles16demo-gmc.domain sles16demo-gmc # my comment
```

marcelmamula (changes requested):

roles/sap_maintain_etc_hosts/tasks/main.yml
```diff
@@ -57,24 +16,316 @@
...
+
+    - name: Hosts File - Ensure 'node_name' elements do not contain a dot
+      ansible.builtin.assert:
+        that: "'.' not in __hosts_item.node_name"
+        fail_msg: |
+          FAIL: node_name '{{ __hosts_item.node_name }}' contains a dot, which is not allowed!"
+      loop: "{{ sap_maintain_etc_hosts_list }}"
+      loop_control:
+        loop_var: __hosts_item
+      when: __hosts_item.node_name is defined
```
marcelmamula, line 48: Missing `None` validation.
`node_name: `

```
failed: [sles16demo-gmc] (item={'node_ip': '10.10.10.11', 'node_name': None, 'aliases': ['alias1', 'alias2'], 'alias_mode': 'overwrite'}) =>
    __hosts_item:
        alias_mode: overwrite
        aliases:
        - alias1
        - alias2
        node_ip: 10.10.10.11
        node_name: null
    ansible_loop_var: __hosts_item
    changed: false
    msg: 'Task failed: Error rendering expression: argument of type ''NoneType'' is not
        iterable'
```

roles/sap_maintain_etc_hosts/tasks/main.yml
```diff
@@ -57,24 +16,316 @@
...
+          {%- for parts in __sap_maintain_etc_hosts_fact_hosts_intermediate -%}
+            {%- if parts | length > 1 -%}
+              {%- for hostname in parts[1:] -%}
+                {%- set _ = result.append({'ip': parts[0], 'hostname': hostname}) -%}
+              {%- endfor -%}
+            {%- endif -%}
+          {%- endfor -%}
+          {{ result }}
+
+    - name: Hosts File - Identify IP address mismatches of hostnames between sap_maintain_etc_hosts_list and {{ __sap_maintain_etc_hosts_file }} entries
```
marcelmamula, line 126: Missing `None` validation.
`node_ip: `

```
fatal: [sles16demo-gmc]: FAILED! =>
    assertion: __sap_maintain_etc_hosts_fact_ip_mismatches_hostnames | length == 0
    changed: false
    evaluated_to: false
    msg: 'FAIL: IP address mismatches found: [{''hostname'': ''host2'', ''defined_ip'':
        None, ''actual_ip'': ''10.10.10.11''}]'
```

marcelmamula, line 126: Missing regex validation for IP that is actually IP.
`node_ip: test`

```
fatal: [sles16demo-gmc]: FAILED! =>
    assertion: __sap_maintain_etc_hosts_fact_ip_mismatches_hostnames | length == 0
    changed: false
    evaluated_to: false
    msg: 'FAIL: IP address mismatches found: [{''hostname'': ''host2'', ''defined_ip'':
        ''test'', ''actual_ip'': ''10.10.10.11''}]'
```

roles/sap_maintain_etc_hosts/tasks/update_host_present.yml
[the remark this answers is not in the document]
marcelmamula: @berndfinger This is not resolution because it is only checking against `None` but missing other validations liek type.
`aliases: test` results in loop over:
- t
- e
- s

```
TASK [sap_maintain_etc_hosts : Hosts File - Presence - Display host, DNS domainname and IP address before the modification] ********************************
 [started TASK: sap_maintain_etc_hosts : Hosts File - Presence - Display host, DNS domainname and IP address before the modification on sles16demo-gmc]
ok: [sles16demo-gmc] =>
    msg:
    - hostname   = 'host2'
    - domain     = 'domain'
    - ip         = '10.10.10.11'
    - comment    = '# managed by Ansible role sap_maintain_etc_hosts'
    - aliases    = 't e s'
    - alias mode = 'overwrite'
```

marcelmamula (changes requested):

berndfinger:
Due to other projects I am working on, with those many more findings and issues to take into account, consider, and/or change the code and re-test, I am expecting that this project will just take too much of my time.
So I am closing this PR now without merging, and leave the role as is for now. Maybe I can find some time later this year, or someone else wants to continue the work?

Issue #1201, opened 2026-03-26
sap_swpm: remove ansible.posix dependency


berndfinger:
The collection `ansible.posix` was used for calling the module `ansible.posix.firewalld` in the files [tasks/pre_install/update_firewall.yml](https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_swpm/tasks/pre_install/update_firewall.yml) and [post_install/update_firewall.yml](https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_swpm/tasks/post_install/update_firewall.yml). Most of the  firewall functionality is now in file [tasks/pre_install/firewall.yml](https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_swpm/tasks/pre_install/firewall.yml) , and the two files named `update_firewall.yml` are on longer referenced.
So we can delete them from the role and also remove the `ansible.posix` entry from the file [roles/sap_swpm/meta/collection-requirements.yml](https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_swpm/meta/collection-requirements.yml).

berndfinger:
Fixed in #1202 .

PR #1202, opened 2026-03-26
sap_swpm: Remove ansible.posix dependency

berndfinger:
Fixes issue #1201.


marcelmamula (approved):
We are not publishing Ansible Roles so that meta file is not used, but removal of old task files makes sense.

LGTM

ja9fuchs (approved):
👍

Issue #1051, opened 2025-06-17
sap_storage_setup: mount point /testnfs2 does not exist


sean-freeman:
### Ansible Role

sap_storage_setup

### OS Family

N/A

### Ansible Controller - Python version

Python 3.13.3

### Ansible-core version

ansible [core 2.16.13]

### Bug Description

Directory for NFS mount may not always exist, and Ansible Role should anticipate this and perform mkdir prior to mount command.

```yaml
TASK [community.sap_install.sap_storage_setup : SAP Storage Setup - (nfs_backup) Attach NFS host root for subdirectory verification/creation] **************************************
task path: sap_install/roles/sap_storage_setup/tasks/generic_tasks/configure_nfs_filesystems.yml:122
fatal: [s4h01]: FAILED! =>
    changed: false
    msg: |-
        Error mounting /tmp/sap_storage_setup_nfs_6ciyo5vd: mount: nfs4: mount point does not exist.
```

### Bug reproduction

Execute Ansible Role with NFS attachment, but do not use existing subdir already created on the OS.

### Community participation

Unfortunately I am not in a position to help with the bug fix

marcelmamula:
@sean-freeman Please check if previous step https://github.com/sap-linuxlab/community.sap_install/blob/8a89687bc58f865f6d77a38d960323345ff2a60f/roles/sap_storage_setup/tasks/generic_tasks/configure_nfs_filesystems.yml#L116 was successful, because it is creating tempfile folder which is used for temporary mount for NFS validation and folder creation.

Allowed mount point names are defined in https://github.com/sap-linuxlab/community.sap_install/blob/8a89687bc58f865f6d77a38d960323345ff2a60f/roles/sap_storage_setup/tasks/generic_tasks/configure_nfs_filesystems.yml#L35 , where only non-SAP mount is `/software` which we use across roles and AP4S.

Mount `/testnfs` or any other non-covered are not part of allowed list.

sean-freeman:
Please excuse my late hour forgetfulness.

OK, this can be considered a feature request then. While containing scope is necessary, permitting generic NFS attachment should be allowed _(without any validation, and with a warning emitted)_ for one key reason - backup files. It falls just outside of the usual scope.

[...]

PR #1205, opened 2026-04-07
sap_storage_setup: Add option to mount generic NFS mounts

marcelmamula:
## Changes
Based on https://github.com/sap-linuxlab/community.sap_install/issues/1051 and our original discussions. @sean-freeman

- Add option to mount generic non-SAP mounts to remove need to add mount steps in playbooks and also to ensure flexibility for mounts like `/lss/shared`.
- Controlled by list of forbidden paths in `__sap_storage_setup_nfs_forbidden_paths`.
- New debug task to show information that non-SAP mount was selected or that mount point is not valid, resulting in skipped temporary task block.
- Fixed `sap_storage_setup_nfs_dirs_usr_sap[type]` that was not sanitized.

## Tests
Tested on SLES_SAP 15 SP7 using various mount points.

**NOTE for testing:** If you are changing `mountpoint` during testing while keeping same `nfs_server`, be catious because `posix.mount` will not actively mount it again, only `/etc/fstab` will get updated.


sean-freeman:
@marcelmamula Looks good, minor three points....

1. `add_software` internal Jinja var rename to `add_other_nfs`
2. `__sap_mount_regex` is missing `/usr/sap/.*` ? Anything under that path should be assumed as SAP? This would also encapsulate `/usr/sap/trans` so that could be removed from the regex string
3. Is there a need for an Ansible Task equivalent to `mount -a / mount --all` at the Ansible Role's conclusion? As a catch all?

For future reference, could go into code comment:
- **SAP System Directories on UNIX** https://help.sap.com/docs/ABAP_PLATFORM_NEW/b5670aaaa2364a29935f40b16499972d/2744f17a26a74a8abfd202c4f5dc9a0f.html?locale=en-US

marcelmamula:
@sean-freeman Thank you for review:
1. I will fix it in coming commit. It was leftover from my previous `/software` implementation.

2. Regex is correct because it will never run for `/usr/sap` as it has `__sap_storage_setup_fact_related_directories > 1`. Logically we dont have to check against it because we know it will never be run in that task.

3. `ansible.posix.mount` is handling `/etc/fstab` as well as mounting already. Only issue is that it does not remount, but that was my blanket testing which will not happen during normal usage.

sean-freeman:
All good 🙂  1 was a nitpick, 2 is my mistake, 3 I agree it was more just a sanity catch suggestion

ja9fuchs (approved):
This looks good, thanks! 👍

PR #1209, opened 2026-04-17
sap_ha_pacemaker_cluster: Add QDevice and Diskless SBD

marcelmamula:
## Premise
QDevice and Diskless SBD are solutions sought for by SAP customers. This PR will add functionality to use both, without dipping into area of using non-cluster host. This is still outside of bounds of this role, as designed.

## Changes
- Add QDevice configuration steps and variables. It is using `ha_cluster_quorum`, without user override or complex logic, because it does not have any other use in `ha_cluster` as of now.
- Split existing STONITH task files to separate custom STONITH from SBD, since Diskless SBD does not work with STONITH agent.
- Fix wrong GCP conditional for health check on sap_hana_secondary.

## Tests
Tested on SLES_SAP 16.1 with ASCS/ERS ENSA2 cluster with both Diskless SBD with and without QDevice.

## NOTE
This change was tested along with https://github.com/linux-system-roles/ha_cluster/pull/374 which enabled this functionality for `crmsh`.

### DIFF
--- a/roles/sap_ha_pacemaker_cluster/tasks/prepare_vars/prepare_stonith_sbd.yml
+++ b/roles/sap_ha_pacemaker_cluster/tasks/prepare_vars/prepare_stonith_sbd.yml
@@ -1,0 +1,223 @@
+# SPDX-License-Identifier: Apache-2.0
+---
+
+# SBD is different to custom fencing resource, because it has dedicated 'sbd' variables and Diskless mode.
+
+# Show warning and disable fencing if user STONITH resources are not defined.
+- name: Block when STONITH list is defined but empty
+  when:
+    - sap_ha_pacemaker_cluster_stonith_custom | d([]) | length == 0
+    # Custom STONITH resource must be empty for Diskless SBD.
+    - not sap_ha_pacemaker_cluster_sbd_devices | d([]) | length == 0
+  block:
+
+    # Ensure that property 'stonith-enabled' is set to 'false' if no Stonith resources are defined.
+    # STONITH property will be disabled only if 'sap_ha_pacemaker_cluster_cluster_properties' is not defined.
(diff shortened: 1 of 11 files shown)


ja9fuchs:
Visibility: I'm checking and testing...

roles/sap_ha_pacemaker_cluster/tasks/prepare_vars/prepare_stonith_sbd.yml
```diff
@@ -0,0 +1,223 @@
...
+          FAIL: The variable 'sap_ha_pacemaker_cluster_sbd_devices' is not defined or empty,
+          which indicates Diskless SBD configuration.
+
+          Diskless SBD requires either of the following:
+          - Watchdog timeout property ('stonith-watchdog-timeout' or 'fencing-watchdog-timeout')
+            configured in the variable 'sap_ha_pacemaker_cluster_cluster_properties'.
+          - QNet host configured in the variable 'sap_ha_pacemaker_cluster_qdevice_host'.
+      when:
+        - not sap_ha_pacemaker_cluster_cluster_properties['stonith-watchdog-timeout'] is defined
+        - not sap_ha_pacemaker_cluster_cluster_properties['fencing-watchdog-timeout'] is defined
```
ja9fuchs: `fencing-watchdog-timeout` was only added in pacemaker 3.0.2. If the option is defined, but run on an older version, the check task passes, but the `ha_cluster` role fails later due to the unsupported property.

`stonith-watchdog-timeout` remains supported in parallel as deprecated. So both property names are valid for 3.0.2+, until dropped from pacemaker.

ja9fuchs (approved):
Looks good now, thanks for covering the differences on RHEL. 👍 

Tested on RHEL 10.0.

Issue #1062, opened 2025-07-01
sap_hostagent: unexpected become removal


crysaki:
### Ansible Role

sap_hostagent

### OS Family

RHEL

### Ansible Controller - Python version

3.9.18

### Ansible-core version

2.14.14

### Bug Description

Since the last release of this collection the collection the `ansible_become: true` is removed from the role.
The installation of the SAP Host Agent has to be done with user root, so at least the task `Install SAPHOSTAGENT` has always to be executed with privilege.

We have the role included in big playbooks with the sar_remote option. Unfortunately the role is not idempotent so it does not determine if an agent is already installed and always extracts the SAR.
So we check if we already have an Agent on the host and do a dynamic include of the role to reduce the runtime.
But since the default become was removed from the role this is not possible anymore (Took us some time to find the change in the role).

So would it possible to add `become` to specific task within the role or do you won't do that, so everybody has to do a privileged static import of the role?

### Bug reproduction

Include the role to a playbook.

### Community participation

Happy to implement this bug fix

berndfinger:
@crysaki Thanks for spending time on bringing this behavior to our attention. I am happy to assist and help solving your issue inside of this collection - instead of allowing unwanted behavior to remain in our code for no good reason (of course, subject to prioritization).

We would like to avoid setting `ansible_become: true` in defaults/main.yml but would rather use `become: true` on the task level. Until then, can you maybe use `import_role` instead of `include_role`? The `ansible.builtin.import_role` module supports setting `become: true` on the task level, and the behavior of that one is similar to calling the role directly with `become: true` as a playbook variable. I tested with ansible-core version 2.18.6, using the following playbook snippet:

```
  tasks:
    - name: Find out if the SAP Host Agent is installed
      ansible.builtin.shell: set -o pipefail && /usr/sap/hostctrl/exe/saphostexec -version | awk 'BEGIN{a=0}/SAPHOSTAGENT information/{a++}END{print a}'
      register: shell_result
      ignore_errors: true

    - name: Display the shell command result
      ansible.builtin.debug:
        msg: |
          "shell_result.stdout: >{{ shell_result.stdout }}<"
          "shell_result.rc: >{{ shell_result.rc }}<"
      ignore_errors: true

    - name: Include a role
#      ansible.builtin.include_role: # won't work with 'become:'
      ansible.builtin.import_role:
        name: sap_hostagent
      become: true
      when:
        - shell_result.rc != 0
[...]

Issue #122, opened 2022-05-05
sap_hostagent: incorrect install status check


myllynen:
sap_hostagent SAR based installation checks for stdout of saphostexec but at least some versions (e.g., SAPHOSTAGENT54_54-80004825.SAR) use stderr, not stdout for installation time messages so the check does not work as intended.

berndfinger:
@myllynen Can you please mention the role name and maybe also the location in the yml file where you found the wrong check?

myllynen:
I came across this with sap_hostagent but haven't checked other roles. See:

https://github.com/sap-linuxlab/community.sap_install/blob/main/roles/sap_hostagent/tasks/deploy_sar.yml#L24

Thanks.

crysaki:
Got also some problems there while installing Hostagent via `community.sap_install.sap_hostagent`.
Installation not finished successful, but the role returned OK. The problem is indeed, that the hostagent installer returns `rc=0` but enteres a text into `stderr`.

```
TASK [community.sap_install.sap_hostagent : Install SAPHOSTAGENT] **********************************************************************************************************************************************************************
task path: /opt/ucc/ansible/collections/ansible_collections/community/sap_install/roles/sap_hostagent/tasks/deploy_sar_remote.yml:13
Tuesday 01 July 2025  06:50:19 +0200 (0:00:06.424)       0:00:50.723 **********
ok: => 
{"changed": false, 
"cmd": ["/tmp/hostagent/saphostexec", "-install"], 
"delta": "0:00:00.676717", 
"end": "2025-07-01 06:50:23.010820", 
"msg": "", 
"rc": 0, 
"start": "2025-07-01 06:50:22.334103", 
"stderr": "Install service\n[ERROR] You have to be a privileged user to execute this operation", 
"stderr_lines": ["Install service", "[ERROR] You have to be a privileged user to execute this operation"], 
"stdout": "", 
"stdout_lines": []}
```

marcelmamula:
@crysaki @myllynen @berndfinger  I know this is very old issue, but it was resolved in https://github.com/sap-linuxlab/community.sap_install/pull/1214. Feel free to test it out and I will close this issue once PR is merged.

PR #1214, opened 2026-05-13
sap_hostagent: Rework and fixes for open issues

marcelmamula:
## Changes
1. Restructure role to create clean separation between types
2. Add detection to skip existing installation, unless `sap_hostagent_overwrite` is set to `true`. This allows also upgrading and downgrading version. Mentioned in https://github.com/sap-linuxlab/community.sap_install/issues/1062
3. Add missing validations and sanitization of variables
4. Add missing validations for presence of files before hard crashing when trying to copy them.
5. Add handling for `non-root` user after over arching become was removed. Issue: https://github.com/sap-linuxlab/community.sap_install/issues/1062
6. Add handling for installation status using also `stderr`, because `saphostexec` does use it instead of `stdout`. Issue: https://github.com/sap-linuxlab/community.sap_install/issues/122
7. Add proper handling for `remote`, instead of type `sar-remote`. Backwards compatibility was added.
8. Remove temp variables and switch to `tempfile`.

## Tests
These changes were tested on:
- SLES 16.0 using all installation types
- RHEL 9.0 using all installation types, including RPM.


roles/sap_hostagent/README.md
```diff
@@ -10,9 +10,15 @@ SAP Host Agent is an agent that can accomplish several life-cycle management tas
...
 
-- SAP SAR file
-- SAP Bundle
-- RPM package (Red Hat only)
+- **SAR file**
+  - Source: SAP Software Center under product `SAP HOST AGENT`.
+
+- **SAP bundle**
+  - Source: SAP Software Center under products like `SAP HANA PLATFORM EDITION`.
+  - These archives contain `HOSTAGENT.TGZ` file, that is used for installation.
```
berndfinger: For better understanding, please change to:
These archives contain the `HOSTAGENT.TGZ` file, which is used for installation.

marcelmamula: Readme was updated

berndfinger (approved):
LGTM.

PR #1213, opened 2026-05-11
sap_swpm: Add Pseudo-Idempotency

marcelmamula:
## Premise
We have been tip toing around this subject for far too long and it is time we fix this. I was discussing this during my presentation at SUSECON26 two weeks ago and I found time to consolidate all points from our discussion https://github.com/sap-linuxlab/community.sap_install/discussions/1005 into working code, that implements **Pseudo-Idempotency**.

### Pseudo-Idempotency definition
Implementation follows logic that I implemented in `sap_hana_install` role already:
1. Detection.
2. Skip destructive tasks, but continue valid post-tasks.
3. Inform users about it.

**This is preventive detection to ensure existing SAP Systems are not destroyed, not fully idempotent code that ensures exact specific SAP Product is installed!**

 ## Changes
1. New `pre_install/detect_sap*.yml` task files added, which is started from within `pre_install/swpm_prepare.yml` in order to follow existing structure of tasks and tags. This file contains whole complex logic for detection and exceptions.

2. New product based dictionaries in `vars/main.yml`. These dictionaries are very complex, but commented in detail to simplify their management.

3. User expiration post-task was split into dedicated tasks in separate file. This ensures that we do not blindly call `chage` and mark it always as change. Note was added about potential use of `ansible.builtin.user`, but it requires `ansible-core 2.18+`.

5. Update debug message blocks to align with rest of changes (remove lines since `pipe` loads it line by line).

6.  Add new override variable `sap_swpm_force: false` and `# sap_swpm_skip_detection: false`.

### Detection Approach
New approach uses detection matrix:
```
### List of scenarios for determining next steps.
# This ensures that installation can be skipped when needed to prevent any potential issues affecting existing installations.
# Optional override is available in defaults/main.yml to execute regardless of warning.
# NOTE: SAP Host Agent can be skipped during installation using 'sap_swpm_install_saphostagent: false'.

## 'saphostctrl' is present:
# 1. SKIP - 'root' saphostctrl check finds the desired SID and instance number.
# 2. INSTALL - 'root' saphostctrl check does not find the desired SID and instance number, and no other indicators are found.
# 3. FAIL - 'root' saphostctrl check does not find the desired SID and instance number, and any other indicators are found.

# 'saphostctrl' is not present:
# 4. SKIP - 'sidadm' sapcontrol check finds the desired SID and instance number.
# 5. INSTALL - 'sidadm' sapcontrol check does not find the desired SID and instance number, and no other indicators are found.
[...]

PR #1215, opened 2026-05-20
sap_hana_install: Update Pseudo-Idempotency

marcelmamula:
## Changes
This pull request is created in alignment with https://github.com/sap-linuxlab/community.sap_install/pull/1213, where strict detection is added. Existing SAP HANA detection was only fairly strict, allowing to see HANA as detected just from files, which is not conclusive.

- Reworked detection
- Added internal variables that were missing (we had unsanitized HANA shared path)

### Detection Approach
New approach uses detection matrix shared with SWPM:
```
### List of scenarios for determining next steps.
# This ensures that installation can be skipped when needed to prevent any potential issues affecting existing installations.
# Optional override is available in defaults/main.yml to execute regardless of warning.
# NOTE: SAP Host Agent can be skipped during installation using 'install_hostagent=n'.

## 'saphostctrl' is present:
# 1. SKIP - 'root' saphostctrl check finds the desired SID and instance number.
# 2. INSTALL - 'root' saphostctrl check does not find the desired SID and instance number, and no other indicators are found.
# 3. FAIL - 'root' saphostctrl check does not find the desired SID and instance number, and any other indicators are found.

# 'saphostctrl' is not present:
# 4. SKIP - 'sidadm' sapcontrol check finds the desired SID and instance number.
# 5. INSTALL - 'sidadm' sapcontrol check does not find the desired SID and instance number, and no other indicators are found.
# 6. FAIL - 'sidadm' sapcontrol check does not find the desired SID and instance number, and any other indicators are found.

# Breakdown of potential scenarios and actions:
# | saphostctrl | sapcontrol | Instance Found | Files Found | Action
# |-------------|------------|----------------|-------------|--------
# | Yes         | N/A        | Yes            | N/A         | SKIP
# | Yes         | N/A        | No             | No          | INSTALL
# | Yes         | N/A        | No             | Any         | FAIL
# | No          | Yes        | Yes            | N/A         | SKIP
# | No          | Yes        | No             | Any         | FAIL
# | No          | No         | No             | No          | INSTALL

# List of non-saphostctrl indicators for potential existing SAP HANA installation:
# A. Files found in /hana/shared/<SID> directory.
# B. Files found in /usr/sap/<SID> directory.
```

## Tests
Tested on SLES_SAP 16.0 on SAP HANA 2.0 SPS08 installation.

### DIFF
--- a/roles/sap_hana_install/README.md
+++ b/roles/sap_hana_install/README.md
@@ -377,7 +377,7 @@ Installs SAP HANA on `host1` and `host2`, while running on host `host0` where ex
 ```yaml
 ---
 - name: Ansible Play for SAP HANA installation - Add hosts
-  hosts: host0, host1
+  hosts: host0, host1, host2
   become: true
   tasks:
     - name: Execute Ansible Role sap_hana_install
--- a/roles/sap_hana_install/defaults/main.yml
+++ b/roles/sap_hana_install/defaults/main.yml
@@ -34,10 +34,9 @@ sap_hana_install_keep_copied_sarfiles: false
 # For installing SAP HANA with fapolicyd support, set the following variable to `true`:
 sap_hana_install_configure_fapolicyd: false
 
-# (RedHat specific) desired fapolicyd service status (only if sap_hana_install_configure_fapolicyd is 'true')
-# For not enabling and not starting the fapolicyd service after the installation has finished, set the following
-# variable to `false`:
-sap_hana_install_enable_fapolicyd: true
+# (RedHat specific) desired fapolicyd service status (only if 'sap_hana_install_configure_fapolicyd' is set to 'true').
+# For enabling and starting the fapolicyd service after the installation has finished, set the following variable to 'true'.
+sap_hana_install_enable_fapolicyd: false
 
 # (RedHat specific) fapolicyd integrity level
 # When using fapolicyd, you can set the following variable to one of `none`, `size`, `sha256`, or `ima`. Note that before setting
--- a/roles/sap_hana_install/defaults/main.yml
+++ b/roles/sap_hana_install/defaults/main.yml
@@ -274,3 +274,7 @@ sap_hana_install_create_initial_tenant: 'y'
 
 # Display SAP HANA hdblcm unattended mode output (hdblcm stdout)
 sap_hana_install_display_unattended_output: false
+
+# (Optional) Set to `true` to skip the filesystem check during SAP HANA installation.
+# This is not recommended, but can be used in case of a non-Production environment.
+sap_hana_install_skip_filesystem_check: false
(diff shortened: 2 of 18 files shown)


berndfinger:
@marcelmamula I found some time for testing and noticed that with fapolicyd running, some Ansible functionality is blocked. So I analyzed the problem and found that it is possible to add Ansible rules to fapolicyd, allowing the role to succeed even with fapolicyd running.

The following code changes on top of your branch achieved the goal in my tests (README.md is not yet updated as there might be follow-up discussions and changes): https://github.com/marcelmamula/community.sap_install/compare/hana-fix...berndfinger:fapolicyd-fix?expand=1 .

The proposed change can deal with an already existing fapolicyd rule file (e.g. with entries for other users) and also with such a rule file not yet present. In both cases, the changes are reverted at the end of the role (but only if the role does not fail in between - there appears to be no mechanism to run tasks in all cases except if we add all tasks to a `try:` block.

Or we could put all fapolicyd prepare and cleanup tasks into separate roles or as separate tasks in the playbook which executes sap_hana_install. But separate roles just for that is probably too unusual and too difficult to explain, and the  latter puts too much burden on the user. Stopping fapolicyd before running the role is not a good alternative either because it opens an otherwise protected system too much.

So maybe an additional warning message related to opening fapolicyd for Ansible is an alternative.

berndfinger:
@marcelmamula I am no longer able to reproduce the issue I reported in https://github.com/sap-linuxlab/community.sap_install/pull/1215#issuecomment-4916517785.

But I found another issue which I would like you to fix as part of this PR, because if users assume that the role has some level of idempotency, they may not be careful enough before running the role against an existing SAP HANA system.

In such cases, on RHEL with `sap_hana_install_configure_fapolicyd` set to true, all the files under `sap_hana_install_directories` will be added to the fapolicyd trust files.

This must be avoided by default, as it may lead to compromised files to be trusted and later be executed.

Can you please modify your code so that the tasks of file `post_tasks/fapolicyd.yml` will only be executed for new HANA installations or for new hosts during addhosts?

marcelmamula:
> @marcelmamula I am no longer able to reproduce the issue I reported in [#1215 (comment)](https://github.com/sap-linuxlab/community.sap_install/pull/1215#issuecomment-4916517785).
> 
> But I found another issue which I would like you to fix as part of this PR, because if users assume that the role has some level of idempotency, they may not be careful enough before running the role against an existing SAP HANA system.
> 
> In such cases, on RHEL with `sap_hana_install_configure_fapolicyd` set to true, all the files under `sap_hana_install_directories` will be added to the fapolicyd trust files.
> 
> This must be avoided by default, as it may lead to compromised files to be trusted and later be executed.
> 
> Can you please modify your code so that the tasks of file `post_tasks/fapolicyd.yml` will only be executed for new HANA installations or for new hosts during addhosts?

@berndfinger I have pushed commit with restricted `fapolicyd` execution as you requested.

I do believe your original issue from https://github.com/sap-linuxlab/community.sap_install/pull/1215#issuecomment-4916517785 still exists, but maybe your host is no longer empty enough to trigger issue.
- Dedicated PR for `fapolicyd` issue you described (your proposal but with improvements I proposed) would still be beneficial when you get time to setup some test environment to have consistent results.

berndfinger (approved):
LGTM!

PR #1217, opened 2026-05-25
collection: Update PR branch policy workflow

marcelmamula:
## Changes
This is update to workflow added in https://github.com/sap-linuxlab/community.sap_install/pull/1192
- Add additional check to block PR with branch from main instead of dev
- Update formatting, syntax and comments to align with other workflows
- Update workflow name to be less ambiguous

## Next Steps
After this change is merged, I will do same for other repositories in project, with big focus on `sap_libs`, `sap_operations` and `sap_infrastructure` where I am having to close pull requests because of these reasons.

## Tests
Tested in my private repository, but real test will have to happen in here.


berndfinger (approved):
Looks like a useful addition, let's try it! LGTM.

Issue #1219, opened 2026-05-28
feat: sap_ha_install_hana_hsr - configure dedicated IP for HSR


ja9fuchs:
### Ansible Role

sap_ha_install_hana_hsr

### OS Family

All

### Feature/Enhancement request

Currently the HSR role does not use the hosts' `node_ip` definition for a configuration of the system replication on a different interface, but only for some tasks during the automated setup.

**Request**
Add support for configuring the HSR on a different IP/interface than the hosts' default.

### Community participation

Happy to implement this feature

marcelmamula:
PR was merged.

Issue #357, opened 2023-04-18
sap_hana_install: multihost scale-out listeninterface


sean-freeman:
When multihost/ scale-out, at minimum the following are required in `global.ini`:

```
[system_replication_communication]
listeninterface = .global
```

For reference, see [SAP Note 2183363 - Configuration of SAP HANA internal network](https://launchpad.support.sap.com/#/notes/2183363

marcelmamula:
Hi @sean-freeman . I had similar issue when dealing with implentation of https://github.com/sap-linuxlab/community.sap_install/pull/1126 / https://github.com/sap-linuxlab/community.sap_install/pull/1135 and issue was related to Firewall, not this parameter.

As per note you linked, this is handled by `addhost` workflow of `hdblcm` and should be tuned when having multiple interfaces. This is out of scope of current role as this configuration is not prefaced with variables or customized.

PR #1224, opened 2026-07-16
sap_ha_install_hana_hsr: Add dedicated replication network support

ja9fuchs:
**NEW: support explicit HSR interface**
Configure `[system_replication_hostname_resolution]` and `[system_replication_communication]` `listeninterface` in `global.ini` to route HSR traffic over a dedicated network interface.

New variables:
- `node_hsr_ip` (per node in cluster_nodes)
- `sap_ha_install_hana_hsr_listeninterface` (.internal/.global)

Solves issue #1219.

**FIX: Refactor `set_log_mode`**
- Use SQL query (M_INIFILE_CONTENTS) instead of grep for idempotency check.
- Fix conditional that always evaluated true due to integer/string comparison (rc != '0'). 
- Drop redundant ALTER HOST layer (silently updates host specific global.ini), set SYSTEM layer (default global.ini) only.

**Pseudo-Idempotency**
All `global.ini` settings run check tasks first, using SQL checks that filter LAYER_NAME = 'SYSTEM' to avoid false positives from DEFAULT or HOST layer values.
The checks are exact matches. If the values change, it overrides the existing parameter with the new value.
If there is already a HANA system replication active between the instances, the setup is skipped. (not a new change)

**Tested**

Preparation: dedicated interface/IP statically configured on each node

| Test | Input | Expected | Result |
| :---- | :---- | :---- | :---- |
| no new vars | baseline | unchanged behavior | PASS |
| node_hsr_ip | both nodes | hostname_resolution set | PASS |
| node_hsr_ip | only on 1 of 2 nodes | validation error | PASS |
| node_hsr_ip | IP not on interface | validation error | PASS |
| listeninterface | `.global`, no node_hsr_ip | listeninterface only | PASS |
| listeninterface | empty | not configured | PASS |
| listeninterface | `foo` | validation error | PASS |
| listeninterface | `.internal`, no IPs | validation error | PASS |
| node_hsr_ip + listeninterface | both nodes + `.global` | both configured | PASS |
| node_hsr_ip + listeninterface | both nodes + `.internal` | both configured | PASS |

### DIFF
--- a/roles/sap_ha_install_hana_hsr/defaults/main.yml
+++ b/roles/sap_ha_install_hana_hsr/defaults/main.yml
@@ -42,7 +42,7 @@ sap_ha_install_hana_hsr_instance_number: ''
 #     node_hsr_ip: "192.168.10.2"
 sap_ha_install_hana_hsr_cluster_nodes: []
 
-# The hdbuserstore key used for connecting to the SYSTEMDB for administrative tasks like backups (String).
+# The hdbuserstore key used for connecting to the SYSTEMDB for administrative tasks like backups or updating the global.ini (String).
 sap_ha_install_hana_hsr_hdbuserstore_system_backup_user: HDB_SYSTEMDB
 
 # Password for the HANA 'SYSTEM' user (String).
@@ -112,8 +112,8 @@ sap_ha_install_hana_hsr_configure_firewall: false
 # Default firewall zone, usually `public`, is used if this variable is undefined.
 # sap_ha_install_hana_hsr_firewall_zone: ''
 
-# SAP HANA System Replication listen interface (String).
+# Optional: SAP HANA System Replication listen interface (String).
 # Controls which network interface HANA listens on for incoming replication connections.
 # Available values: '.internal', '.global'.
 # When not set, listeninterface is not configured and HANA uses its default.
-sap_ha_install_hana_hsr_listeninterface: ''
+# sap_ha_install_hana_hsr_listeninterface: ''
--- a/roles/sap_ha_install_hana_hsr/defaults/main.yml
+++ b/roles/sap_ha_install_hana_hsr/defaults/main.yml
@@ -23,16 +23,23 @@ sap_ha_install_hana_hsr_instance_number: ''
 #   node_role: (string) Role of the node in the HSR setup. Valid options: 'primary', 'secondary'.
 #   hana_site: (string) A unique logical site name (e.g., DC01, EU-WEST).
 #
+# Optional key for dedicated replication network:
+#   node_hsr_ip: (string) IP address of the dedicated replication interface.
+#     When set, configures [system_replication_hostname_resolution] in global.ini
+#     to map each peer's hostname to its replication IP.
+#
 # Example:
 # sap_ha_install_hana_hsr_cluster_nodes:
 #   - node_name: "hana01"
 #     node_ip: "10.0.0.1"
 #     node_role: primary
 #     hana_site: DC01
+#     node_hsr_ip: "192.168.10.1"
 #   - node_name: "hana02"
 #     node_ip: "10.0.0.2"
 #     node_role: secondary
 #     hana_site: DC02
+#     node_hsr_ip: "192.168.10.2"
 sap_ha_install_hana_hsr_cluster_nodes: []
 
 # The hdbuserstore key used for connecting to the SYSTEMDB for administrative tasks like backups (String).
@@ -104,3 +111,9 @@ sap_ha_install_hana_hsr_configure_firewall: false
 # Optional name of firewall zone where service or ports will be configured.
 # Default firewall zone, usually `public`, is used if this variable is undefined.
 # sap_ha_install_hana_hsr_firewall_zone: ''
+
+# SAP HANA System Replication listen interface (String).
+# Controls which network interface HANA listens on for incoming replication connections.
+# Available values: '.internal', '.global'.
+# When not set, listeninterface is not configured and HANA uses its default.
+sap_ha_install_hana_hsr_listeninterface: ''
--- a/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
+++ b/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
@@ -1,0 +1,84 @@
+# SPDX-License-Identifier: Apache-2.0
+---
+
+- name: SAP HSR - Check current system_replication_hostname_resolution in global.ini
+  become: true
+  become_user: "{{ __sap_ha_install_hana_hsr_sid | lower }}adm"
+  ansible.builtin.shell:
+    cmd: |
+      /usr/sap/{{ __sap_ha_install_hana_hsr_sid }}/SYS/exe/hdb/hdbsql \
+      -U {{ sap_ha_install_hana_hsr_hdbuserstore_system_backup_user }} \
+      -m <<EOF
+      SELECT KEY, VALUE FROM M_INIFILE_CONTENTS
+        WHERE FILE_NAME = 'global.ini'
+        AND LAYER_NAME = 'SYSTEM'
+        AND SECTION = 'system_replication_hostname_resolution'
+        AND KEY = '{{ node_item.node_hsr_ip }}'
+        AND VALUE = '{{ node_item.node_name }}';
+      EOF
+  loop: "{{ sap_ha_install_hana_hsr_cluster_nodes }}"
+  loop_control:
+    label: "{{ node_item.node_hsr_ip | default('undefined') }} = {{ node_item.node_name }}"
+    loop_var: node_item
+  register: __sap_ha_install_hana_hsr_register_hostname_resolution
+  changed_when: false
+  failed_when: false
+  when: __sap_ha_install_hana_hsr_fact_replication_network
+
+- name: SAP HSR - Configure system_replication_hostname_resolution in global.ini
+  become: true
+  become_user: "{{ __sap_ha_install_hana_hsr_sid | lower }}adm"
+  ansible.builtin.shell:
+    cmd: |
+      /usr/sap/{{ __sap_ha_install_hana_hsr_sid }}/SYS/exe/hdb/hdbsql \
+      -U {{ sap_ha_install_hana_hsr_hdbuserstore_system_backup_user }} \
+      -m <<EOF
+      ALTER SYSTEM ALTER CONFIGURATION('global.ini','SYSTEM')
+        SET ('system_replication_hostname_resolution','{{ node_item.node_item.node_hsr_ip }}')
+        = '{{ node_item.node_item.node_name }}' WITH RECONFIGURE;
+      EOF
+  loop: "{{ __sap_ha_install_hana_hsr_register_hostname_resolution.results }}"
+  loop_control:
+    label: "{{ node_item.node_item.node_hsr_ip | default('undefined') }} = {{ node_item.node_item.node_name }}"
+    loop_var: node_item
+  changed_when: true
+  when:
+    - __sap_ha_install_hana_hsr_fact_replication_network
+    - node_item.node_item.node_name not in node_item.stdout
+
+- name: SAP HSR - Check current listeninterface in global.ini
+  become: true
+  become_user: "{{ __sap_ha_install_hana_hsr_sid | lower }}adm"
+  ansible.builtin.shell:
+    cmd: |
+      /usr/sap/{{ __sap_ha_install_hana_hsr_sid }}/SYS/exe/hdb/hdbsql \
+      -U {{ sap_ha_install_hana_hsr_hdbuserstore_system_backup_user }} \
+      -m <<EOF
+      SELECT VALUE FROM M_INIFILE_CONTENTS
+        WHERE FILE_NAME = 'global.ini'
+        AND LAYER_NAME = 'SYSTEM'
+        AND SECTION = 'system_replication_communication'
+        AND KEY = 'listeninterface'
+        AND VALUE = '{{ sap_ha_install_hana_hsr_listeninterface }}';
+      EOF
+  register: __sap_ha_install_hana_hsr_register_listeninterface
+  changed_when: false
+  failed_when: false
+  when: sap_ha_install_hana_hsr_listeninterface | d('') | length > 0
+
+- name: SAP HSR - Configure listeninterface in global.ini
+  become: true
+  become_user: "{{ __sap_ha_install_hana_hsr_sid | lower }}adm"
+  ansible.builtin.shell:
+    cmd: |
+      /usr/sap/{{ __sap_ha_install_hana_hsr_sid }}/SYS/exe/hdb/hdbsql \
+      -U {{ sap_ha_install_hana_hsr_hdbuserstore_system_backup_user }} \
+      -m <<EOF
+      ALTER SYSTEM ALTER CONFIGURATION('global.ini','SYSTEM')
+        SET ('system_replication_communication','listeninterface')
+        = '{{ sap_ha_install_hana_hsr_listeninterface }}' WITH RECONFIGURE;
+      EOF
+  changed_when: true
+  when:
+    - sap_ha_install_hana_hsr_listeninterface | d('') | length > 0
+    - sap_ha_install_hana_hsr_listeninterface not in __sap_ha_install_hana_hsr_register_listeninterface.stdout
--- a/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
+++ b/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
@@ -18,7 +18,7 @@
       EOF
   loop: "{{ sap_ha_install_hana_hsr_cluster_nodes }}"
   loop_control:
-    label: "{{ node_item.node_hsr_ip | default('undefined') }} = {{ node_item.node_name }}"
+    label: "{{ node_item.node_hsr_ip | d('undefined') }} = {{ node_item.node_name }}"
     loop_var: node_item
   register: __sap_ha_install_hana_hsr_register_hostname_resolution
   changed_when: false
@@ -39,7 +39,7 @@
       EOF
   loop: "{{ __sap_ha_install_hana_hsr_register_hostname_resolution.results }}"
   loop_control:
-    label: "{{ node_item.node_item.node_hsr_ip | default('undefined') }} = {{ node_item.node_item.node_name }}"
+    label: "{{ node_item.node_item.node_hsr_ip | d('undefined') }} = {{ node_item.node_item.node_name }}"
     loop_var: node_item
   changed_when: true
   when:
--- a/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
+++ b/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 ---
+# Configure replication network settings at SYSTEM layer only. HOST layer overrides are not managed by this role.
 
 - name: SAP HSR - Check current system_replication_hostname_resolution in global.ini
   become: true
--- a/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
+++ b/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
@@ -281,7 +281,7 @@
     that:
       - node_item.node_hsr_ip is defined
       - node_item.node_hsr_ip is string
-      - node_item.node_hsr_ip | string | length > 0
+      - node_item.node_hsr_ip | trim | string | length > 0
     success_msg: |
       PASS: Node '{{ node_item.node_name }}' has 'node_hsr_ip' defined.
     fail_msg: |
--- a/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
+++ b/roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
@@ -23,6 +23,7 @@
         FAIL: The variable 'sap_ha_install_hana_hsr_state' is not one of valid values.
         Available values: present, absent
       {% endif %}
+    quiet: true
 
 
 # SID
@@ -48,6 +49,7 @@
       {% else %}
         FAIL: The length of SAP HANA System ID '{{ __sap_ha_install_hana_hsr_sid }}' is not 3 characters!
       {% endif %}
+    quiet: true
 
 
 # Instance Number
@@ -74,6 +76,7 @@
       {% else %}
         FAIL: The SAP HANA Instance Number '{{ __sap_ha_install_hana_hsr_instance_number }}' is not 2 digits!
       {% endif %}
+    quiet: true
 
 
 # Database password
@@ -101,6 +104,7 @@
       {% else %}
         FAIL: The variable 'sap_ha_install_hana_hsr_db_system_password' is empty.
       {% endif %}
+    quiet: true
 
 
 # Domain
@@ -125,6 +129,7 @@
       {% else %}
         FAIL: The variable 'sap_ha_install_hana_hsr_domain' is empty.
       {% endif %}
+    quiet: true
 
 
 # Replication mode
@@ -145,6 +150,7 @@
         FAIL: The variable 'sap_ha_install_hana_hsr_rep_mode' is not one of valid values.
         Available values: sync, syncmem, async
       {% endif %}
+    quiet: true
 
 
 # Operation mode
@@ -165,6 +171,7 @@
         FAIL: The variable 'sap_ha_install_hana_hsr_oper_mode' is not one of valid values.
         Available values: delta_datashipping, logreplay, logreplay_readaccess
       {% endif %}
+    quiet: true
 
 
 # Backup user
@@ -181,6 +188,7 @@
       {% else %}
         FAIL: The variable 'sap_ha_install_hana_hsr_hdbuserstore_system_backup_user' is empty.
       {% endif %}
+    quiet: true
 
 
 # $HOME path
@@ -235,6 +243,7 @@
       {% else %}
       FAIL: Duplicate 'node_name' found in 'sap_ha_install_hana_hsr_cluster_nodes'. Each node must have a unique 'node_name'.
       {% endif %}
+    quiet: true
 
 - name: SAP HSR - Assert that each node in 'sap_ha_install_hana_hsr_cluster_nodes' has required attributes
   ansible.builtin.assert:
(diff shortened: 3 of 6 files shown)


roles/sap_ha_install_hana_hsr/defaults/main.yml
```diff
@@ -104,3 +111,9 @@ sap_ha_install_hana_hsr_configure_firewall: false
 # Optional name of firewall zone where service or ports will be configured.
 # Default firewall zone, usually `public`, is used if this variable is undefined.
 # sap_ha_install_hana_hsr_firewall_zone: ''
+
+# SAP HANA System Replication listen interface (String).
+# Controls which network interface HANA listens on for incoming replication connections.
+# Available values: '.internal', '.global'.
+# When not set, listeninterface is not configured and HANA uses its default.
```
marcelmamula, line 118: Since you are not hiding variable as we do with other optional vars, then it should say:
```yaml
# When undefined or empty string, listeninterface is not configured and HANA uses its default.
```

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
```diff
@@ -0,0 +1,84 @@
+# SPDX-License-Identifier: Apache-2.0
+---
+
+- name: SAP HSR - Check current system_replication_hostname_resolution in global.ini
+  become: true
+  become_user: "{{ __sap_ha_install_hana_hsr_sid | lower }}adm"
+  ansible.builtin.shell:
+    cmd: |
+      /usr/sap/{{ __sap_ha_install_hana_hsr_sid }}/SYS/exe/hdb/hdbsql \
+      -U {{ sap_ha_install_hana_hsr_hdbuserstore_system_backup_user }} \
```
marcelmamula, line 11: Do we require that backup user `sap_ha_install_hana_hsr_hdbuserstore_system_backup_user` has SYSTEM privileges to change global.ini parameters?

Previously it was used to create backup and set log_mode so it would make sense to add comment to defaults and README mentioning how it is used: backup, log_mode, global.ini update

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
```diff
@@ -270,6 +270,52 @@
...
+      "{{ __sap_ha_install_hana_hsr_cluster_nodes | selectattr('node_hsr_ip', 'defined') | list | length > 0 }}"
+
+- name: SAP HSR - Assert that all nodes define 'node_hsr_ip' when any node does
+  ansible.builtin.assert:
+    that:
+      - node_item.node_hsr_ip is defined
+      - node_item.node_hsr_ip | string | length > 0
+    success_msg: |
+      PASS: Node '{{ node_item.node_name }}' has 'node_hsr_ip' defined.
+    fail_msg: |
```
marcelmamula, line 288: Please improve this fail message, because if value is non-string, it should not say missing and it should not check string length before knowing it is string.

```yaml
- name: SAP HSR - Assert that all nodes define 'node_hsr_ip' when any node does
  ansible.builtin.assert:
    that:
      - node_item.node_hsr_ip is defined
      - node_item.node_hsr_ip is string      
      - node_item.node_hsr_ip | trim | string | length > 0
    fail_msg: |
      FAIL: Node '{{ node_item.node_name }}' parameter 'node_hsr_ip' is undefined, empty, or invalid.
      When using a dedicated replication network, all nodes must define 'node_hsr_ip' parameter.
```

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
```diff
@@ -270,6 +270,52 @@
...
+      - sap_ha_install_hana_hsr_listeninterface in ['.internal', '.global']
+    success_msg: |
+      PASS: The variable 'sap_ha_install_hana_hsr_listeninterface' is '{{ sap_ha_install_hana_hsr_listeninterface }}'.
+    fail_msg: |
+      FAIL: The variable 'sap_ha_install_hana_hsr_listeninterface' must be '.internal' or '.global'.
+  when: sap_ha_install_hana_hsr_listeninterface | d('') | length > 0
+
+- name: SAP HSR - Assert that 'node_hsr_ip' is defined when listeninterface is '.internal'
+  ansible.builtin.assert:
+    that:
```
marcelmamula, line 313: We need to stop asserting against internal variable `__var` because it will be exposed in ansible error output.

## **We have to fix this in ALL roles, so it would be better to not add it in new asserts.**
@berndfinger FYI

I have been replacing it with fail module, where message is clear and `when` conditional can use internal variable.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
```diff
@@ -280,6 +326,7 @@
       node_ip: "{{ node_item.node_ip }}"
       node_role: "{{ node_item.node_role }}"
       hana_site: "{{ node_item.hana_site }}"
+      node_hsr_ip: "{{ node_item.node_hsr_ip | default('') }}"
```
marcelmamula: Role is already primarily using `d()` instead of `default()` so better synchronize it with rest.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
```diff
@@ -316,6 +363,16 @@
 
       If this host is not supposed to be part of the HSR cluster, please remove it from hosts in Ansible Play.
 
+- name: SAP HSR - Assert that 'node_hsr_ip' is present on this host
+  ansible.builtin.assert:
+    that:
+      - __sap_ha_install_hana_hsr_connection.node_hsr_ip in ansible_facts['all_ipv4_addresses']
```
marcelmamula, line 372: Same comment with assert against internal variable.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/set_log_mode.yml
```diff
@@ -22,9 +30,10 @@
       /usr/sap/{{ __sap_ha_install_hana_hsr_sid }}/SYS/exe/hdb/hdbsql \
       -U {{ sap_ha_install_hana_hsr_hdbuserstore_system_backup_user }} \
       -m <<EOF
-      ALTER SYSTEM ALTER CONFIGURATION ('global.ini', 'SYSTEM') SET ('persistence', 'log_mode') = 'normal' WITH RECONFIGURE;
-      ALTER SYSTEM ALTER CONFIGURATION('global.ini','HOST','{{ ansible_facts['hostname'] }}') SET ('persistence','log_mode') = 'normal' WITH RECONFIGURE;
+      ALTER SYSTEM ALTER CONFIGURATION('global.ini','SYSTEM')
```
marcelmamula, line 34: Why was second command removed?

marcelmamula (changes requested):
Hi @ja9fuchs This is just initial review, but I am unable to test until mid next week, when I am back.

roles/sap_ha_install_hana_hsr/defaults/main.yml
ja9fuchs: By hiding you mean commenting it out in the defaults? I could do that, default '' is applied in the tasks for conditionals if undefined.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
ja9fuchs: Yeah, the parameter name is not very accurate. I can add the mention to the README and defaults.

roles/sap_ha_install_hana_hsr/defaults/main.yml
marcelmamula: It is all about wording, because if you default to empty string then it is not `unset` which invalidates rest of sentence `listeninterface is not configured and HANA uses its default.`

Which means optional vars should be either:
- Commented out and marked as optional
- Kept as empty of correct data type, but comment explains that undefined or empty value is not processed or used.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: It says "invalid", not missing.

Why should it check for string length before checking if it's a string? When it's not a string, the length does not matter.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: Ah, I will look into that.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
marcelmamula: First data type: Ansible is Lazy by default (less since 2.18, but still lazy).
- If you use `| string` before length, then it can assume `list` is also valid because it goes through lazy filter.
- This is why data type check is mandatory.

Second message: `FAIL: Node '{{ node_item.node_name }}' is missing 'node_hsr_ip'.`
- You will get it when value is '', so it is not missing but empty.
- There should always be only 3 fail states:
  - undefined
  - wrong data type
  - correct data type but empty

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/set_log_mode.yml
ja9fuchs: Because it creates a duplicate entry in a host specific file in a different path. The role never checks that file. For a new HSR setup, we configure a standard system-wide global.ini and not host-specific settings.

System: `/hana/shared/<SID>/global/hdb/custom/config/global.ini`
Host: `/hana/shared/<SID>/HDB<NN>/<host>/global.ini`

Host-specific overrides system-wide global.ini. It silently adds unwanted host settings if we configure both.

Something that was in the original role code and inherited. I noticed when I debugged the broken log_mode check.

marcelmamula: Good, but add it as explanation above task to document why it was removed and why it makes sense.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
marcelmamula: Since I have been applying this to new upcoming role, I dont have existing code to show for it.
Just remember that switch to `fail` module requires `anti-test`.

```yaml
when:
  - sap_ha_install_hana_hsr_listeninterface | d('') == '.internal'
  - not __sap_ha_install_hana_hsr_fact_replication_network | d(false)
```

sean-freeman:
For reference purposes, also relates back to #357

roles/sap_ha_install_hana_hsr/defaults/main.yml
ja9fuchs: Commented out.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
ja9fuchs: Added the global.ini update info.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: Ah, my bad! I confused your suggestion with the actual code.

Fixed it.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/set_log_mode.yml
ja9fuchs: What do you mean? Documenting above that task that something was removed??
This is visible in the git commit text.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
marcelmamula: Please add `trim` I mentioned. Empty `' '` will pass your assert.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/set_log_mode.yml
marcelmamula: Someone working on code will not go and open git log to see changes.
Add explanation above task that command to update host was removed, because what you said before.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: I don't see how the internal variable name can confuse users.

The FAIL message is very clear about what is wrong:

```
fatal: [rhel100-h2]: FAILED! =>
    assertion: __sap_ha_install_hana_hsr_fact_replication_network
    changed: false
    evaluated_to: false
    msg: |-
        FAIL: 'sap_ha_install_hana_hsr_listeninterface' is set to '.internal' but no 'node_hsr_ip' is defined.
        HANA needs [system_replication_hostname_resolution] entries to bind to the internal interface.
        Define 'node_hsr_ip' on all nodes in 'sap_ha_install_hana_hsr_cluster_nodes'.
```

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: Entries updated for consistency. Including some that were not added in this PR.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
marcelmamula: You also have missing space there `is undefined,empty, or invalid.`

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: Same as above. The FAIL message makes it clear.

```
fatal: [rhel100-h2]: FAILED! =>
    assertion: __sap_ha_install_hana_hsr_connection.node_hsr_ip in ansible_facts['all_ipv4_addresses']
    changed: false
    evaluated_to: false
    msg: |-
        FAIL: The 'node_hsr_ip' (10.99.50.196)
        is not configured on host 'rhel100-h2'.
        The replication IP must be present on a network interface before configuring HSR.
```

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
marcelmamula: Users never defined `__sap_ha_install_hana_hsr_fact_replication_network` and they do not know that this is you counting how many nodes have parameter `node_hsr_ip` defined.

This is problem with internal vars - they are transformed and do not reflect what user defined.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/set_log_mode.yml
ja9fuchs: I have added comments at the top of both files that modify the global.ini via SQL:

```
# Set log_mode at SYSTEM layer only. HOST layer overrides are not managed by this role.
```

and

```
# Configure replication network settings at SYSTEM layer only. HOST layer overrides are not managed by this role.
```

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: Added trim and fixed the typo earlier already. :)

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
```diff
@@ -270,16 +270,64 @@
       {% endif %}
 
 
+# Replication network
+- name: SAP HSR - Set fact for replication network availability
+  ansible.builtin.set_fact:
+    __sap_ha_install_hana_hsr_fact_replication_network:
+      "{{ __sap_ha_install_hana_hsr_cluster_nodes | selectattr('node_hsr_ip', 'defined') | list | length > 0 }}"
+
+- name: SAP HSR - Assert that all nodes define 'node_hsr_ip' when any node does
```
marcelmamula, line 280: QoL: Add `quiet: true` because default output is too much as it shows content of all items.

Current:
```bash
ok: [h02hana1] => (item=h02hana0) =>
    ansible_loop_var: node_item
    changed: false
    msg: |-
        PASS: Node 'h02hana0' has 'node_hsr_ip' defined.
    node_item:
        hana_site: DC01
        node_hsr_ip: 10.10.10.50
        node_ip: 10.10.10.178
        node_name: h02hana0
        node_role: primary
failed: [h02hana1] (item=h02hana1) =>
    ansible_loop_var: node_item
    assertion: node_item.node_hsr_ip is defined
    changed: false
    evaluated_to: false
    msg: |-
        FAIL: Node 'h02hana1' parameter 'node_hsr_ip' is undefined, empty, or invalid.
        When using a dedicated replication network, all nodes must define 'node_hsr_ip'.
    node_item:
        hana_site: DC02
        node_ip: 10.10.10.221
        node_name: h02hana1
        node_role: secondary
```

With quiet:
```bash
ok: [h02hana1] => (item=h02hana0)
failed: [h02hana1] (item=h02hana1) =>
    ansible_loop_var: node_item
    assertion: node_item.node_hsr_ip is defined
    changed: false
    evaluated_to: false
    msg: |-
        FAIL: Node 'h02hana1' parameter 'node_hsr_ip' is undefined, empty, or invalid.
        When using a dedicated replication network, all nodes must define 'node_hsr_ip'.
    node_item:
        hana_site: DC02
        node_ip: 10.10.10.221
        node_name: h02hana1
        node_role: secondary
```

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
```diff
@@ -0,0 +1,85 @@
+# SPDX-License-Identifier: Apache-2.0
+---
+# Configure replication network settings at SYSTEM layer only. HOST layer overrides are not managed by this role.
+
```
marcelmamula, line 4: Add new task to handle IP validation. We should not be force feeding IP address into HANA configuration without knowing that it exists.

Simple check for `node_hsr_ip` in `ansible_facts['all_ipv4_addresses']` is enough (Loop + host validation as this is host specific fact).

marcelmamula (changes requested):
Test went fine and I added 2 more points for improvement.
```bash
"10.10.10.51","h02hana1"
"log_mode","normal"
"10.10.10.50","h02hana0"
"backup_encryption","on"
"listeninterface",".global"
```

Regarding idempotency:
- Role is pseudo-idempotent, because it skips after detection.
- This PR is also not idempotent, because all these `hdbsql` updates are  skipped. It would make sense to run proper pass on idempotency in separate PR to split what tasks can be run and cannot run.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
berndfinger: @marcelmamula Regarding https://github.com/sap-linuxlab/community.sap_install/pull/1224#discussion_r3594126104 , in which you stated: "We have to fix this in ALL roles, so it would be better to not add it in new asserts."

I am against changing all the `__var` asserts in all roles, because the `success_msg` and `fail_msg` already explains (at least in most cases) why the assertion passed or failed, just as Janine explained above. From my point of view, the fact that "internal" vars (they are not really internal) are part of the assert output (in the line starting with `assertion:`) is not confusing because it's the content of the `msg:` section which explains what has failed and offers suggestions on what should be changed in order to let the assertion pass.

Changing the asserts would result in additional coding and testing effort without adding real value. It's the contrary: Having to use another Ansible module (e.g. `fail`) could lead to less readable code. For example, in many cases we want to have an extra `success_msg`, and for that, we would need an extra task.

marcelmamula: @berndfinger I already conceded for this particular case, but I will be doing `fail` module for any further changes I do, if internal fact is used, while using `assert` as usual when variable is used.

This comment can be closed.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: I added `quiet` to all assert tasks on the file that contain `success_msg`.

Please note that out of the 14 changed assert tasks with the verbose "issue" only 2 were newly added in this PR.

marcelmamula: @ja9fuchs I have raised this specifically for this assert because you have combination of 2 causes:
1. Looping item which is dictionary.
2. Success message.

This generates big output message.

You did not need to add it to previous asserts or simple loops, because issue is not with them.
> This is now invalidating `success_msg` as it is quieted down.

ja9fuchs:
> Test went fine and I added 2 more points for improvement.
> 
> ```shell
> "10.10.10.51","h02hana1"
> "log_mode","normal"
> "10.10.10.50","h02hana0"
> "backup_encryption","on"
> "listeninterface",".global"
> ```
> 
> Regarding idempotency:
> 
> * Role is pseudo-idempotent, because it skips after detection.
> * This PR is also not idempotent, because all these `hdbsql` updates are  skipped. It would make sense to run proper pass on idempotency in separate PR to split what tasks can be run and cannot run.

Well, if a working system replication is detected on the running HANA systems, it's not a great idea to make any changes in the related configuration. Hence skipping.

Edit: Ah you mean that in the case when there is no SR detected to be running, the tasks should overwrite existing SR configuration?
Yes, that should be a separate enhancement.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/configure_hsr_network.yml
ja9fuchs: Um, yes... `prepare_variables` lines 381ff:

```
- name: SAP HSR - Assert that 'node_hsr_ip' is present on this host
  ansible.builtin.assert:
    that:
      - __sap_ha_install_hana_hsr_connection.node_hsr_ip in ansible_facts['all_ipv4_addresses']
    fail_msg: |
      FAIL: The 'node_hsr_ip' ({{ __sap_ha_install_hana_hsr_connection.node_hsr_ip }})
      is not configured on host '{{ __sap_ha_install_hana_hsr_connection.node_name }}'.
      The replication IP must be present on a network interface before configuring HSR.
  when: __sap_ha_install_hana_hsr_fact_replication_network
```

marcelmamula:
@ja9fuchs My idempotency comment was about updating PR description to make sure that you are not using word `idempotency`, but rather `pseudo-idempotency` to ensure that there is no misconception.

When you say once that it is idempotent, then someone might question why this PR does not change global.ini in existing HSR setup.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: Well other assert tasks loop as well and produce quite a lot of verbose output.

The `success_msg` is still visible in verbose mode, when `quiet` is applied.

At least the task `SAP HSR - Assert that each node in 'sap_ha_install_hana_hsr_cluster_nodes' has required attributes` produces the same amount of output. 

So just a change to this one task? But what formal rule do you want to apply to make this some standard that is applied to all roles, not just this PR and one task?

ja9fuchs:
@marcelmamula I updated the PR description with regards to the skipping of a running SR. 
The actual hdbnsql query and update tasks apply exact matches and adjust when the existing parameter is different. This is proper idempotency, or not?

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
marcelmamula: @ja9fuchs This is just about readability and being sensible when deciding how much is too much. When this task failed during my test (intended fail) it threw 4x10 lines of text at me and I had to carefully go through them to find issue.

There is no need to make this formal requirement or project guidance as this is about sensible output.
Few examples:
1. `name: SAP HSR - Assert that the variable 'sap_ha_install_hana_hsr_sid' is defined as String consisting of 3 characters`
   - This is string variable without loop so it will never pollute output, because only `success_msg` will be shown.
2. `name: SAP HSR - Assert that each node in 'sap_ha_install_hana_hsr_cluster_nodes' has required attributes`
   - This is loop over list of dictionaries so it will show whole content of dictionary (list item), resulting in large output.

As you can see, problem is with loops and dictionaries just make it worse.

marcelmamula:
> @marcelmamula I updated the PR description with regards to the skipping of a running SR. The actual hdbnsql query and update tasks apply exact matches and adjust when the existing parameter is different. This is proper idempotency, or not?

New task file that you added is still inside of block with `not __sap_ha_install_hana_hsr_fact_replication_enabled`, so it will never execute if HSR is already detected. This means this is not idempotency and it will never change it as long as that fact says HSR is present.

That is definition of pseudo-idempotency.

ja9fuchs:
> New task file that you added is still inside of block with `not __sap_ha_install_hana_hsr_fact_replication_enabled`, so it will never execute if HSR is already detected. This means this is not idempotency and it will never change it as long as that fact says HSR is present.
> 
> That is definition of pseudo-idempotency.

Check `tasks/pre_tasks/get_replication_status.yml`. There are checks if the existing SR is the same as what the role wants to configure and if anything is misaligned. It does skip the setup either way, because an existing, running HSR needs to be deconfigured before it can be reconfigured. This is not in scope of the current role.

roles/sap_ha_install_hana_hsr/tasks/pre_tasks/prepare_variables.yml
ja9fuchs: I reverted the previous commit and the new one only adds `quiet` to the 2 similar tasks that loop over the dict list.

marcelmamula:
> Check `tasks/pre_tasks/get_replication_status.yml`. There are checks if the existing SR is the same as what the role wants to configure and if anything is misaligned. It does skip the setup either way, because an existing, running HSR needs to be deconfigured before it can be reconfigured. This is not in scope of the current role.

That task file checks if:
- HSR is present
- Primary is configured as primary
- Secondary is configured as secondary
- Secondary contains primary hostname

We do not go into detail if all variables and inputs are reflected inside of existing HSR, therefore we cannot say it is fully Idempotent. I am not asking for any code changes, I just want to make sure we do not over-promise by using bad words :)

ja9fuchs:
> We do not go into detail if all variables and inputs are reflected inside of existing HSR, therefore we cannot say it is fully Idempotent. I am not asking for any code changes, I just want to make sure we do not over-promise by using bad words :)

Hence my edit:
> If there is already a HANA system replication active between the instances, the setup is skipped.

marcelmamula (approved):
LGTM
Thank you for fixing issues I raised.

PR #1225, opened 2026-07-22
sap_hana_install: New functionality to add Fast Restart Option config to HANA

rob0d:
This PR adds functionality to sap_hana_install which will enable/disable HANA Fast Restart Option.
The default behaviour of this role not to do anything in relation to FRO so this change doesn't break anything.
This functionality allows to enable FRO as an after-installation step (as this is the correct time to do it).
This functionality also allows to change/disable FRO as is basically the same code. It is slightly illogical that hana_install role can do this, but currently there is no other option. 
Moving this to a separate role will create a dependency of this role to the new role which is undesirable.

### DIFF
--- a/roles/sap_hana_install/README.md
+++ b/roles/sap_hana_install/README.md
@@ -228,7 +228,7 @@ Steps:
 2. Gather list of all Instance profiles in `/hana/shared/<SID>/profile/`.
 3. Gather list of all directories in `/usr/sap/<SID>/HDB<Instance Number>/`.
 4. If the existing instance profile or directory was found in hosts list, create separate list without them.
-    - This list of the `new hosts` is used for one-time tasks. 
+    - This list of the `new hosts` is used for one-time tasks.
 
 
 #### Pre-Tasks for Installation
@@ -312,7 +312,8 @@ Steps:
 7. Apply SELinux policies if the variable `sap_hana_install_configure_selinux` is set to `true`.
 8. (Red Hat specific) Configure `fapolicyd` if the variable `sap_hana_install_configure_fapolicyd` is set to `true`.
 Additionally, if `sap_hana_install_enable_fapolicyd` is set to `true`, also enable and start the `fapolicyd` service.
-9. Output final status of installed system.
+9. Perform Fast Restart Option (FRO) configuration if requested (`sap_hana_install_fro_state` is set to `enabled` or `disabled`).
+10. Output final status of installed system.
 
 
 #### Post-Tasks for Addhosts
@@ -328,7 +329,8 @@ Steps:
 7. Apply SELinux policies if the variable `sap_hana_install_configure_selinux` is set to `true`.
 8. (Red Hat specific) Configure `fapolicyd` if the variable `sap_hana_install_configure_fapolicyd` is set to `true`.
 Additionally, if `sap_hana_install_enable_fapolicyd` is set to `true`, also enable and start the `fapolicyd` service.
-9. Output final status of installed system.
+9. Perform Fast Restart Option (FRO) configuration if requested (`sap_hana_install_fro_state` is set to `enabled` or `disabled`).
+10. Output final status of installed system.
 <!-- END Execution Flow -->
 
 ### Example
@@ -407,6 +409,7 @@ With the following tags, the role can be called to perform certain activities on
 - tag `sap_hana_install_create_configfile`: Only generate the hdblcm configfile.
 - tag `sap_hana_install_check_hana_exists`: Only check if there is already HANA installed with the
   desired SID and instance number.
+- tag `sap_hana_install_fro_configure`: Run only the Fast Restart Option configuration steps. This will have only effect if `sap_hana_install_fro_state` is set to `enabled` or `disabled`.
 
 The following tags have been removed in this release:
 
@@ -456,8 +459,20 @@ ansible-playbook sap-hana-install.yml --tags=sap_hana_install_check_hana_exists
     sap_hana_install_lss_backup_password: ''
     sap_hana_install_secure_store: localsecurestore
     ```
-> **NOTE**: SAP HANA 2.0 SPS08 requires LSS filesystem to be shared across all Scale-Out (cluster) hosts.</br>
-> This role will check if directory defined in `sap_hana_insall_lss_inst_path` is present for SPS08 or higher and if it is mounted as shared file system.
+  > **NOTE**: SAP HANA 2.0 SPS08 requires LSS filesystem to be shared across all Scale-Out (cluster) hosts.</br>
+  > This role will check if directory defined in `sap_hana_insall_lss_inst_path` is present for SPS08 or higher and if it is mounted as shared file system.
+
+- ### Fast Restart Option (FRO)
+  SAP HANA Fast Restart Option is available as of HANA2.0 SPS04. It uses storage in the file system to preserve and reuse MAIN data fragments to speed up SAP HANA restarts. This is effective in cases where the operating system is not restarted.
+
+  For explantation of how this role handles the FRO configuration see the comments in `defaults/main.yml`.
+  For details how FRO works see [HANA online help](https://help.sap.com/docs/SAP_HANA_PLATFORM/6b94445c94ae495c83a19646e7c3fd56/ce158d28135147f099b761f8b1ee43fc.html?locale=en-US&version=LATEST).
+
+  The role parameters that impact FRO configuration are:
+  ```
+  sap_hana_install_fro_state:
+  sap_hana_install_fro_mem_percent:
+  ```
 
 - ### Additional examples
   For more examples on how to use this role in different installation scenarios, refer to the [ansible.playbooks_for_sap](https://github.com/sap-linuxlab/ansible.playbooks_for_sap) playbooks.
--- a/roles/sap_hana_install/defaults/main.yml
+++ b/roles/sap_hana_install/defaults/main.yml
@@ -128,6 +128,31 @@ sap_hana_install_recreate_tenant_database: false
 # If this not desired, set the following variable to `false`.
 sap_hana_install_configure_selinux: true
 
+# Enables/Disables/Keeps the Fast Restart Option (FRO) for SAP HANA.
+# When set to 'enabled', the role will configure tmpfs mounts for all NUMA nodes and set the corresponding HANA parameters.
+# When set to 'disabled', the role will remove any existing tmpfs mounts for FRO and unset the corresponding HANA parameters.
+# When set to 'unchanged', the role will not modify the FRO configuration.
+# NOTE: When enabling FRO the HANA database must be restarted for the changes to take effect. This is not done automatically by the role.
+# NOTE: When disabling FRO, the HANA database must be restarted. This is not handled automatically by the role.
+#   This means that the disabling will fail and will have to be executed twice!
+#   The flow is: 1) run disable (it will fail) 2) restart HANA manually 3) run disable again (it will succeed).
+#   This is a security precaution to avoid data loss and to allow the administrator to follow their own restart procedure.
+sap_hana_install_fro_state: 'unchanged'
+
+# When enabling FRO, this variable defines the percentage of the main memory per NUMA node to use for FRO tmpfs mounts.
+# Make sure you understand the implication of settings this and you have performed the necessary calculations to set this correctly.
+# Default value is 40% which is considered safe and a good starting point for testing (assuming your server was sized according
+#   to SAP HANA hardware sizing guidelines).
+# Setting this to 0 will make the tmpfs unlimited i.e. all available memory will be useable for FRO.
+#   This can lead to out of memory situations if the memory is not restricted using HANA parameters.
+# Setting this to 100 will use all memory at the point of execution of this role i.e. the size will be fixed to 100% as of today.
+#   This can lead to out of memory situations if the memory is not restricted using HANA parameters.
+# Setting this to any other number will set the size of the tmpfs to that percentage of the main memory of the NUMA node at the point of execution of this role.
+#   The calculation is done per each NUMA node and is rounded down (decimals are truncated) to the nearest gigabyte.
+#   For example, if you set this to 50 and you have a NUMA node with 383 GB of main memory, the tmpfs size will be set to 191 GB.
+#   Sensible values are probably somewhere between 40 and 60, but this really depends on your workload and on how much memory you have available.
+sap_hana_install_fro_mem_percent: 40
+
 ################
 # Parameters for hdblcm:
 ################
--- a/roles/sap_hana_install/defaults/main.yml
+++ b/roles/sap_hana_install/defaults/main.yml
@@ -143,7 +143,7 @@ sap_hana_install_fro_state: 'unchanged'
 # Make sure you understand the implication of settings this and you have performed the necessary calculations to set this correctly.
 # Default value is 40% which is considered safe and a good starting point for testing (assuming your server was sized according
 #   to SAP HANA hardware sizing guidelines).
-# Setting this to 0 will make the tmpfs unlimited i.e. all available memory will be useable for FRO.
+# Setting this to 0 will make the tmpfs unlimited i.e. all available memory will be usable for FRO.
 #   This can lead to out of memory situations if the memory is not restricted using HANA parameters.
 # Setting this to 100 will use all memory at the point of execution of this role i.e. the size will be fixed to 100% as of today.
 #   This can lead to out of memory situations if the memory is not restricted using HANA parameters.
--- a/roles/sap_hana_install/tasks/main.yml
+++ b/roles/sap_hana_install/tasks/main.yml
@@ -135,6 +135,14 @@
       ansible.builtin.include_tasks:
         file: post_addhosts.yml
 
+- name: SAP HANA - Main - Post-Tasks - Fast Restart Option (FRO)
+  ansible.builtin.include_tasks:
+    file: post_tasks/fro.yml
+    apply:
+      tags:
+        - sap_hana_install_fro_configure
+  when: sap_hana_install_fro_state in ['enabled', 'disabled']
+  tags: sap_hana_install_fro_configure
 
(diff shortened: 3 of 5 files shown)


marcelmamula:
@rob0d Thank you for contribution.
I will put this PR on hold for now until https://github.com/sap-linuxlab/community.sap_install/pull/1215 is reviewed by @berndfinger  and merged. Then you can rebase or re-create this PR.

FYI: Since `FRO` word is flagged by codespell action, workflow will have to be adjusted in https://github.com/sap-linuxlab/community.sap_install/blob/3e57f3c319f95470df184044f96dd1a55186fe28/.github/workflows/codespell.yml#L25

marcelmamula:
@rob0d As sidenote: You have to follow structure with runner files and add it to `post_install.yml` and `post_addhosts.yml` where it has to follow pseudo-idempotency awareness.

E.g: Use `__sap_hana_install_fact_is_installed` If it is one time activity, otherwise run it without to execute always.

rob0d:
> @rob0d As sidenote: You have to follow structure with runner files and add it to `post_install.yml` and `post_addhosts.yml` where it has to follow pseudo-idempotency awareness.
> 
> E.g: Use `__sap_hana_install_fact_is_installed` If it is one time activity, otherwise run it without to execute always.

Hi @marcelmamula / @berndfinger . This is on purpose. If this code is inside post*.yml it will be executed ONLY if HANA is installed at the same time. This means you won't be able to add FRO to already existing installations (which there are currently 100% of them unless people did it manually). It will also prevent the side function of removing/modifying FRO (I agree it's not the most intuitive functionality of sap_hana_install, however FRO enabled is a part of HANA installation by default so it must sit in this role).

However I have added a condition to check  `__sap_hana_install_fact_is_installed` if the appropriate tag is not defined which means this will execute only if HANA was installed OR if user specifically asks for this functionality via the tag.

Also this code is idempotent as enabling FRO if it's already enabled with the same parameters will make net zero change (this same should apply to disabling, but I can't remember if I definitely checked disabling more than twice - see the note regarding disabling in the comments).

roles/sap_hana_install/defaults/main.yml
```diff
@@ -128,6 +128,31 @@ sap_hana_install_recreate_tenant_database: false
...
 sap_hana_install_configure_selinux: true
 
+# Enables/Disables/Keeps the Fast Restart Option (FRO) for SAP HANA.
+# When set to 'enabled', the role will configure tmpfs mounts for all NUMA nodes and set the corresponding HANA parameters.
+# When set to 'disabled', the role will remove any existing tmpfs mounts for FRO and unset the corresponding HANA parameters.
+# When set to 'unchanged', the role will not modify the FRO configuration.
+# NOTE: When enabling FRO the HANA database must be restarted for the changes to take effect. This is not done automatically by the role.
+# NOTE: When disabling FRO, the HANA database must be restarted. This is not handled automatically by the role.
+#   This means that the disabling will fail and will have to be executed twice!
+#   The flow is: 1) run disable (it will fail) 2) restart HANA manually 3) run disable again (it will succeed).
```
marcelmamula, line 138: Sorry, but this is against idempotency or any kind of intent to automate. If restart is required then it must be handled within FRO workflow, otherwise it defeats purpose.

roles/sap_hana_install/defaults/main.yml
```diff
@@ -128,6 +128,31 @@ sap_hana_install_recreate_tenant_database: false
...
+#   This is a security precaution to avoid data loss and to allow the administrator to follow their own restart procedure.
+sap_hana_install_fro_state: 'unchanged'
```
marcelmamula, line 140: Missing pretasks to handled assert against variable for `if defined`:
- must be non-empty String and one of those 3 options

roles/sap_hana_install/defaults/main.yml
```diff
@@ -128,6 +128,31 @@ sap_hana_install_recreate_tenant_database: false
...
+#   to SAP HANA hardware sizing guidelines).
+# Setting this to 0 will make the tmpfs unlimited i.e. all available memory will be usable for FRO.
+#   This can lead to out of memory situations if the memory is not restricted using HANA parameters.
+# Setting this to 100 will use all memory at the point of execution of this role i.e. the size will be fixed to 100% as of today.
+#   This can lead to out of memory situations if the memory is not restricted using HANA parameters.
+# Setting this to any other number will set the size of the tmpfs to that percentage of the main memory of the NUMA node at the point of execution of this role.
+#   The calculation is done per each NUMA node and is rounded down (decimals are truncated) to the nearest gigabyte.
+#   For example, if you set this to 50 and you have a NUMA node with 383 GB of main memory, the tmpfs size will be set to 191 GB.
+#   Sensible values are probably somewhere between 40 and 60, but this really depends on your workload and on how much memory you have available.
+sap_hana_install_fro_mem_percent: 40
```
marcelmamula, line 154: Missing pretasks to handled assert against variable for `if defined`:
- must be integer between 0 and 100

roles/sap_hana_install/README.md
```diff
@@ -312,7 +312,8 @@ Steps:
 7. Apply SELinux policies if the variable `sap_hana_install_configure_selinux` is set to `true`.
 8. (Red Hat specific) Configure `fapolicyd` if the variable `sap_hana_install_configure_fapolicyd` is set to `true`.
 Additionally, if `sap_hana_install_enable_fapolicyd` is set to `true`, also enable and start the `fapolicyd` service.
-9. Output final status of installed system.
+9. Perform Fast Restart Option (FRO) configuration if requested (`sap_hana_install_fro_state` is set to `enabled` or `disabled`).
```
marcelmamula, line 315: Please keep existing syntax from lines above. Applies to addhosts section below as well.
```
Perform Fast Restart Option (FRO) configuration if the variable `sap_hana_install_fro_state` is set to either `enabled` or `disabled`
```

roles/sap_hana_install/tasks/post_tasks/fro.yml
```diff
@@ -0,0 +1,145 @@
...
+  become: true
+  changed_when: false
+
+- name: SAP HANA - Post-Tasks - FRO - Extract NUMA nodes and memory sizes
+  ansible.builtin.set_fact:
+    __sap_hana_install_fact_numa_nodes: "{{ __sap_hana_install_numa_memory_sizes_output.stdout_lines | map('split') | list }}"
+
+- name: SAP HANA - Post-Tasks - FRO - Define fact for tmpfs base paths
+  ansible.builtin.set_fact:
+    __sap_hana_install_fact_fro_base_path: "{{ __sap_hana_install_fact_numa_nodes | map(attribute=0) | map('regex_replace', '^(.*)$', '/hana/tmpfs\\1/' + sap_hana_install_sid | upper) | join(';') }}"
```
marcelmamula, line 22: Lines are too long.
```yaml
  ansible.builtin.set_fact:
    __sap_hana_install_fact_fro_base_path:
      "{{ __sap_hana_install_fact_numa_nodes
         | map(attribute=0)
         | map('regex_replace', '^(.*)$', '/hana/tmpfs\\1/' + sap_hana_install_sid | upper)
         | join(';') }}"
```

marcelmamula, line 22: This PR will have to go after https://github.com/sap-linuxlab/community.sap_install/pull/1215 so you should use `__sap_hana_install_sid`

roles/sap_hana_install/tasks/post_tasks/fro.yml
```diff
@@ -0,0 +1,145 @@
+---
+# As of SPS08 this can be done using
+# hdblcm --configure_fast_restart_option --tmpfs_mountpoints_path=<path>
+#        --restrict_tmpfs_global_allocation_limit --tmpfs_global_allocation_limit=<limit in MB>
+# However, the approach below is universal regardless of HANA version and also allows changing or removal of the FRO configuration which
+# is not currently possible using hdblcm
+
+- name: SAP HANA - Post-Tasks - FRO - Gather NUMA nodes and their memory sizes
+  ansible.builtin.shell:
+    cmd: "set -o pipefail && cat /sys/devices/system/node/node*/meminfo | grep MemTotal | awk '{print $2, $4}'"
```
marcelmamula, line 10: Consider using Native Ansible code, which removes all shell pipes and has better return code handing.
1. Stat or find files
2. Slurp/Read files /sys/devices/system/node/node*/meminfo if they exist
3. Run filters on stdout of command to extract information.

This applies to all commands below.

roles/sap_hana_install/tasks/post_tasks/fro.yml
```diff
@@ -0,0 +1,145 @@
...
+  register: __sap_hana_install_numa_memory_sizes_output
+  failed_when: __sap_hana_install_numa_memory_sizes_output.rc != 0
+  become: true
+  changed_when: false
+
+- name: SAP HANA - Post-Tasks - FRO - Extract NUMA nodes and memory sizes
+  ansible.builtin.set_fact:
+    __sap_hana_install_fact_numa_nodes: "{{ __sap_hana_install_numa_memory_sizes_output.stdout_lines | map('split') | list }}"
```
marcelmamula, line 18: You are using this unsanitized in follow up loops. Ensure that there is default to handle failures.
Example (I am not sure if map() is sanitized to handle empty input):
```
"{{ __sap_hana_install_numa_memory_sizes_output.stdout_lines | d([]) | map('split') | list }}"
```

roles/sap_hana_install/tasks/post_tasks/fro.yml
```diff
@@ -0,0 +1,145 @@
...
+    group: sapsys
+    mode: '1775'
+  loop: "{{ __sap_hana_install_fact_numa_nodes }}"
+  loop_control:
+    loop_var: numa_node
+    label: "Processing NUMA node {{ numa_node[0] }}"
+  when: sap_hana_install_fro_state == 'enabled'
+
+- name: SAP HANA - Post-Tasks - FRO - Disable FRO by unsetting basepath parameters in HANA and reconfiguring the system
+  ansible.builtin.shell: |
```
marcelmamula, line 45: Always use `cmd:` to keep formatting.
It is safe to use `>-` instead of pipe with `\`, because it will all get processed as one line.
```yaml
  ansible.builtin.shell:
    cmd: >-
      source ~/.profile &&
      HDBSettings.sh setParameter.py
      -unset="SYSTEM/global.ini/persistence/basepath_persistent_memory_volumes"
  args:
```

roles/sap_hana_install/tasks/post_tasks/fro.yml
```diff
@@ -0,0 +1,145 @@
...
+# As of SPS08 this can be done using
+# hdblcm --configure_fast_restart_option --tmpfs_mountpoints_path=<path>
+#        --restrict_tmpfs_global_allocation_limit --tmpfs_global_allocation_limit=<limit in MB>
+# However, the approach below is universal regardless of HANA version and also allows changing or removal of the FRO configuration which
+# is not currently possible using hdblcm
+
+- name: SAP HANA - Post-Tasks - FRO - Gather NUMA nodes and their memory sizes
+  ansible.builtin.shell:
+    cmd: "set -o pipefail && cat /sys/devices/system/node/node*/meminfo | grep MemTotal | awk '{print $2, $4}'"
+  register: __sap_hana_install_numa_memory_sizes_output
```
marcelmamula, line 11: please follow naming convention:
- Registered facts: `__<role>_register_<fact name>`
- Set facts: `__<role>_fact_<fact name>`

Applies to all wrong names below.

roles/sap_hana_install/tasks/post_tasks/fro.yml
```diff
@@ -0,0 +1,145 @@
...
+  ansible.builtin.shell: |
+    source ~/.profile && HDBSettings.sh \
+    setParameter.py  -unset="SYSTEM/global.ini/persistence/basepath_persistent_memory_volumes"
+  args:
+    executable: /bin/bash
+  register: __sap_hana_install_hana_setparam_disable
+  become: true
+  become_user: "{{ sap_system_sid | lower }}adm"
+  changed_when: __sap_hana_install_hana_setparam_disable.rc == 0
+  failed_when: __sap_hana_install_hana_setparam_disable.rc != 0
```
marcelmamula, line 54: When using `failed_when`, you must be 100% sure that Ansible will output `stderr` or `stdout`, otherwise users will never know why it failed.

This is reason why I am avoiding `failed_when`, but rather add clear `fail` task afterwards. Example:
```yaml
    - name: SAP SWPM Pre Install - Detect - Show reason why sapcontrol check was skipped - command failed - {{ instance_item.number | d('') }}
      ansible.builtin.debug:
        msg: |
          INFO: Detection of existing SAP instance installation with 'sapcontrol' command was skipped.
          Return code: {{ __sap_swpm_register_sapcontrol_instance_properties.rc | d('1') }}
          Output:      {{ __sap_swpm_register_sapcontrol_instance_properties.stdout | d('') }}
          Error:       {{ __sap_swpm_register_sapcontrol_instance_properties.stderr | d('') }}
      when:
        - __sap_swpm_register_sapcontrol_instance_properties.rc is defined
        - __sap_swpm_register_sapcontrol_instance_properties.rc != 0
```

roles/sap_hana_install/tasks/post_tasks/fro.yml
```diff
@@ -0,0 +1,145 @@
...
+  when: sap_hana_install_fro_state == 'disabled'
+
+- name: SAP HANA - Post-Tasks - FRO - Unmount TMPFS for all NUMA nodes if FRO is disabled to allow cleanup of mount points
+  become: true
+  ansible.posix.mount:
```
marcelmamula, line 59: You cannot add external dependency without mentioning it in dependency list in README.md.

Dependencies have to be properly reviewed and agreed upon. FYI @berndfinger @rhmk

roles/sap_hana_install/tasks/post_tasks/fro.yml
```diff
@@ -0,0 +1,145 @@
...
+    src: "tmpfs{{ sap_hana_install_sid | lower }}{{ numa_node[0] }}"
+    path: "/hana/tmpfs{{ numa_node[0] }}/{{ sap_hana_install_sid | upper }}"
+    fstype: tmpfs
+    opts: "\
+      rw,relatime,mpol=prefer:{{ numa_node[0] }},\
+      {{ ('size=' + ((numa_node[1] | int * sap_hana_install_fro_mem_percent | int // 100) // 1048576) | string + 'G,') if sap_hana_install_fro_mem_percent > 0 else '' }}\
```
marcelmamula, line 65: Long line

roles/sap_hana_install/tasks/main.yml
```diff
@@ -135,6 +135,15 @@
       ansible.builtin.include_tasks:
         file: post_addhosts.yml
 
+- name: SAP HANA - Main - Post-Tasks - Fast Restart Option (FRO)
```
marcelmamula, line 138: This task must be moved to `post_install.yml` and `post_addhosts.yml`.

Also using `__sap_hana_install_fact_is_installed` does not serve any purpose here, because it is detected at beginning and does not change afterwards. It is used to control if installation is executed, but also to limit one-time post tasks.

Prerequisite https://github.com/sap-linuxlab/community.sap_install/pull/1215 can give you more details, but essentially when workflow gets here:
- Installation was done, `__sap_hana_install_fact_is_installed` is `false` and task must run.
- Installation was skipped, `__sap_hana_install_fact_is_installed` is `true` and task must run for idempotency.
- Installation failed and it never gets here

This explains why using it for idempotent tasks does not serve any purpose unless you want to ensure it runs only once per installation, aka never again.

marcelmamula:
@rob0d Your comment about `post_tasks` running only when it is installed is not true. See my review comments for more details. This is excerpt from `main.yml`:
```yaml
# Instructions for separating tasks to achieve idempotency:
# - Installation - Run once for new installations, never repeated.
# - Configuration - Run always to ensure idempotent outcome.
```

roles/sap_hana_install/README.md
```diff
@@ -456,8 +459,20 @@ ansible-playbook sap-hana-install.yml --tags=sap_hana_install_check_hana_exists
...
+  > This role will check if directory defined in `sap_hana_insall_lss_inst_path` is present for SPS08 or higher and if it is mounted as shared file system.
+
+- ### Fast Restart Option (FRO)
+  SAP HANA Fast Restart Option is available as of HANA2.0 SPS04. It uses storage in the file system to preserve and reuse MAIN data fragments to speed up SAP HANA restarts. This is effective in cases where the operating system is not restarted.
+
+  For explantation of how this role handles the FRO configuration see the comments in `defaults/main.yml`.
+  For details how FRO works see [HANA online help](https://help.sap.com/docs/SAP_HANA_PLATFORM/6b94445c94ae495c83a19646e7c3fd56/ce158d28135147f099b761f8b1ee43fc.html?locale=en-US&version=LATEST).
+
+  The role parameters that impact FRO configuration are:
+  ```
```
marcelmamula, line 472: Change it to markdown list without values. Current format could lead to people setting `None` variables.
```markdown
  The role parameters that impact FRO configuration are:

  - sap_hana_install_fro_state
  - sap_hana_install_fro_mem_percent
```

NOTE: Empty line between string line and actual list is not a typo, but mandatory for Galaxy formatting.

marcelmamula (changes requested):
@rob0d This looks like interesting addition, but there are issues that must resolved before it is tested and/or merged.
First one: Pre-requisite https://github.com/sap-linuxlab/community.sap_install/pull/1215 must be merged.

Second: All review comments below span range of different issues from role consistency, best practices, project rules as well as formatting.

marcelmamula (changes requested):
@rob0d This looks like interesting addition, but there are issues that must resolved before it is tested and/or merged.
First one: Pre-requisite https://github.com/sap-linuxlab/community.sap_install/pull/1215 must be merged.

Second: All review comments below span range of different issues from role consistency, best practices, project rules as well as formatting.

PR #1227, opened 2026-07-23
collection: Update min Ansible version and update workflows

marcelmamula:
## Changes
- Change of minimum Ansible version to `2.18` which is currently being applied across whole project.
  - **NOTE:** This change caused incompatibility with ansible-test workflow for EOL version and as result those tests were commented out.

- Remove obsolete `meta/main.yml` files from Ansible Roles, because we do not publish roles, only collections which do not need them.

- Update README.md to reflect version changes.

- Update workflows:
  - Updated schedule to avoid conflict with other collections in project.
    - `sap_libs` workflows run daily.
    - All other collections have dedicated weekday and run weekly (`ansible-lint` runs 1 hour after `ansible-test`).
  -  Add parallel limit to reduce API issues.
  - Updated Ansible Lint to `v26`
  - Updated path list to disable sanity tests when only documentation is updated.


rhmk:
LGTM

berndfinger (approved):
LGTM.

PR #1228, opened 2026-07-29
sap_general_preconfigure, sap_hana_preconfigure: Fixes for Ansible 2.19

berndfinger:
These avoid error messages which occur on Ansible 2.19 and later on RHEL 7, 8 and 9 if the length of null defined vars is requested, for example:

`[ERROR]: Task failed: The filter plugin 'ansible.builtin.length' failed: object of type 'NoneType' has no len()`

Related task is: https://github.com/sap-linuxlab/community.sap_install/blob/931e140bd23bb1ae55ddb960f4b1c8dacb8dd7a2/roles/sap_hana_preconfigure/tasks/RedHat/installation.yml#L143 , and this task is inside a block starting here: https://github.com/sap-linuxlab/community.sap_install/blob/931e140bd23bb1ae55ddb960f4b1c8dacb8dd7a2/roles/sap_hana_preconfigure/tasks/RedHat/installation.yml#L135 .

The other changes in this PR are related:
- ensuring correct variable type definitions for two vars in `sap_general_preconfigure/meta/argument_specs.yml`
- removing superfluous comments in the RHEL vars files of `sap_hana_preconfigure` after the all affected vars have been correctly defined

I have not tested on SLES (I also changed the SLES related vars file of `sap_hana_preconfigure`) and maybe no code is evaluating those vars for SLES, but I modified them just so that all these vars are defined the same way as they are now in RHEL.

### DIFF
--- a/roles/sap_general_preconfigure/meta/argument_specs.yml
+++ b/roles/sap_general_preconfigure/meta/argument_specs.yml
@@ -157,7 +157,8 @@ argument_specs:
         example:
           '@minimal-environment'
         required: false
-        type: str
+        type: list
+        elements: str
 
       sap_general_preconfigure_envgroups:
         default: "{{ __sap_general_preconfigure_envgroups }}"
@@ -167,7 +168,8 @@ argument_specs:
         example:
           '@minimal-environment'
         required: false
-        type: str
+        type: list
+        elements: str
 
       sap_general_preconfigure_patterns:
         default: "{{ __sap_general_preconfigure_patterns }}"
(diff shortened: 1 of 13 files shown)


ja9fuchs (approved):
lgtm

marcelmamula (approved):
LGTM
- there is no change impacting SUSE as  those are Red Hat specific variables.
- It brings forth topic from https://github.com/sap-linuxlab/community.sap_launchpad/pull/78#pullrequestreview-4784514307, because you dont have to default something in OS vars if you default it in main. OS files should be used only as override of vars/main file.

PR #1231, opened 2026-08-14
sap_swpm: Replace SAP detection fail with warning for wrong GID and UID

marcelmamula:
## Changes
https://github.com/sap-linuxlab/community.sap_install/pull/1213 brought a lot of improvements including much stricter validations.
One of them was bit too strict as re-testing has showed that `sapinst` has some leniency when handling existing users and groups.

- Change `fail` modules to `debug` to show `WARN` instead.

**Accepts:**
- Existing users `sidadm` and `sapadm` with different IDs.
- Existing group `sapsys` with different ID.

**Rejects:**
- Existing files under user with different group ID. Example: ASCS with `ae1adm GID: 3333` and ERS with `ae1adm GID: 3334`
   **NOTE: This is specific only to distributed systems with shared filesystems!**
```bash
Group sapsys exists with GID 3334, while the default profile /sapmnt/AE1/profile/DEFAULT.PFL is owned by GID 3333. <br>SOLUTION: Change the GID of group sapsys to 3333 or remove the group account.
``` 

## HANA Details
- No changes to strict validation of `sap_hana_install`, because of its own strict behavior documented in test section below.

**Accepts:**
- **Wrong** UID with **correct** GID and **correct** `sidadm` password provided in parameter file under `password=`

**Rejects:**
- **Wrong** UID with **wrong** GID and **correct** `sidadm` password provided in parameter file under `password=`
```bash
The defined user 'h01adm' already exists on the system. Neither the password, nor any other attribute of the user will be changed.
Verify that the user is correct.
Running in batch mode
  Group ID '2001' for 'sapsys' is invalid because 'sapsys' already exists with ID '3335' on the local host
```
- **Wrong** UID with **wrong** GID and **missing** `sidadm` password provided in parameter file under `password=`
```bash
The defined user 'h01adm' already exists on the system. Neither the password, nor any other attribute of the user will be changed.
Verify that the user is correct.
Running in batch mode
  Mandatory parameter 'password' (Password) is missing or invalid
```

### DIFF
--- a/roles/sap_swpm/tasks/pre_install/detect_sap.yml
+++ b/roles/sap_swpm/tasks/pre_install/detect_sap.yml
@@ -34,7 +34,7 @@
 
 # Check 1: Check if the group 'sapsys' exists with correct ID.
 # The role supports specifying the 'sapsys' group id in variable `sap_swpm_sapsys_gid`.
-# The installation will fail if there is already a group named sapsys but with a different ID.
+# The installation will show warning if there is already a group named sapsys but with a different ID.
 # NOTE: This is executed even if the above checks detect an existing installation, because we want to identify conflicts.
 
 # This is also used for validation before using 'sidadm' sapcontrol.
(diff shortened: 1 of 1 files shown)


berndfinger (approved):
LGTM

PR #1232, opened 2026-08-31
sap_hana_preconfigure/RHEL: Add package for GCC 15 compatibility

berndfinger:
Also provide the `RedHat_x.y.plus.yml` vars file logic in `sap_general_preconfigure` as we already have in `sap_hana_preconfigure`. I have tested it extensively before I decided to implement the GCC 15 package logic only in the role `sap_hana_preconfigure`, and then decided to leave the logic in place for `sap_general_preconfigure`, in case we need it in the future. With that, this logic is still missing in `sap_netweaver_preconfigure`. We will probably never need it there because we will likely cover it by `sap_general_preconfigure`.

Also no longer ensure the presence of package `compat-sap-c++-13` for RHEL 9.1 and RHEL 9.3, where it is not available (and not neeeded by SAP HANA).

The package compat-sap-c++-15 is present in:
- RHEL 9.6
- RHEL 9.8 and all later minor versions

The package compat-sap-c++, currently in version 15, is present in:
- RHEL 10.0
- RHEL 10.2 and all later minor versions


marcelmamula (approved):
LGTM. Since SLES related changes are only in task main, then test was done to see no impact.

```bash
TASK [community.sap_install.sap_general_preconfigure : Examine the OS minor version] ***********************************************************************
ok: [s02aas]

TASK [community.sap_install.sap_general_preconfigure : Include OS major specific vars] *********************************************************************
skipping: [s02aas] => (item=Suse.yml)
skipping: [s02aas] => (item=SLES_SAP.yml)
skipping: [s02aas] => (item=SLES_16.yml)
ok: [s02aas] => (item=SLES_SAP_16.yml)

TASK [community.sap_install.sap_general_preconfigure : Include minor plus specific vars, RHEL only] ********************************************************
skipping: [s02aas]

TASK [community.sap_install.sap_general_preconfigure : Include OS minor specific vars] *********************************************************************
skipping: [s02aas] => (item=SLES_16.0.yml)
skipping: [s02aas] => (item=SLES_SAP_16.0.yml)
skipping: [s02aas]
```

ja9fuchs (approved):
lgtm 👍 

Tested on RHEL 10.0.
Verified: compat-sap-c++ gets installed

PR #1235, opened 2026-09-07
sap_install_media_detect: Rework for performance and accuracy

marcelmamula:
## Premise
Rework of the `sap_install_media_detect` role for speed, parity with the other roles in this collection, and removal of lazy validations. Detection now drives the behaviour instead of single-purpose parameters, and every step reports what it did.

## Changes
- Reworked `sapfile` utility script to remove `awk` text blocks and simplify any future changes by using `lists` and `arrays`.
- The source and target directory state decides the workflow, so a read-only source is handled without a dedicated parameter or a second run.
- Copying into the target directory uses `ansible.posix.synchronize` when the collection is available on the Control node and falls back to `ansible.builtin.copy` when it is not. Recursive finds over the whole tree were replaced with targeted finds. This is the bulk of the runtime improvement.
- Requested components are validated during classification, so missing installation media fails before anything is copied into a subdirectory or extracted. `assert` was replaced with `fail` and an inverted condition, so failures no longer leak the internal expression as `assertion:`.
- Reduced tags to `sap_install_media_detect_reset`, `sap_install_media_detect_prepare` and `sap_install_media_detect_classify`. Each runs the workflow from the beginning up to a defined point rather than being an entry point into the middle of it.
- New detailed summary added at end of execution showing even partial execution (controlled by tags or failure).
- 9 ownership and permission parameters for the created subdirectories, the installation media files and `SAPCAR`; `sap_install_media_detect_sapcar_path` as a fully qualified path; `sap_install_media_detect_rar_use_unar`, with EPEL disabled again at the end of the role if the role enabled it.

### Removals and deprecations
- `sap_install_media_detect_file_server_only` - replaced by scenario detection.
- `sap_install_media_detect_rar_package` - replaced by `sap_install_media_detect_rar_use_unar`.
- 11 tags that existed for development purposes only.
- All unecessary occurrences of `ignore_errors: true`.
- 13 exported variables that no longer exist in subsequent roles.

## Scenarios
| Scenario | Source directory | Target directory | Main directory | Result |
| :--- | :--- | :--- | :--- | :--- |
| **A** | writable | not defined | source directory | runs |
| **B** | writable | defined | target directory | runs |
| **C** | read only | defined | target directory | runs |
| **D** | read only | not defined | - | **fails** |

## Tests
First value is a clean run, following values are re-runs.
> **NOTE:** All times were captured twice: Once in initial state, second was immediate re-run.
> This primarily shows difference in speed of copy commands, but it can be affected by caching.

| Media set | Size | Scenario | Current | Rework | Performance increase |
| :--- | ---: | :--- | :--- | :--- | ---: |
| SAP S/4HANA 2023 | 44 GB | A | 5m00 / 5m56 | 4m48 / 4m58 | +4% / +16% |
| SAP BW/4HANA 2023 | 14 GB | A | 4m03 / 4m35 | 3m49 / 3m58 | +6% / +14% |
| SAP HANA 2 SPS07 | 3.4 GB | A | 2m20 / 2m21 | 2m16 / 2m22 | +3% / 0% |
| SAP S/4HANA 2023 | 44 GB | B/C | 15m08 / 12m24 | 8m41 / 4m15 | **+43% / +66%** |
| SAP BW/4HANA 2023 | 14 GB | B/C | 6m15 / 4m32 | 3m50 / 2m59 | **+39% / +34%** |
| SAP HANA 2 SPS07 | 3.4 GB | B/C | 2m07 / 1m41 | 1m38 / 1m29 | **+23% / +12%** |

All of these were executed using variables set in AP4S.
```
# SAP S/4HANA 2023
        sap_install_media_detect_swpm: true
        sap_install_media_detect_hostagent: true
        sap_install_media_detect_igs: true
        sap_install_media_detect_kernel: true
        sap_install_media_detect_kernel_db: 'saphana'
        sap_install_media_detect_db: "saphana"
        sap_install_media_detect_db_client: "saphana"
        sap_install_media_detect_export: "saps4hana"

# SAP BW/4HANA 2023
        sap_install_media_detect_swpm: true
        sap_install_media_detect_hostagent: true
        sap_install_media_detect_igs: true
        sap_install_media_detect_kernel: true
        sap_install_media_detect_db: "saphana"
        sap_install_media_detect_db_client: "saphana"
        sap_install_media_detect_export: "sapbw4hana"

# SAP HANA 2 SPS07
        sap_install_media_detect_hostagent: true
        sap_install_media_detect_db: "saphana"
```


berndfinger (approved):
LGTM!

Issue #1236, opened 2026-09-14
sap_hana_install: Installation without sap_install_media_detect fails at first attempt


berndfinger:
### Ansible Role

sap_hana_install

### OS Family

N/A

### Ansible Controller - Python version

Python 3.14.7

### Ansible-core version

ansible [core 2.20.0]

### Bug Description

When using the role `sap_hana_install` to install SAP HANA SPS08 the first time without the role `sap_install_media_detect` and without an existing `extracted` directory, the file `SIGNATURE.SMF` gets created by the role `sap_hana_install` with permissions `0600`, causing the installation of SAP HANA to fail with:
```
    stderr: |-
        Running in batch mode
          SAP HANA Local Secure Store requires that the selected component(s) 'server' have a valid signature. See SAP note 2078425, section 'How to prepare SAP HANA packages for installation or update' for instructions how to prepare packages for update or implications of ignoring this error.
        To trust the 'server' component(s) without signature use the option '--lss_trust_unsigned_components'.
```

### Bug reproduction

Install SAP HANA SPS08 without first extracting the file `IMDB_SERVER20_089_4-80002031.SAR`.

### Community participation

Happy to implement this bug fix

berndfinger:
Fixed in #1237.

PR #1237, opened 2026-09-14
sap_hana_install: Ensure correct permissions of SIGNATURE.SMF

berndfinger:
Solves issue #1236.


marcelmamula (approved):
I am not able to reproduce this issue, even after trying it together with @berndfinger in many different configurations.

Regardless, this change is low impact and gets LGTM

Issue #1240, opened 2026-09-15
sap_swpm: Replace set_fact with non-internal facts with internal facts


marcelmamula:
## Problem
Current `sap_swpm` role is setting user variables based on two criteria:
- If deprecated variable was used.
- If parameter dictionary variable is used.

This creates multiple problems:
- Newly set `user facts` are on different precedence rank (19) compared to actually user set variables.
- Subsequent run will override all user variables, unless they are set on precedence ranks 20-22. See https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_variables.html#understanding-variable-precedence

## Affected facts that are set in the role
- `sap_swpm_inifile_sections_list`
- `sap_swpm_inifile_parameters_dict`
- `sap_swpm_product_catalog_id`
- `sap_swpm_role_parameters_dict`
- `sap_swpm_sid`
- `sap_swpm_fqdn`
- `sap_swpm_db_schema`
- `sap_swpm_db_schema_password`
- `sap_swpm_sapadm_password`
- `sap_swpm_sap_sidadm_password`
- `sap_swpm_diagnostics_agent_password`
- `sap_swpm_mp_stack_file_name`

## Solution
1. Create dedicated internal `__sap_swpm_fact_*` facts for all these variables and set them based on user input or dictinary, never set user facing variables.
2. Create dedicated file for handling allowed keys in `sap_swpm_role_parameters_dict` so that correct internal facts can be set for valid user variables.
3. Explain purpose of `sap_swpm_role_parameters_dict` in README.md and defaults, which is missing now.

marcelmamula:
Assigned to Bernd with lower priority as it will require some investigation how to replace current workflows.

Issue #1238, opened 2026-09-14
sap_swpm: Persistent facts in same play


marcelmamula:
## Premise
The `ansible.playbooks_for_sap` are executing variety of our Ansible roles on same hosts in process of configuring system.
This issue was addressed in https://github.com/sap-linuxlab/community.sap_install/pull/1235 and we need to do same here.

Distributed scenarios like `sap_s4hana_distributed_ha` are executing `sap_swpm` twice on same PAS host.
1. DB load with `NW_ABAP_DB:S4HANA2025.CORE.HDB.ABAPHA`
2. PAS installation with `NW_ABAP_CI:S4HANA2025.CORE.HDB.ABAP`

## Problem
First run with DB load execution contains `.ABAPHA` which sets `__sap_swpm_fact_installation_type: "ha"`
```bash
TASK [community.sap_install.sap_swpm : SAP SWPM Pre Install - Display contents of required variables] ******************************************************
ok: [s02pas] =>
    msg: |-
        sap_swpm_product_catalog_id:   'NW_ABAP_DB:S4HANA2025.CORE.HDB.ABAPHA'
...

TASK [community.sap_install.sap_swpm : SAP SWPM Pre Install - Check if SAP High Availability installation] *************************************************
ok: [s02pas]
...

TASK [community.sap_install.sap_swpm : SAP SWPM Pre Install - Assert that necessary variable for HA is defined] ********************************************
ok: [s02pas] =>
    changed: false
    msg: 'PASS: ''sap_swpm_virtual_hostname'' is defined, as ''s02pas''.'
```

Second run with PAS installation does not contain it, but fact is still set because `vars/main.yml` does not override facts set in previous run in same playbook.
```bash
TASK [community.sap_install.sap_swpm : SAP SWPM Pre Install - Display contents of required variables] ******************************************************
ok: [s02pas] =>
    msg: |-
        sap_swpm_product_catalog_id:   'NW_ABAP_CI:S4HANA2025.CORE.HDB.ABAP'
...

TASK [community.sap_install.sap_swpm : SAP SWPM Pre Install - Check if SAP High Availability installation] *************************************************
skipping: [s02pas]
...

TASK [community.sap_install.sap_swpm : SAP SWPM Pre Install - Assert that necessary variable for HA is defined] ********************************************
[ERROR]: Task failed: Action failed: FAIL: 'sap_swpm_virtual_hostname' is not defined or is defined with an empty string!
[...]

PR #1241, opened 2026-09-15
sap_swpm: Add fact reset for variable precedence and improve fact naming

marcelmamula:
## Premise
Fix for https://github.com/sap-linuxlab/community.sap_install/issues/1238 which is caused by:
- Accidental removal in https://github.com/sap-linuxlab/community.sap_install/pull/1213 and moving them to `vars/main.yml`.
- Existing unhandled facts.
- Reliance on `default()` filter, which hides this issue.

## Changes
- Add dedicated `reset_facts.yml` task file, which contains all tasks that are not handled by code and need to be reset.
- Rename all facts to project wide naming:
  - Fact: `__<role name>_fact_<fact name>`
  - Register: `__<role name>_register_<register name>`
- Remove `default()` where it was made obsolete by fact reset.
- Fix `pre_install.yml` summary j2, because it could attempt to show files when those were not found.
- Fix formatting and messages in `detect_variables_from_inifile.yml`.
- Fix typo in  `SAP SWPM Pre Install - Report if 'sap_swpm_fqdn' has been defined differently`
- Split some longer lines.

## NOTE: Some facts do not need reset
Explanation is part of `reset_facts.yml` task file:
```yaml
# NOTE: Not every fact of this role is reset here.
# A fact needs an entry above only if a code path exists where a consumer of the fact runs,
# but the task setting the fact does not. This happens in three cases:
#   1. The setting task, or a block containing it, has a 'when' condition.
#   2. The task file containing it is not included, or is included with an empty loop.
#   3. The setting task is excluded by tags, while a consumer is not.
# If none of these apply, the fact is overwritten on every run and needs no reset.
#
# Example: '__sap_swpm_fact_detect_instance_dirs_matching' is not reset.
# - It is set unconditionally in 'pre_install/detect_sap_instance.yml', which is only ever
#   included after the setting task, so every consumer sees a value from the current run.
# - Stale input is handled at the point of use instead, with the default filter on
#   '__sap_swpm_register_instance_dirs_find.files'.
```

## NOTE
This change does not include solution for identified issue with handling non-internal facts described in https://github.com/sap-linuxlab/community.sap_install/issues/1240


berndfinger (approved):
LGTM

PR #1243, opened 2026-09-18
collection: Add new plugins for SID and Instance Number validation

marcelmamula:
## Premise
Recent rework activities included adding assert for SID and Instance Numbers so I though back to hostname validation plugins created by @berndfinger. Adding SID and Instance Number plugins is first step, after which we can add them to affected roles.

Affected roles where SID needs to be validated with new plugins:
- Role `sap_anydb_install_oracle` with the variable `sap_anydb_install_oracle_sid`
- Role `sap_ha_install_anydb_ibmdb2` with the variable `sap_ha_install_anydb_ibmdb2_sid`
- Role `sap_ha_install_hana_hsr` with the variable `sap_ha_install_hana_hsr_sid`
- Role `sap_ha_pacemaker_cluster` with variables `sap_ha_pacemaker_cluster_hana_sid` or `sap_ha_pacemaker_cluster_nwas_sid`
- Role `sap_hana_install` with variable `sap_hana_install_sid`
- Role `sap_storage_setup` with variable `sap_storage_setup_sid`
- Role `sap_swpm` with variable `sap_swpm_sid`
- Plus other roles in other collections.

Affected roles where Instance Number needs to be validated with new plugins:
- Role `sap_ha_install_hana_hsr` with the variable `sap_ha_install_hana_hsr_instance_number`
- Role `sap_ha_pacemaker_cluster` with range of the variables like `sap_ha_pacemaker_cluster_nwas_ascs_instance_nr`
- Role `sap_hana_install` with the variable `sap_hana_install_number`
- Role `sap_storage_setup` with range of the variables like `sap_storage_setup_nwas_abap_ascs_instance_nr`
- Role `sap_swpm` with range of the variables like `sap_swpm_ascs_instance_nr`

## Changes
- Move duplicated code from hostname validation and create `plugin_utils` from it, then call it from both test and filter.
- Create SID validation using hostname as role model.
  - **NOTE:** SWPM and `hdblcm` have different reserved SID lists so current usage is incorrect.
  - Reserved IDs are now hardcoded in `sap_ha_pacemaker_cluster` and `sap_hana_install`.
- Create Instance Number validation using hostname as role model.

> **NOTE:** This PR introduces new filters, while role adjustments will have to go in separate PRs.

## Tests
Tested all possible scenarios using following test playbooks:

<details>
<summary>Click here to expand the code block Hostname testing</summary>

```yaml
- name: Ansible Play for testing the SAP validation test and filter plugins
  hosts: localhost
  connection: local
[...]

PR #1245, opened 2026-09-18
sap_hana_install : Replace SID and Instance Number validation with new plugins

marcelmamula:
## Changes
- Replace SID and Instance Number asserts with newly merged plugins in https://github.com/sap-linuxlab/community.sap_install/pull/1243
- Remove hardcoded list of prohibited SIDs.

## Tests
Tested with various values as mentioned in PR 1243.

```bash
    msg: |-
        FAIL: The variable 'sap_hana_install_number' is not a valid SAP HANA Instance Number.
        - Length must be exactly 2 characters, got 6!
        - Must consist of digits only, got 'X- 11.'!

    msg: |-
        FAIL: The variable 'sap_hana_install_sid' is not a valid SAP SID.
        - Must not start with a number, starts with '2'!
        - Must not contain any characters other than alpha characters or digits!

    msg: |-
        FAIL: The variable 'sap_hana_install_number' is not a valid SAP HANA Instance Number.
        - Must be a string, got _AnsibleTaggedInt '0'. Quote the value to keep it a string and to preserve a leading zero!

    msg: |-
        FAIL: The variable 'sap_hana_install_number' is not a valid SAP HANA Instance Number.
        - Must be between 00 and 97 for hdblcm, got '99'!
```

But also passing values were tested
```
TASK [community.sap_install.sap_hana_install : SAP HANA - Pre-Tasks - Assert that the variable 'sap_hana_install_sid' is defined as String with valid SAP SID] ***
ok: [b01hana] =>
    changed: false
    msg: All assertions passed

TASK [community.sap_install.sap_hana_install : Assert that the variable 'sap_hana_install_number' is defined as String with valid SAP HANA Instance Number] ***
ok: [b01hana] =>
    changed: false
    msg: All assertions passed
```


berndfinger (approved):
LGTM!

